PR #5, opened 2018-12-21
Remove unnecessary legacy classes

bameyrick:
There were references to the mars font size in some of the templates, I've removed them as they're no longer necessary


bameyrick:
[References to old font sizes](https://trello.com/c/yh1H88OT/59-references-to-old-font-sizes)

PR #11, opened 2018-12-21
Fix example component not rendering scss if no nunjucks template found

bameyrick:


bameyrick:
[Missing scss for beta-bar example](https://trello.com/c/xQSqMUt9/63-missing-scss-for-beta-bar-example)

PR #14, opened 2018-12-21
Checkboxes documentation

jrbarnes9:


jrbarnes9:
[Checkboxes](https://trello.com/c/JS5NheQm/1-checkboxes)

src/components/checkboxes/index.njk
```diff
@@ -1,13 +1,44 @@
...
     patternlibExample("components/checkboxes/examples/checkboxes/index.njk")
 }}
 
+## When to use this component
+Use the checkboxes component when you need to help users select multiple options from a list.
+
+## When to not use this component
+Do not use the checkboxes component if users can only choose one option from a list. In this case, use the [radios](/components/radios) component.
+
+## How to use this component
```
bameyrick: As the macro does this for a user, does this need to be in the docs?

boxadesign: @bameyrick It'll only be runner that uses the macro approach in the short term when that gets implemented. We could rename the title "Structure of the component" (or something).

bameyrick: Fair enough, I'll mark this as resolved

PR #18, opened 2019-01-02
Fix indented markdown (from nunjucks renderer) from being wrapped in code blocks by removing indents

### LINKED ISSUE #17: Adding two "`code`" blocks in markdown breaks next line
**Describe the bug**
If you add two code tags in markdown the next line will be wrapped in code tags. E.g.

This is a `tag` and this is a `tag`

but this should not be a tag.

**Screenshots**
screenshot 2019-01-02 at 09 51 20

bameyrick:
Nunjucks loader is run first and will add indentation to markdown which then the markdown renderer will wrap in code blocks. This uses a regex replacement to remove the indentation before the markdown to html renderer is run.

Resolves #17

### DIFF
--- a/lib/nunjucks-html-loader.js
+++ b/lib/nunjucks-html-loader.js
@@ -6,7 +6,6 @@ import marked from 'marked';
 import pretty from 'pretty';
 import chalk from 'chalk';
 import { htmlErrorStack } from 'print-error';
-import stripIndent from 'strip-indent';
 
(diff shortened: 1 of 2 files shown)


bameyrick:
[Adding two "`code`" blocks in markdown breaks next line](https://trello.com/c/BZhUH990/65-adding-two-code-blocks-in-markdown-breaks-next-line)

PR #44, opened 2019-01-03
Add breadcrumbs forum link

bameyrick:


bameyrick:
[Breadcrumbs](https://trello.com/c/UccXArd2/10-breadcrumbs)

src/components/breadcrumb/index.njk
```diff
@@ -24,4 +24,11 @@ If you’re using other navigational elements on the page, such as a sidebar, co
 - The separators are part of the visual presentation that signifies the breadcrumb trail. This is already semantically represented by the `nav` element with its `aria-label` of Breadcrumb. Therefore using a display technique that is not used by screen readers prevents redundant and potentially distracting verbosity.
 
 ## Research on this component
-Based upon [W3 principles](https://w3c.github.io/aria-practices/examples/breadcrumb/index.html)
+{% from "components/panel/src/_macro.njk" import onsPanel %}
+{% call onsPanel() %}
+    <p>Based upon <a href="https://w3c.github.io/aria-practices/examples/breadcrumb/index.html" target="_blank">W3 principles</a></p>
```
jrbarnes9: I think any research findings, such as the 'Based upon W3 principles' should be written up under the heading, and we should just end the section with the alert panel which only asks people for contributions.  So I would take this line out of the panel and place it above.

PR #65, opened 2019-01-09
Rationalised panels

boxadesign:
No breaking changes.


jrbarnes9:
@boxadesign Need to update the error examples now that error list examples have moved under styles/typography/lists

PR #68, opened 2019-01-11
35 definition

bameyrick:
I've changed from using plain divs to using `<details>` and `<summary>` elements as they semantically make more sense. I've still got all the aria stuff on there and JavaScript to run it because some browsers don't support it, although we should probably check all the accessibility stuff still makes sense in browsers that do support details and summary.

I've also renamed the BEM class from `collapsible` to `details` to be more in convention with how we assign classes to other components.


boxadesign:
@bameyrick With this approach I take it there will need to be a separate set of scripts for `accordion`?

Other than the question LGTM!

bameyrick:
@boxadesign I will make collapsible work for accordion too, so I'll modify what i've added now, but I thought I'd do that work in a separate PR for accordion

PR #74, opened 2019-01-14
Add ability to have child items in pattern library secondary navigation

bameyrick:


bameyrick:
[Enable child items in secondary nav](https://trello.com/c/Zi9gS4sT/81-enable-child-items-in-secondary-nav)

bameyrick:
Will change the way that this works to work off heading anchors rather than pages so closing this PR

PR #75, opened 2019-01-14
45 status

jrbarnes9:


jrbarnes9:
[Status](https://trello.com/c/CAt94mc6/45-status)

src/scss/vars/_colors.scss
```diff
@@ -9,13 +9,15 @@ $color-white: #fff;
 // brand
 $color-astral-blue: #3b7a9e;
 $color-green: #0f8243;
+$color-bright-green: scale-color($color-green, $lightness: 20%, $saturation: 20%);
```
bameyrick: It might be best just to write these out as hex values? Is there any benefit in the colour being based our `$color-green`?

jrbarnes9: @bameyrick The thinking was that if the values of the base colours ever change, we'd only have to change the one value, and the bright variant would automatically retain a 20% lighter/more saturated colour.  Would probably be worth applying to the tint too.

PR #83, opened 2019-01-17
31 accordion

bameyrick:
This could probably do with more documentation, design wise and technical implementation. I could do the technical but might not be best for the design and rationale documentation.


src/components/accordion/index.njk
```diff
@@ -0,0 +1,27 @@
...
+
+
+## When to use this component 
+This component can be used to condense large sections of related content on a page. It allows a user to quickly scan a collated group of content titles.
+
+## When not to use this component
+Accordions should not be used if the content inside the accordion is critial to the user to be able to progress with their task.
+
+{{
+    patternlibExample("components/accordion/examples/accordion/index.njk")
```
boxadesign: Need to update to use `patternlibExample({"path":"components/accordion/examples/accordion/index.njk"})`

bameyrick: Good catch 👍

bameyrick:
[Accordion](https://trello.com/c/8aacxSBp/31-accordion)

jrbarnes9:
It would be useful to explain when to use accordions over tabs, and vice versa.  Happy to push some more design/rationale content for this if you like.

bameyrick:
> versa. Happy to push some more design/rationale content for this if you like.

That would be good!

PR #90, opened 2019-01-18
48 tabs

boxadesign:


src/views/layouts/_master.njk
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en-GB">
+<html lang="en-GB" class="no-js">
```
bameyrick: I don't know whether this `no-js` `has-js` method is that robust. If JavaScript is enabled but downloading of the JavaScript bundle hasn't completed yet/doesn't complete or the component crashes, the user will be presented with UI for a component that doesn't work. 

If you look at what I did for the collapsible component:
https://github.com/ONSdigital/pattern-library-v2/blob/master/src/components/details/src/collapsible.js#L25
that wont change any layout until the component has successfully initialised.

Don't know if changing this is in scope of this PR but I thought it was worth mentioning.

PR #92, opened 2019-01-22
43 metadata

jrbarnes9:


src/components/metadata/src/_metadata.scss
```diff
@@ -0,0 +1,34 @@
...
+      }
+    }
+  }
+
+  dd {
+    margin-left: 0; /* As normalize adds a 40px left margin */
+
+    &:not(:last-child) {
+      @include mq(xs, m) {
+        margin-bottom: 0;
```
boxadesign: I don't think this is doing anything and so probably the `&:not(:last-child) {...}` can be removed. I think the `margin-top` on the `dt` is handling the spacing between sets.

jrbarnes9:
[Metadata](https://trello.com/c/j0G20sAr/43-metadata)

PR #106, opened 2019-01-29
31 accordion documentation

jrbarnes9:
Added more guidance on when/how to use accordions, and aligned it with tabs component regarding 'alternative components'.


jrbarnes9:
[Accordion](https://trello.com/c/8aacxSBp/31-accordion)

PR #107, opened 2019-01-30
47 table

boxadesign:


src/components/table/src/_table.scss
```diff
@@ -0,0 +1,178 @@
+.table {
```
bameyrick: Potentially the child `.table__<thing>` could be `&__<thing>` which will reduce the amount of classnames outputted (e.g. `.table .table__caption {` would just be `.table__caption { `), but also specificity which could affect overriding things. 🤷🏻‍♂️

src/components/table/src/_table.scss
```diff
@@ -0,0 +1,178 @@
...
+      outline: 3px solid $color-focus;
+      outline-offset: 3px;
+    }
+    .table__header,
+    .table__cell {
+      @include mq(xs, m) {
+        white-space: nowrap;
+      }
+    }
+    .right-shadow,
```
bameyrick: Should these classes not be `table__right-shadow` and `table__left-shadow`?

src/components/table/src/_template.njk
```diff
@@ -0,0 +1,38 @@
+{% if params.scrollable %}
```
bameyrick: All of the hardcoded aria-labels etc should probably be parameterised to allow for internationalisation

src/components/table/index.njk
```diff
@@ -0,0 +1,69 @@
...
+- **Output** - This reduces the `font-size` to compact the table for basic tables which have many rows but only a few columns.
+
+### Numeric table
+{{
+    patternlibExample({"path": "components/table/examples/table-numeric/index.njk"})
+}}
+- **Class** - `table__header--numeric` + `table__cell--numeric`
+- **Output** - This aligns the content to the right for tables with all numeric values for best practice.
+
+### Reponsive table
```
jrbarnes9: Typo!

src/components/table/src/_table.scss
boxadesign: Yep you're right.

bameyrick:
@boxadesign I already did?

PR #110, opened 2019-01-31
Language Selector

bameyrick:


bameyrick:
[Language Selector](https://trello.com/c/Br20D5aj/41-language-selector)

will-madebymoon:
Visually looks fine to me! Behave as intended. Small point! Where you've put it in the pattern library should probably be Help users to... Change language

jrbarnes9:
Uneven spacing between footer links...
screenshot 2019-02-01 at 09 06 07

jrbarnes9:
> Visually looks fine to me! Behave as intended. Small point! Where you've put it in the pattern library should probably be Help users to... Change language

I think we need to do more work to understand exactly the definition of a component vs a pattern. I've tried to start that in a [trello card.](https://trello.com/c/FHdwIQpY/90-definitions-of-a-style-component-pattern)  

This probably is, and should be just be a pattern, because essentially is a link or [select](https://ons-design-system.netlify.com/components/select/) component displayed in the header/footer, and we're describing how it functions to change the content language.   However, if there are classes and js functions for these specifically to work for language switching, then they should probably be in a component too.  Or, we could add them as variations to the links/select components?

src/patterns/language-selection/index.njk
```diff
@@ -0,0 +1,42 @@
+---
+title: Change Language
+group: Help users to...
+---
+{% from "views/partials/example/_macro.njk" import patternlibExample %}
+
+## When to use this pattern
+
+This pattern should be if you service is available in multiple languages.
```
jrbarnes9:
```suggestion
This pattern should be used if your service is available in multiple languages.
```

src/patterns/language-selection/index.njk
```diff
@@ -0,0 +1,42 @@
...
+
+## How to use this pattern
+
+Language selection is achieved by rendering a language selector component in the header element, and a list of available lanauges in the footer. The header lanauge selector is for viewports larger than mobile, and the footer language selector is for viewports smaller than tablet.
```
jrbarnes9:
```suggestion
Language selection is achieved by rendering a language selector component in the header element, and a list of available languages in the footer. The header language selector is for viewports larger than mobile, and the footer language selector is for viewports smaller than tablet.
```
...but perhaps this shouldn't refer to a 'component' unless we decide to create it as a component.

src/patterns/language-selection/index.njk
```diff
@@ -0,0 +1,42 @@
...
+
+If there are 3 or less available languages, the header language picker will render two links, otherwise a dropdown menu will be renedered.
```
jrbarnes9:
```suggestion
If there are 3 or less available languages, the header language picker will render one or two links, otherwise a select component will be rendered.
```

src/patterns/language-selection/index.njk
```diff
@@ -0,0 +1,42 @@
...
+    patternlibExample({ "path": "patterns/language-selection/examples/english-ireland/index.njk", "useIframe": true, "iframeHeight": 400 })
+}}
+
+### 2 Available languages
+{{
+    patternlibExample({ "path": "patterns/language-selection/examples/cymraeg/index.njk", "useIframe": true, "iframeHeight": 400 })
+}}
+
+### 5 Available languages
+{{
```
jrbarnes9: What happens when the user selects 'all languages'?

boxadesign:
> > Visually looks fine to me! Behave as intended. Small point! Where you've put it in the pattern library should probably be Help users to... Change language
> 
> I think we need to do more work to understand exactly the definition of a component vs a pattern. I've tried to start that in a [trello card.](https://trello.com/c/FHdwIQpY/90-definitions-of-a-style-component-pattern)
> 
> This probably is, and should be just be a pattern, because essentially it is just a link or [select](https://ons-design-system.netlify.com/components/select/) component displayed in the header/footer, and we're providing guidance and describing how it functions to change the content language. However, if there are classes and js functions for these specifically to work for language switching, then they should probably be in a component too. Or, we could add them as variations to the links/select components?

I think this is a component and part of a pattern. If the pattern is to "Help users to select a language" it would include the pattern for asking for many languages i.e. a page listing all the other languages.

src/patterns/language-selection/index.njk
bameyrick: So I did originally have a page documenting the language selector component, but it was strange because that only really takes care of the desktop switcher, there isn't a component for the footer language selector. The language selector dropdown should only ever show in the header, and is imported by the header so I'm not really sure whether we should be mentioning it as a component, as really it's a feature of the header. It only has a component folder to stop the header component from getting too complex.

src/patterns/language-selection/index.njk
bameyrick: It would link to a separate page with a list of all languages, although we won't be using that for a while. If the all languages link isn't supplied it shouldn't render the all languages option so maybe I should do that in all the examples until we take on loads of languages.

src/patterns/language-selection/index.njk
jrbarnes9: Maybe it could be documented as a variant of the header, which we could reference here?  or a variant of the select component?

src/patterns/language-selection/index.njk
jrbarnes9: Yeah leave it off the example for now then.  I'm just trying to think of questions designers might ask when reading this.

src/patterns/language-selection/index.njk
jrbarnes9: @will-madebymoon Should we align the styles of these? 
screenshot 2019-02-01 at 10 09 29
screenshot 2019-02-01 at 10 09 14

jrbarnes9: Agreed.

will-madebymoon: Whilst it isn't a select component it should be consistent.

will-madebymoon:
> Uneven spacing between footer links...
> screenshot 2019-02-01 at 09 06 07

Good spot I didn't check the mobile behaviour

src/patterns/language-selection/index.njk
will-madebymoon: Not sure how it ended up light grey with a blue arrow
screen shot 2019-02-01 at 11 15 56

will-madebymoon: It should have the same states as these https://deploy-preview-110--ons-design-system.netlify.com/components/select/

bameyrick: I've now updated the styling to match the select component https://deploy-preview-110--ons-design-system.netlify.com/patterns/language-selection/examples/many-languages/. @will-madebymoon are the menu item heights ok? I've made them match the height of the "button" that opens the menu. Only thing is that it makes the white part of the header taller than it would be without the dropdown variant of the language selector - 65px vs 49px.

src/patterns/language-selection/index.njk
bameyrick: I've removed the all languages options for now... we can add them in at a later point when we take on more than say 5 languages

bameyrick:
> > Uneven spacing between footer links...
> > screenshot 2019-02-01 at 09 06 07
> 
> Good spot I didn't check the mobile behaviour

This should be resolved now

will-madebymoon:
One small point. On hover the select changes colour, we should use this behaviour as well.
screen shot 2019-02-01 at 12 04 37

armstrongb:
**5** languages example only displays 4 options but English is one of them.
**3** languages example displays 2 options (presumably as the example is Eng + 2?)
**2** languages is Welsh + 1?

src/patterns/language-selection/index.njk
bameyrick: After discussion with @will-madebymoon I've dropped the dropdown version of the language selector's height down to 36px to match the sketch file, and adjusted the header css to prevent it from getting taller when the dropdown variant is added.

PR #120, opened 2019-02-01
Fix pattern library overflow layout

### LINKED ISSUE #119: Flexbox layout doesn't work in firefox or latest chrome
**Describe the bug**
The header should be static, and then the main content area and side navigation should be individually scrollable. Currently the whole body scrolls. Likely an issue with flexbox.

bameyrick:
Resolves #119


bameyrick:
[Issue #119: Flexbox layout doesn't work in firefox or latest chrome](https://trello.com/c/2NiyFkRD/123-issue-119-flexbox-layout-doesnt-work-in-firefox-or-latest-chrome)

PR #122, opened 2019-02-01
Allow .html examples to be rendered, but not show nunjucks tab in example component

bameyrick:


bameyrick:
[Fix broken example paths in typography](https://trello.com/c/Gs1u5cfH/106-fix-broken-example-paths-in-typography)

PR #123, opened 2019-02-01
Adjust out patternlib spacing

jrbarnes9:


jrbarnes9:
[Sort out patternlib spacing](https://trello.com/c/UZVCP7on/110-sort-out-patternlib-spacing)

jrbarnes9:
[Sort out patternlib spacing](https://trello.com/c/UZVCP7on/110-sort-out-patternlib-spacing)

src/scss/patternlib.scss
```diff
@@ -16,6 +16,27 @@
   &__body {
     display: flex;
     flex: 1;
+
```
bameyrick: This will potentially affect headings in the example components as well. The selectors may need to be a bit more specific to prevent this

jrbarnes9: @bameyrick I've amended this now, and moved typography headings into an example.

PR #125, opened 2019-02-04
Ensure prefixed/suffixed inputs are accessible

bameyrick:
When focusing on a prefixed/suffixed input before there was no mention of the contents of the `<abbr>` tag. I've tacked this by adding a `title` attribute to the input with the title value form the `<abbr>` tag, and added `aria-hidden="true"` to the `<abbr>` to prevent the screen reader from focusing to it as the user moves to the next item. Combined with a `<label>` for the input this seems like the best experience, although its worth everyone having a go at testing this themselves to see what they think.

### Components to test

- [x] [Prefixed](https://5c58275229ef6d000b419778--ons-design-system.netlify.com/components/input/examples/prefixed/)
- [x] [Suffixed](https://5c58275229ef6d000b419778--ons-design-system.netlify.com/components/input/examples/suffixed/)
- [x] [Durations (example of suffixed in a `<fieldset>`)](https://5c58275229ef6d000b419778--ons-design-system.netlify.com/patterns/durations/examples/duration-address/)


bameyrick:
[Prefixed/Suffixed inputs are not fully accessible](https://trello.com/c/YU9nhNCC/117-prefixed-suffixed-inputs-are-not-fully-accessible)

PR #127, opened 2019-02-04
Ensure u-vh text is not selectable by the user

bameyrick:


bameyrick:
[Visually hidden text should not be highlightable and selectable by the user](https://trello.com/c/emJvTls5/116-visually-hidden-text-should-not-be-highlightable-and-selectable-by-the-user)

PR #128, opened 2019-02-04
No right align on suffixed fields

bameyrick:


bameyrick:
[Suffixed inputs should only have text aligned to the right if the width is constrained](https://trello.com/c/WmEzFXy2/118-suffixed-inputs-should-only-have-text-aligned-to-the-right-if-the-width-is-constrained)

PR #143, opened 2019-02-08
Create npm package of built CSS, JS, and macros

bameyrick:
This change will make Travis build and deploy a bundle of built JS, CSS, and a collection of macros to npm when a release is created. It will also create a commit and push the release tag as the version parameter in the package.json on the master branch.


bameyrick:
[Create npm package of built CSS/JS and macros](https://trello.com/c/SJOs36qx/129-create-npm-package-of-built-css-js-and-macros)

PR #145, opened 2019-02-08
Move templates and macros out of src so that references are correct in npm bundle

bameyrick:


bameyrick:
[Create npm package of built CSS/JS and macros](https://trello.com/c/SJOs36qx/129-create-npm-package-of-built-css-js-and-macros)

PR #157, opened 2019-02-12
Fix SVGs, move all non production images to a different folder, and add documentation about exporting SVGs correctly

### LINKED ISSUE #155: Grid SVG is broken
**Describe the bug**
Some of the SVGs on the grid page have broken since adding SVGO optimisation.

**Screenshots**
screenshot 2019-02-11 at 16 44 11

bameyrick:
Will resolve #155.

I think Sketch exports SVGs with inlined styles so I haven't documented anything about this but correct me if I am wrong.


src/get-started/svgs/index.njk
```diff
@@ -0,0 +1,20 @@
+---
+title: Exporting SVGs
+sortOrder: 3
+---
+
+When exporting SVG images for the web you should ensure that styles are exported as *inline* rather than as CSS. This is to ensure that styles are not accidentally overridens as the generated class names are not unique, and allows [SVGO](https://github.com/svg/svgo) to optimise the file without breaking it.
```
jrbarnes9:
```suggestion
When exporting SVG images for the web you should ensure that styles are exported as *inline* rather than as CSS. This is to ensure that styles are not accidentally overridden as the generated class names are not unique, and allows [SVGO](https://github.com/svg/svgo) to optimise the file without breaking it.
```

PR #159, opened 2019-02-12
Allow field legend to accept HTML

bameyrick:


bameyrick:
[Make field legend parameter accept HTML](https://trello.com/c/CxBjVJB3/143-make-field-legend-parameter-accept-html)

PR #170, opened 2019-02-14
151 issue 169 accordion missing id and classes parameters

### LINKED ISSUE #169: Accordion missing ID and classes parameters
No way to pass classes or an ID to the accordion via the macro.

jrbarnes9:
Resolves #169

### DIFF
--- a/src/components/accordion/_template.njk
+++ b/src/components/accordion/_template.njk
@@ -1,6 +1,6 @@
 {% from "components/details/_macro.njk" import onsDetails %}
 
-<div class="accordion">
+<div id="{{params.id}}" class="accordion {{ params.classes }}">
     {% if params.openAll and params.closeAll %}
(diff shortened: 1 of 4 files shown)


src/components/accordion/_template.njk
```diff
@@ -1,6 +1,6 @@
 {% from "components/details/_macro.njk" import onsDetails %}
 
-<div class="accordion">
+<div id="{{params.id}}" class="accordion {{ params.classes }}">
```
bameyrick: Could add an if to avoid whitespace when no classes are passed
```suggestion
<div id="{{params.id}}" class="accordion{{ ' ' + params.classes if params.classes }}">
```

jrbarnes9:
[Issue #169: Accordion missing ID and classes parameters](https://trello.com/c/K5yNr7qW/151-issue-169-accordion-missing-id-and-classes-parameters)

PR #179, opened 2019-02-15
177 fire customEvent from domready

### LINKED ISSUE #177: Fire customEvent from domready so that prototype kits can hook into it
Fire a [`customEvent`](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) from the `onReady` function of `domready.js` so that the prototype kit can listen for that event and delay booting of components until the dom is ready and polyfills have loaded.

bameyrick:
Added a customEvent (`onsDOMReady`) and window variable (`window.onsDOMReady`) so that other other JavaScript bundles can hook into the domready event.

Will resolve #177


bameyrick:
[Issue #177: Fire customEvent from domready so that prototype kits can hook into it](https://trello.com/c/kt0eruOG/153-issue-177-fire-customevent-from-domready-so-that-prototype-kits-can-hook-into-it)

PR #193, opened 2019-02-19
Move `patternlib-example__link` out of frame so it doesn't block logo

jrbarnes9:


jrbarnes9:
[Move `patternlib-example__link` out of frame so it doesn't block logo](https://trello.com/c/LftJMHLw/152-move-patternlib-examplelink-out-of-frame-so-it-doesnt-block-logo)

src/views/partials/example/_example.scss
```diff
@@ -9,18 +9,15 @@
...
   &__link {
-    position: absolute;
-    z-index: 1;
-    top: 0;
-    left: 0;
     display: inline-block;
     padding: 0.25rem 0.5rem;
     line-height: 1.3rem;
-    background: $color-grey-4;
+    background: #fff;
```
bameyrick: Should probably a variable
```suggestion
    background: $color-white;
```

PR #196, opened 2019-02-19
100 page template

boxadesign:


bameyrick:
I think some documentation about the different blocks and how they can be overrided, along with a graphic showing the blocks and their nesting (like on [GDS](https://design-system.service.gov.uk/styles/page-template/)) would be good. I can assist with this if you want?

boxadesign:
> I think some documentation about the different blocks and how they can be overrided, along with a graphic showing the blocks and their nesting (like on [GDS](https://design-system.service.gov.uk/styles/page-template/)) would be good. I can assist with this if you want?

As per our chat yesterday I have left out the Nunjucks side of things and kept it to being a HTML only doc page (for now).

PR #209, opened 2019-02-20
Set checkbox and radio right padding to 1rem

### LINKED ISSUE #207: When radios or checkboxes are greater than the standard length the right padding should be 1rem from the label.
When radios or checkboxes are greater than the standard length the right padding should be 1rem from the label.

_Originally posted by @will-madebymoon in https://github.com/ONSdigital/pattern-library-v2/issues/56#issuecomment-465548500_

bameyrick:
Will resolve #207

### DIFF
--- a/src/components/checkboxes/_checkbox.scss
+++ b/src/components/checkboxes/_checkbox.scss
@@ -44,7 +44,7 @@ $checkbox-label-vertical-padding: 0.75rem;
 
   &__label {
     display: block;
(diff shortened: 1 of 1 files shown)


bameyrick:
[Issue #207: When radios or checkboxes are greater than the standard length the right padding should be 1rem from the label.](https://trello.com/c/uf1C4cwb/166-issue-207-when-radios-or-checkboxes-are-greater-than-the-standard-length-the-right-padding-should-be-1rem-from-the-label)

PR #216, opened 2019-02-21
Fix field margins

### LINKED ISSUE #215: Field item margins
The last `field__item` or the last nested field shouldn't have a `margin-bottom` if it is not wrapped in `field__group`.

Also, `field__group` should have a `margin-bottom` of -1 * the `margin-bottom` of the inner `field__items`.

_Originally posted by @bameyrick in https://github.com/ONSdigital/pattern-library-v2/issues/35#issuecomment-466001706_

bameyrick:
Will resolve #215

### DIFF
--- a/src/components/field/_field.scss
+++ b/src/components/field/_field.scss
@@ -17,6 +17,7 @@
 
   &__group {
     @extend .u-cf;
+    margin-bottom: -1rem;
   }
 
   &__item {
@@ -27,6 +28,10 @@
   & > &__item {
     margin: 0 0.5rem 0.5rem 0;
     width: 100%;
+
+    &:last-of-type {
+      margin-bottom: 0;
+    }
 
     @include mq(m) {
       &:not([class*='field__item--w-']) {


bameyrick:
[Issue #215: Field item margins](https://trello.com/c/ALp7K07v/171-issue-215-field-item-margins)

src/components/field/_field.scss
```diff
@@ -17,6 +17,7 @@
 
   &__group {
     @extend .u-cf;
+    margin-bottom: -1rem;
```
jrbarnes9: The clearfix above already seems to being doing this job?  There are now spacing issues caused by this negative margin...
screenshot 2019-02-21 at 14 31 35

bameyrick: Hmm the clearfix didn't appear to have the effect of preventing the bottom margin on the date input. I'll take a deeper look into it.

bameyrick: Having had a look this behaviour is correct. Currently field doesn't actually apply any spacing between itself and another field (see https://sdc-prototypes.netlify.com/prototypes/example/ for an example). In runner this is achieved by wrapping each field in a custom `answer` class. 

We should maybe discuss the this margin being applied to field themselves outside of this PR.

bameyrick: I've began a discussion about this on the forum issue for field: https://github.com/ONSdigital/pattern-library-v2/issues/35#issuecomment-466045681

PR #224, opened 2019-02-22
Add margin top to a field that is preceded by another field

### LINKED ISSUE #223: Fields should apply their own margins
Having looked at the layout of a question form, and following discussions with @will-madebymoon and @jrbarnes9, we should probably move the styling for spacing between fields to the fields themselves rather than from `_question.scss`.

_Originally posted by @bameyrick in https://github.com/ONSdigital/pattern-library-v2/issues/35#issuecomment-466045681_

bameyrick:
Will resolve #223

### DIFF
--- a/src/components/field/_field.scss
+++ b/src/components/field/_field.scss
@@ -1,6 +1,10 @@
 .field {
   background: none;
 
+  & + & {
+    margin-top: 1rem;
+  }
+
   &__legend {
     margin-bottom: 0.5rem;
     font-size: 1rem;


bameyrick:
[ Fields should apply their own margins](https://trello.com/c/YKi3pJWO/178-fields-should-apply-their-own-margins)

bameyrick:
[ Fields should apply their own margins](https://trello.com/c/YKi3pJWO/178-fields-should-apply-their-own-margins)

PR #241, opened 2019-02-25
62 provide feedback on a service

bameyrick:


bameyrick:
[Provide feedback on a service](https://trello.com/c/YMbyjo7a/62-provide-feedback-on-a-service)

bameyrick:
@jrbarnes9 I've not been able to fix the iframe resizer issue, this appears to be an issue with the library itself. 

I've addressed the column issue, and I'll create examples of the page template with the feedback in when I go into more detail documenting the page template nunjucks template and macro options.

boxadesign:
@bameyrick @jrbarnes9 I've created an issue for the iframe resize problem and will take a look.

https://github.com/ONSdigital/design-system/issues/253

PR #245, opened 2019-02-26
190 remove timeout component

bameyrick:
Removing timeout component and all dependencies as discussed with @will-madebymoon and @ajmaddaford as this will not be used in runner v3.


bameyrick:
[Remove Timeout Component](https://trello.com/c/vVgxjogK/190-remove-timeout-component)

PR #246, opened 2019-02-26
189 missing params in page template footer

jrbarnes9:


jrbarnes9:
[ Missing params in page-template footer](https://trello.com/c/hI6htclp/189-missing-params-in-page-template-footer)

PR #257, opened 2019-02-27
196 input width modifier doesnt work correctly when used in a fielditem

### LINKED ISSUE #255: Input width modifier doesn't work correctly when used in a `field__item`
Input width modifier doesn't work correctly when used in a `field__item`, for example the date component. Inputs will only grow as wide as their labels.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/163#issuecomment-467866623_

bameyrick:
Will resolve #255


bameyrick:
[ Input width modifier doesn't work correctly when used in a `field__item`](https://trello.com/c/qcb8Wzbu/196-input-width-modifier-doesnt-work-correctly-when-used-in-a-fielditem)

PR #263, opened 2019-02-28
Improve mutually exclusive screen reader alerts

### LINKED ISSUE #270: Vertical rhythm between a field of radios or select boxes isn't right
**Describe the bug**
The vertical rhythm of these boxes doesn't feel right. It should be 0.5 rem. See screen shot  

screenshot 2019-03-01 at 15 17 45
screenshot 2019-03-01 at 15 17 12

### LINKED ISSUE #271: Select all that apply cannot be a label or legend
Select all that apply cannot be a label or legend. 
In runner we include 'Select all that apply' as default for all checkbox lists. 
The legend would be Pizza toppings or Passports
The styling is also incorrect. See screenshot

screenshot 2019-03-01 at 15 17 12
screenshot 2019-03-01 at 15 17 45

bameyrick:
I've updated the mutually exclusive component to improve the screen reader alerts for when inputs/checkboxes are programatically cleared.

It would be worth having a play with a few of the different mutually exclusive example types using a screen reader to see how the alerts sound. One to be aware of is the "please specify" input in the "Other" option in mutually exclusive checkboxes, although that may be an issue with the "other" checkbox/radio option itself rather than the mutually exclusive code.

Will resolve #270 
Will resolve #271


bameyrick:
[Mutually exclusive aria-live isn't very good](https://trello.com/c/Gjy18SEs/52-mutually-exclusive-aria-live-isnt-very-good)

jrbarnes9:
[ Vertical rhythm between a field of radios or select boxes isn't right](https://trello.com/c/YBpER78l/207-vertical-rhythm-between-a-field-of-radios-or-select-boxes-isnt-right)

bameyrick:
@jrbarnes9 I've changed it to only say "selecting this will uncheck all other checkboxes" if checkboxes have been checked.

I get an alert for "no central heating deselected" when i check another options so it could be a device specific issue... what are you testing with?

As for the other field, I think we will have to address that as a fix for checkboxes and radios as they have no link between other and the input field it shows at the moment. I've raised an issue here: #272

jrbarnes9:
> @jrbarnes9 I've changed it to only say "selecting this will uncheck all other checkboxes" if checkboxes have been checked.
> 
> I get an alert for "no central heating deselected" when i check another options so it could be a device specific issue... what are you testing with?
> 
> As for the other field, I think we will have to address that as a fix for checkboxes and radios as they have no link between other and the input field it shows at the moment. I've raised an issue here: #272

1. Good, this works well now.
2. Seems fine now, perhaps I missed it first time round, was using chrome with mac voiceover.
3. Sure, sounds makes sense.

jrbarnes9:
When using the feedback text area example, the character count is read out after the 'I don't want to provide feedback' checkbox is selected.

jrbarnes9:
Still a spacing issue:...
screenshot 2019-03-04 at 14 28 47

jrbarnes9:
Mutually exclusive duration... "[legend] deselected" is read out twice after selecting the checkbox.  One for each field cleared presumably.  Preferably should just read this out once to indicate all fields in the field_group have been deselected.

bameyrick:
[ Select all that apply cannot be a label or legend](https://trello.com/c/AqooeVap/206-select-all-that-apply-cannot-be-a-label-or-legend)

bameyrick:
> Mutually exclusive duration... "[legend] deselected" is read out twice after selecting the checkbox. One for each field cleared presumably. Preferably should just read this out once to indicate all fields in the field_group have been deselected.

I've updated this to read out the title of the abbreviation so it should read something like "Years cleared. Months cleared" rather than the [legend] deselected. I'm not sure whether saying "[legend] cleared" and not mentioning which fields were cleared in the group 'e.g. years and months here, or day, month, year in the date input' would be clear enough for screen reader users.

src/components/mutually-exclusive/examples/textarea/index.njk
```diff
@@ -6,22 +6,22 @@
         "name": "feedback",
         "classes": "input--w-30",
         "label": {
-            "text": "Please provide some feedback",
-            "description": "For example describe any difficulties you experienced in the use of this service"
+            "text": "Why did you leave you last job?"
```
jrbarnes9:
```suggestion
            "text": "Why did you leave your last job?"
```

PR #265, opened 2019-03-01
Moved the date input component to a pattern and updated guidance

jrbarnes9:


jrbarnes9:
[Dates](https://trello.com/c/aLDI1yLx/134-dates)

bameyrick:
Could you also update the design forum issue to be pattern rather than component as part of this too please?

jrbarnes9:
> Could you also update the design forum issue to be pattern rather than component as part of this too please?

Think I did?

bameyrick:
Oh yes you did.... I was too lazy to actually check...

PR #268, opened 2019-03-01
201 inputs with type number causing validation errors

### LINKED ISSUE #261: Inputs with type "number" causing validation errors
**Describe the bug**
Inputs with type "number" causing validation errors in some browsers. Users can still input letters and this will cause the form to post with no value for this field.

We still need to be able to specify a number type field to bring up a number keyboard on iOS and Android.

**Desktop**
- Browser: Firefox

bameyrick:
Will resolve #261

### DIFF
--- a/src/components/input/_template.njk
+++ b/src/components/input/_template.njk
@@ -2,8 +2,11 @@
 {% from "components/label/_macro.njk" import onsLabel %}
 
 {% if params.type == "number" %}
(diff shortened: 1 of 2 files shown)


bameyrick:
[ Inputs with type "number" causing validation errors](https://trello.com/c/kDULDJTI/201-inputs-with-type-number-causing-validation-errors)

PR #286, opened 2019-03-15
Fix focus/click race condition on language dropdown

### LINKED ISSUE #285: Language dropdown focus click race condition
If you haven't focused the dropdown version and then long click, the menu will open from focusing but  it will then register the click event when the mouse is released and close the menu.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/109#issuecomment-472906469_

bameyrick:
Will resolve #285

### DIFF
--- a/src/components/language-selector/language.js
+++ b/src/components/language-selector/language.js
@@ -11,8 +11,8 @@ export class LanguageSelector {
     this.throttledSetOpen = throttle(this.setOpen.bind(this), 200);
     this.setAriaAttributes();
     this.button.addEventListener('mousedown', this.toggle.bind(this));
-    this.button.addEventListener('focus', () => this.throttledSetOpen(true));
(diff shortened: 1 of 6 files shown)


bameyrick:
[ Language dropdown focus click race condition](https://trello.com/c/ebYnkymp/214-language-dropdown-focus-click-race-condition)

boxadesign:
@bameyrick When I click each language it just seems to hide the menu and doesn't execute the link.

bameyrick:
@boxadesign in this example thats correct, as the hrefs are just #... It should work if there were actually language links in there. I could update the example to link to the same url but with a query string lang value?

boxadesign:
@bameyrick Yeah I think that would be good just to test that the url updates.

bameyrick:
@boxadesign I've updated the example 👍

PR #287, opened 2019-03-15
Remove label--small variant

### LINKED ISSUE #284: We should remove the `label--small` variant of a label as this should never be used.
We should remove the `label--small` variant as this should never be used.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/40#issuecomment-472380576_

bameyrick:
Will resolve #284

### DIFF
--- a/src/components/label/_label.scss
+++ b/src/components/label/_label.scss
@@ -9,8 +9,4 @@
     display: inline-block;
     line-height: 1.4;
   }
-
-  &--small {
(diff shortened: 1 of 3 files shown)


bameyrick:
[ We should remove the `label--small` variant of a label as this should never be used.](https://trello.com/c/Fs8Sc7p0/213-we-should-remove-the-label-small-variant-of-a-label-as-this-should-never-be-used)

PR #288, opened 2019-03-15
Ensure panel link colour is the same as all other links

### LINKED ISSUE #282: Links within panels should not have their colour modified. Links should always be the same blue everywhere.
Links within panels should not have their colour modified. Links should always be the same blue everywhere.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/49#issuecomment-472043674_

bameyrick:
Will resolve #282

### DIFF
--- a/src/components/panel/_panel.scss
+++ b/src/components/panel/_panel.scss
@@ -17,9 +17,7 @@
 
 .panel {
   border-radius: 0;
-  a {
-    color: $color-dark-blue;
-  }
+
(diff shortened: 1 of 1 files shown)


bameyrick:
[ Links within panels should not have their colour modified. Links should always be the same blue everywhere.](https://trello.com/c/0t4qJ15T/211-links-within-panels-should-not-have-their-colour-modified-links-should-always-be-the-same-blue-everywhere)

PR #290, opened 2019-03-15
Don't run iOS 9 mobile tests as they are continually timing out and causing all builds to fail

### LINKED ISSUE #289: iOS 9 tests timing out
**Describe the bug**
The iOS 9 browserstack tests are always timing out causing all builds to fail

bameyrick:
Will resolve #289


bameyrick:
[ iOS 9 tests timing out](https://trello.com/c/uATKoTCj/215-ios-9-tests-timing-out)

PR #291, opened 2019-03-15
Fix errors prototype

### LINKED ISSUE #281: Buttons for error prototype don't work now that we removed ability to render links as buttons

bameyrick:
Will resolve #281


bameyrick:
[ Buttons for error prototype don't work now that we removed ability to render links as buttons](https://trello.com/c/JCNlIqwC/212-buttons-for-error-prototype-dont-work-now-that-we-removed-ability-to-render-links-as-buttons)

PR #300, opened 2019-04-02
Make header title wrapping tag a div by default with option to change to a h1

### LINKED ISSUE #295: Design system header title should not be a h1
**Describe the bug**
The design system header title should not be a h1 as every page has its own h1 in the main

bameyrick:
Will resolve #295


bameyrick:
[ Design system header title should not be a h1](https://trello.com/c/78Ort9S0/219-design-system-header-title-should-not-be-a-h1)

PR #308, opened 2019-04-15
Added ability to add theme settings and a census output

boxadesign:
Main features:

1) Created a census.scss manifest file.
2) Webpack bundles and outputs a census.css file.
3) Created 'settings/_census.scss' for any theme related styles.
4) Added `!default` to variables and typography in existing files.


bameyrick:
Looks good, only suggestion I have is to put everything but settings into one file (maybe something like `_core.scss`, and then have `main.scss` import the default settings and `_core.scss`, and then `census.scss` to import the census settings and `_core.scss`... keeps it dry and means if anything else is changed or added its only done in one file.

bameyrick:
Oh also, if you could search for anywhere where `responsive.css` was referenced and change that to `main.css`, because I think you might have missed some templates.

bameyrick:
Looks good, I think you still need to change the `responsive.css` reference in  `src/styles/page-template/_template.njk`

PR #310, opened 2019-04-16
Ensure NVDA can read labels

### LINKED ISSUE #309: Labels on checkboxes and radios not reading labels in NVDA
**Describe the bug**
Some work was done to prevent the labels for checkboxes and radios to be individually selectable using a screen reader, however NVDA is not reading the `aria-labeledby` attribute effectively meaning checkboxes and radios do not have a label.

This affects #13 and #56.

bameyrick:
Removed attributes that prevented screen readers from being able to focus on just the label for checkboxes/radios as NVDA was not working with the `aria-labeledby` attribute.

Will resolve #309


bameyrick:
[ Labels on checkboxes and radios not reading labels in NVDA](https://trello.com/c/KYcJdyxw/223-labels-on-checkboxes-and-radios-not-reading-labels-in-nvda)

PR #318, opened 2019-04-23
Remove dependency on no-js and has-js classes

### LINKED ISSUE #316: Remove dependencies on has-js classes
We should remove ability to style components using `has-js`, `no-js`, `u-no-js-hide`, and `u-no-js-show` classes.

The `has-js` class is set on document load, which doesn't guarantee that the component relying on that for styling has booted yet or even booted correctly. Also the main JS bundle may have failed to load, leaving the user with broken UI.

If a component relies on JavaScript its appearance and classes should be modified by itself rather than external sources.

bameyrick:
Will resolve #316


bameyrick:
[ Remove dependencies on has-js classes](https://trello.com/c/vcKISojN/227-remove-dependencies-on-has-js-classes)

PR #344, opened 2019-05-02
Correct secondary button colours and reorder css attributes to satisfy linter

### LINKED ISSUE #343: Secondary button background colour is incorrect
After speaking with @will-madebymoon we noticed that the secondary button background colour should be `#dedede`, and the text colour should be `#222222` with regular font weight.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/232#issuecomment-488308575_

bameyrick:
Will resolve #343


bameyrick:
[ Secondary button background colour is incorrect](https://trello.com/c/D8PGdRXk/261-secondary-button-background-colour-is-incorrect)

bameyrick:
> I think the print icon should be made black to match the text colour. Can be raised as an issue and dealt with separately if required.

@will-madebymoon did mention this and I forgot to do it so I'll update that now.

PR #346, opened 2019-05-03
Modified summary component to handle household summary and simplified CSS

bameyrick:


bameyrick:
[Create list pattern for adding householders within a questionnaire](https://trello.com/c/6E7wI94c/246-create-list-pattern-for-adding-householders-within-a-questionnaire)

PR #348, opened 2019-05-03
Added Element.closest polyfill

### LINKED ISSUE #347: Element.closest polyfill required for IE11
IE11 doesn't natively support Element.closest, so this should be added to the conditional polyfiller.

bameyrick:
Will resolve #347


bameyrick:
[ Element.closest polyfill required for IE11](https://trello.com/c/atspdpUI/262-elementclosest-polyfill-required-for-ie11)

PR #356, opened 2019-05-09
Fix summary values condition for jinja2

### LINKED ISSUE #355: Summary rowItem has values check does not work in jinja2
The condition that checks if a rowItem has values does not work in jinja2

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/84#issuecomment-490793386_

bameyrick:
Will resolve #355

### DIFF
--- a/src/components/summary/_macro-options.md
+++ b/src/components/summary/_macro-options.md
@@ -17,12 +17,12 @@
 
 ## SummaryRowItem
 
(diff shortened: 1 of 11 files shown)


bameyrick:
[ Summary rowItem has values check does not work in jinja2](https://trello.com/c/2xKLjYfA/271-summary-rowitem-has-values-check-does-not-work-in-jinja2)

PR #361, opened 2019-05-09
Make concourse alert in slack when a CDN deployment is successful

bameyrick:


bameyrick:
[Have concourse send slack alerts when builds are successful](https://trello.com/c/8O4sH4L7/274-have-concourse-send-slack-alerts-when-builds-are-successful)

PR #368, opened 2019-05-14
Fix size of small button variant

### LINKED ISSUE #364: Small button variant is too tall
Small button variant is too tall since underline change has gone in. Its currently 45px tall where as it should be 35px tall.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/45#issuecomment-491809531_

bameyrick:
Will resolve #364


bameyrick:
[Small button variant is too tall](https://trello.com/c/lQFjY3kb/276-small-button-variant-is-too-tall)

PR #369, opened 2019-05-14
Correct details and accordion button types

### LINKED ISSUE #363: Buttons used in details components are type submit. These should be type button.
Buttons used in details components are type submit. These should be type button.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/148#issuecomment-491767036_

bameyrick:
Will resolve #363

### DIFF
--- a/src/components/accordion/_macro.njk
+++ b/src/components/accordion/_macro.njk
@@ -13,6 +13,7 @@
             
             {{
                 onsButton({
+                    "type": "button",
                     "text": params.allButton.open,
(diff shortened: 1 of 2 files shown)


bameyrick:
[ Buttons used in details components are type submit. These should be type button.](https://trello.com/c/joDtyQt0/275-buttons-used-in-details-components-are-type-submit-these-should-be-type-button)

boxadesign:
@bameyrick I've approved as it doesn't cause an issue but I think the `type=button` on a `<button>` is only required when it's in a form, so it doesn't submit it.

PR #373, opened 2019-05-15
The h3 font size on cards is too large for long titles

### LINKED ISSUE #365: The h3 font size on cards is too large for long titles
The h3 font size on cards is too large for long titles 
Screenshot 2019-05-13 at 14 21 54

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/48#issuecomment-491820017_

jrbarnes9:
Reduced the heading font size on cards component by changing the font size utility class to ```u-fs-m```

Fixes #365

### DIFF
--- a/src/components/card/_macro-options.md
+++ b/src/components/card/_macro-options.md
@@ -1,5 +1,6 @@
-| Name  | Type   | Required | Description                    |
-| ----- | ------ | -------- | ------------------------------ |
-| url   | string | false    | Will wrap the text in a link   |
-| title | string | true     | The title for the card element |
-| text  | string | true     | The text for the card element  |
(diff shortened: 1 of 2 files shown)


jrbarnes9:
[The h3 font size on cards is too large for long titles](https://trello.com/c/DImjFnvi/278-the-h3-font-size-on-cards-is-too-large-for-long-titles)

jrbarnes9:
> Would it not be better if there was a default for the title font-size class but this could be overridden in params?

I don't think this need to be configurable.  Otherwise cards will just be inconsistent everywhere they're used.  I think we just need a size that looks right in terms of typographical hierarchy.

PR #375, opened 2019-05-15
Set default font sizes for headings

### LINKED ISSUE #371: Heading default font sizes too big
Currently the default font sizes for headings are always one step larger than we actually use by default, meaning everywhere we use a heading we have to use a font size utility class. We should change the default font sizes for headings to what we set them to with utility classes, and then remove utility classes from the headings.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/233#issuecomment-492556265_

bameyrick:
Will resolve #371


bameyrick:
[Heading default font sizes too big](https://trello.com/c/Y8YlDcdf/279-heading-default-font-sizes-too-big)

PR #380, opened 2019-05-16
Collapsible modules and components refactor

saedwards:
How to test:
- Run 'yarn start' to check behaviour from user level
- Run 'yarn test' to check functional tests


boxadesign:
The accordion refactor looks good and appears to have reverted back to a previous approach?

I'm happy to educated on this but the levels of abstraction in this seem to add complexity to creating components/modules. I guess my question is, what benefit is there to this approach? Would this make it harder for a developer work with?

saedwards:
> The accordion refactor looks good and appears to have reverted back to a previous approach?
> 
> I'm happy to be educated on this but the levels of abstraction in this seem to add complexity to creating components/modules. I guess my question is, what benefit is there to this approach? Would this make it harder for a developer work with?

There are no design changes as part of this PR, only a bug fix to mend an issue with the Show/Hide all toggle.

The intention was to make it easier to create components and to start unit testing. The new AbstractComponent creates a scope for selecting DOM elements within its realm of responsibility only, a problem that has caused many bugs in the past.

The separation of modules & components to concisely define their responsibilities for single collapsibles and collapsible groups should make it clearer for other developers to understand code semantics and isolate code specifically for unit testing. This also introduces a uniform way of accessing the DOM rather than inconsistently using getElementsByClassName & querySelector reducing the potential for browser inconsistencies. There are also plans to start using TypeScript to further enhance this. Ben can give an update on this work.

Most simple components can be initialised from a module with a single line of code to attached all behaviours to DOM elements: `dom.getElements('js-example-class-name').forEach(ExampleComponent.create)`. I can walk you through it if you come and grab me.

.prettierrc.yml
```diff
@@ -1,2 +1,2 @@
 singleQuote: true
-printWidth: 140
```
bameyrick: Could you put this back to 140 please

bameyrick:
[collapsible.js](https://trello.com/c/gK6sBrmv/237-collapsiblejs)

bameyrick:
Codacy seems to have some issues, otherwise looks good to me

bameyrick:
Github isn't allowing me to merge the PR as shane's commits weren't signed with a GPG key. I will have to look into how to fix this

PR #390, opened 2019-05-24
Fix mutually exclusive when no legend supplied

### LINKED ISSUE #389: Mutually exclusive does not work when there is no legend supplied
Mutually exclusive does not work when there is no legend supplied for checkboxes, e.g. when wrapping the field in a question component.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/29#issuecomment-495555846_

bameyrick:
Will resolve #389

### DIFF
--- a/src/components/field/_macro.njk
+++ b/src/components/field/_macro.njk
@@ -27,7 +27,7 @@
 
     {% if params.legend or (params.mutuallyExclusive and params.legend) %}
         {% if not params.noField %}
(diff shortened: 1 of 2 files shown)


bameyrick:
[ Mutually exclusive does not work when there is no legend supplied](https://trello.com/c/tLTSwlrk/288-mutually-exclusive-does-not-work-when-there-is-no-legend-supplied)

PR #405, opened 2019-06-06
Fix descriptions and checked state on mutually exclusive checkbox

bameyrick:


bameyrick:
[Descriptions don't show on mutually exclusive checkbox](https://trello.com/c/LX5WmbmG/297-descriptions-dont-show-on-mutually-exclusive-checkbox)

PR #406, opened 2019-06-06
Replace Lato with Open Sans stylistic set 2

bameyrick:


bameyrick:
[Replace Lato with Open Sans](https://trello.com/c/xsZNd9Fs/298-replace-lato-with-open-sans)

PR #415, opened 2019-06-07
UAC Input

bameyrick:
This still requires documentation of the pattern


bameyrick:
[UAC Entry](https://trello.com/c/IX0IvJFO/302-uac-entry)

PR #417, opened 2019-06-11
Use p tags for fields

### LINKED ISSUE #239: Use <p> tags for fields
We should probably to update the field to component to use `<p>` tags (when not rendering as a `<fieldset>`) as recommended by WHATWG: https://html.spec.whatwg.org/multipage/forms.html#writing-a-form's-user-interface

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/35#issuecomment-466981298_

bameyrick:
Will resolve #239


bameyrick:
[Use <p> tags for fields](https://trello.com/c/CBJM07nB/185-use-p-tags-for-fields)

PR #432, opened 2019-06-14
Set textarea charcount element to a span

### LINKED ISSUE #431: The textarea charcount element cannot be a `div` as the wrapping field element is a `p`
The charcount element cannot be a `div` as the wrapping field element is a `p`

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/139#issuecomment-502012549_

bameyrick:
Will resolve #431


bameyrick:
[ The textarea charcount element cannot be a `div` as the wrapping field element is a `p`](https://trello.com/c/kBnpf0VD/310-the-textarea-charcount-element-cannot-be-a-div-as-the-wrapping-field-element-is-a-p)

PR #433, opened 2019-06-17
Added NISRA logo and sorted out ONS logos to allow for variable width logos

bameyrick:


bameyrick:
[Create NISRA example of a header](https://trello.com/c/Wvsunqtm/311-create-nisra-example-of-a-header)

PR #446, opened 2019-06-19
Fix layout of radio fields

### LINKED ISSUE #444: Radios break the field layout
Radios break the field layout because the `field__items` element is a `div` rather than a `span`

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/56#issuecomment-503496360_

bameyrick:
Will resolve #444


bameyrick:
[ Radios break the field layout](https://trello.com/c/1Fr6wI0r/321-radios-break-the-field-layout)

PR #451, opened 2019-06-19
Added cta component and examples

boxadesign:
CTA component.

Button as link component.

### DIFF
--- a/src/components/call-to-action/_macro.njk
+++ b/src/components/call-to-action/_macro.njk
@@ -4,9 +4,10 @@
         <div class="container">
             <div class="grid grid--flex grid--vertical-center grid--no-wrap@s">
                 <div class="grid__col col-auto u-flex-shrink@s">
-                    <p class="call-to-action__desc u-fs-r u-mb-no@s"><strong>{{ params.headingText }}</strong> {{ params.paragraphText }}</p>
+                    <h2 class="u-fs-r--b u-di">{{ params.headingText }}</h2> 
+                    <p class="u-di">{{ params.paragraphText }}</p>
                 </div>
-                <div class="grid__col col-auto u-flex-no-grow">
+                <div class="grid__col col-auto u-flex-no-grow u-mt-xs@xs@s">
                     {{ onsButton({
                         "text": params.button.text,
                         "url": params.button.url,
--- a/src/components/call-to-action/_macro.njk
+++ b/src/components/call-to-action/_macro.njk
@@ -1,0 +1,20 @@
+{% macro onsCallToAction(params) %}
+    {% from "components/button/_macro.njk" import onsButton %}
+    <div class="call-to-action">
+        <div class="container">
+            <div class="grid grid--flex grid--vertical-center grid--no-wrap@s">
+                <div class="grid__col col-auto u-flex-shrink@s">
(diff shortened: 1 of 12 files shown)


src/components/call-to-action/_macro.njk
```diff
@@ -0,0 +1,20 @@
+{% macro onsCallToAction(params) %}
+    {% from "components/button/_macro.njk" import onsButton %}
+    <div class="call-to-action">
+        <div class="container">
+            <div class="grid grid--flex grid--vertical-center grid--no-wrap@s">
+                <div class="grid__col col-auto u-flex-shrink@s">
+                    <p class="call-to-action__desc u-fs-r u-mb-no@s"><strong>{{ params.headingText }}</strong> {{ params.paragraphText }}</p>
```
bameyrick: After our visit to DAC yesterday I have a feeling`<strong>` tags cause screen readers to pause. I'll check this on JAWS, NVDA, and VoiceOver today

src/views/layouts/_master.njk
```diff
@@ -6,8 +6,14 @@
     {% set page = {} %}
 {% endif %}
 
+
 {% if example %}
     {% set bodyClasses = bodyClasses + " patternlib-page--example" %}
+    {% if page.theme %}
```
bameyrick: What's this for?

boxadesign: So we can add a theme to an individual example. Look at the deployment.

src/components/call-to-action/_macro.njk
jrbarnes9: Yes we saw that too @bameyrick. Maybe test if it does the same with `<em>`

src/components/call-to-action/_macro.njk
```diff
@@ -0,0 +1,20 @@
...
+                </div>
+                <div class="grid__col col-auto u-flex-no-grow">
+                    {{ onsButton({
+                        "text": "Start survey",
```
jrbarnes9: Doesn't this need to be a parameter?

boxadesign: Yes

src/components/call-to-action/_macro.njk
boxadesign: JAWS and voiceover reads out tag by tag. The string in the strong tag could be a heading which would be announced in the same way. Perhaps it should be a heading so screen readers would read it out if all headings are read out?

jrbarnes9: Not sure if matters if screen reader pauses in this instance.

bameyrick: If it is a heading and a description, wouldn't it make more sense for this to be a heading tag followed by a paragraph tag then?

boxadesign: Yes that's what I'm saying. I'll update to be a `h2` and `p`

src/views/layouts/_master.njk
bameyrick: Ah cool I see what you've done here, I was an idiot and thought this was the main page-template.njk

bameyrick:
[ Get Started button on homepage doesn't link to anywhere](https://trello.com/c/HegoTuy1/267-get-started-button-on-homepage-doesnt-link-to-anywhere)

PR #463, opened 2019-06-21
Video macro

boxadesign:


src/components/video/_macro.njk
```diff
@@ -0,0 +1,7 @@
+{%- macro onsVideo(params) -%}
+    <div class="video">
+        {% if params.youtubeUrl %}
+            <iframe class="video__iframe" width="560" height="315" src="{{ params.youtubeUrl }}" frameborder="0" allow="accelerometer; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
```
bameyrick: Should the width and height on this be hardcoded?

src/components/video/examples/video/index.njk
```diff
@@ -0,0 +1,6 @@
+{% from "components/video/_macro.njk" import onsVideo %}
+{{ 
+    onsVideo({
+        "youtubeUrl": 'https://youtu.be/BiZk2pnSM6w'
```
bameyrick: Youtube short URLs are blocked by most adBlockers, so it's best to use full URLs
```suggestion
        "youtubeUrl": "https://www.youtube.com/watch?v=BiZk2pnSM6w"
```

src/components/video/_macro.njk
boxadesign: Well it's what Youtube give you but it's overridden with css.

bameyrick: If it's overridden it's probably best to remove it

PR #474, opened 2019-06-26
Moved image component to components from styles

boxadesign:


src/components/call-to-action/_macro.njk
```diff
@@ -4,8 +4,8 @@
         <div class="container">
             <div class="grid grid--flex grid--vertical-center grid--no-wrap@s">
                 <div class="grid__col col-auto u-flex-shrink@s">
-                    <h2 class="u-fs-r--b u-di">{{ params.headingText }}</h2> 
-                    <p class="u-di">{{ params.paragraphText }}</p>
+                    <h2 class="call-to-action__heading u-fs-r--b u-di">{{ params.headingText }}</h2> 
+                    <p class="call-to-action__heading__text u-di">{{ params.paragraphText }}</p>
```
bameyrick: With BEM classnames can only have one element part of the classname, e.g. Block__Element--modifier. 

```suggestion
                    <p class="call-to-action__heading-text u-di">{{ params.paragraphText }}</p>
```

boxadesign: Yeah that was an error on my part. Thanks for highlighting Ben.

boxadesign: It should actually be `call-to-action__text`

PR #477, opened 2019-06-27
Make address component more configurable

### LINKED ISSUE #476: The address input needs to be made more configurable
The address input needs to be made more configurable to allow only some fields being used. The current conditions used in the macro aren't compatible with Jinja so these need to be checked also.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/412#issuecomment-506330166_

bameyrick:
**BREAKING CHANGES**
- Macro renamed from `onsManualAddressInput` to `onsAddressInput`

Will resolve #476


bameyrick:
[ The address input needs to be made more configurable](https://trello.com/c/uzWr74TQ/341-the-address-input-needs-to-be-made-more-configurable)

PR #492, opened 2019-07-03
Add state to details component

rmccar:
This PR is to save in localstorage the state of a details component. This will mean that when a details component is open and the user navigates back to the page in the same session it will still be open.


rmccar:
[Remember accordion states when returning to a page](https://trello.com/c/WNCGHmCv/108-remember-accordion-states-when-returning-to-a-page)

PR #494, opened 2019-07-04
Remove old hack for legend text wrapping in unsupported version of IE as it causes issues with screen readers

### LINKED ISSUE #489: The `legend` has `display: table` set by default, which may prevent screen readers from reading it
The `legend` has `display: table` set by default, which may prevent screen readers from reading it when they're not visually hidden.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/35#issuecomment-507623867_

bameyrick:
Will resolve #489

### DIFF
--- a/src/scss/base/_global.scss
+++ b/src/scss/base/_global.scss
@@ -48,9 +48,3 @@ main,
     outline: none;
   }
 }
-
-legend {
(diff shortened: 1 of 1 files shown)


bameyrick:
[ The `legend` has `display: table` set by default, which may prevent screen readers from reading it](https://trello.com/c/g9nAbC4y/351-the-legend-has-display-table-set-by-default-which-may-prevent-screen-readers-from-reading-it)

PR #498, opened 2019-07-04
Ensure if there is only one summary item it still gets given a border bottom

### LINKED ISSUE #447: We need to add the top border back on to the first item of summaries
We need to add the top border back on to the first item always so that zoom text users can discover the action links.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/84#issuecomment-503528870_

bameyrick:
Will resolve #447


bameyrick:
[ We need to add the top border back on to the first item of summaries](https://trello.com/c/iuLkf5Gn/324-we-need-to-add-the-top-border-back-on-to-the-first-item-of-summaries)

PR #503, opened 2019-07-05
 Split field macro into separate macros for field, fieldset, and mutually exclusive

### LINKED ISSUE #499: Split field macro into 3 separate macros
Using to automatically switch between fieldsets, and p tags based on whether they are calling input has `mutuallyExclusive`(#29) config specified is too complex and causes issues when in use in projects like runner. We should separate all the logic held in the field component into separate mutually exclusive, field, and fieldset macros that are simpler to use.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/35#issuecomment-508524865_

bameyrick:
The logic for fields, fieldsets, and mutually exclusive has been simplified by separating out in to 3 separate macros. This is to make it easier to configure inputs with the `dontWrap` parameter enabled to simplify the logic of when to wrap answers in fieldsets in runner.

Will resolve #499


bameyrick:
[ Split field macro into 3 separate macros](https://trello.com/c/GmbThf75/353-split-field-macro-into-3-separate-macros)

PR #512, opened 2019-07-08
Ensure value can be passed to text input on radios and checkboxes

### LINKED ISSUE #510: The value parameter for the "other" text input on checkboxes and radios
The value parameter for the "other" text input isn't passed to the input macro so the field cannot be repopulated.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/13#issuecomment-509175004_

bameyrick:
Will resolve #510


bameyrick:
[ The value parameter for the "other" text input on checkboxes and radios](https://trello.com/c/nbwQiaSH/358-the-value-parameter-for-the-other-text-input-on-checkboxes-and-radios)

PR #514, opened 2019-07-09
Increase contrast ratio of input borders

### LINKED ISSUE #513: Contrast ratio on form input borders is too low
The contrast ratio of the border on the inputs and radios etc is too low. We need to increase the contrast to improve visibility for users with low vision.

bameyrick:
Will resolve #513


bameyrick:
[ Contrast ratio on form input borders is too low](https://trello.com/c/J7bxWZyZ/360-contrast-ratio-on-form-input-borders-is-too-low)

PR #515, opened 2019-07-09
Add example for PO approval

rmccar:
This PR is to add the example for https://github.com/ONSdigital/design-system/pull/492 so that it can be PO approved

### DIFF
--- a/src/components/details/_macro-options.md
+++ b/src/components/details/_macro-options.md
@@ -5,6 +5,7 @@
 | content           | string          | true     | HTML content for the details                                                |
 | button            | `DetailsButton` | false    | Settings for the close button. If not specified button will not render      |
 | classes           | string          | false    | Classes to add to the details element                                       |
+| saveState         | string          | false    | Allows saving of details state (open or closed) localy                      |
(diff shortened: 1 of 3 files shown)


src/components/details/_macro-options.md
```diff
@@ -5,6 +5,7 @@
 | content           | string          | true     | HTML content for the details                                                |
 | button            | `DetailsButton` | false    | Settings for the close button. If not specified button will not render      |
 | classes           | string          | false    | Classes to add to the details element                                       |
+| saveState         | string          | false    | Allows saving of details state (open or closed) localy                      |
```
bameyrick:
```suggestion
| saveState         | boolean          | false    | Allows saving of details state (open or closed) localy                      |
```

src/components/details/index.njk
```diff
@@ -38,6 +38,10 @@ Don't use the details component to hide information that the majority of your us
 
 Use the details component instead of [tabs](/components/tabs) or [accordions](/components/accordion) if there is only one section of content that needs to be condensed.
```
bameyrick:
```suggestion
## Details with save state enabled
```

PR #527, opened 2019-07-10
Fix field with no params set in jinja

### LINKED ISSUE #526: Input does not render in eq-survey runner with mutually exclusive set
Input does not render in eq-survey runner with mutually exclusive set as the wrapping onsField element has no params object set.

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/35#issuecomment-510102555_

bameyrick:
Will resolve #526

### DIFF
--- a/src/components/field/_macro.njk
+++ b/src/components/field/_macro.njk
@@ -1,4 +1,7 @@
 {% macro onsField(params) %}
+    {% if not params %}
+        {% set params = {} %}
+    {% endif %}
     <p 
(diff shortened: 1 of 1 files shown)


bameyrick:
[ Input does not render in eq-survey runner with mutually exclusive set](https://trello.com/c/MTTCF6XU/367-input-does-not-render-in-eq-survey-runner-with-mutually-exclusive-set)

PR #537, opened 2019-07-15
Create duration component to make integrating it easier

bameyrick:
Created a duration component to make integrating it with runner easier.

Updated examples and unit tests to use the new component.


bameyrick:
[Create duration component to make integrating it easier](https://trello.com/c/TYQhMRBI/369-create-duration-component-to-make-integrating-it-easier)

PR #544, opened 2019-07-16
Fix onsMutuallyExclusive component for jinja2 renderer

### LINKED ISSUE #543: caller() is not working correctly when using jinja render engine
**Describe the bug**
Using caller in macros that are called from other macros does not work in jinja (see: https://github.com/pallets/jinja/issues/371)

All macros using caller need to be refactor.

bameyrick:
Will resolve #543

### DIFF
--- a/src/components/field/_macro.njk
+++ b/src/components/field/_macro.njk
@@ -6,12 +6,15 @@
     {% endif %}
 
     {% set field %}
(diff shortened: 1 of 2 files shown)


bameyrick:
[ caller() is not working correctly when using jinja render engine](https://trello.com/c/xCpdjbD9/372-caller-is-not-working-correctly-when-using-jinja-render-engine)

PR #551, opened 2019-07-16
Correct markup of fields and durations when an error is provided

### LINKED ISSUE #549: Durations error markup incorrect
When durations are configured to only render one field, and that field has an error, the field wraps the error panel rather than the field being within the error panel.

Screenshot 2019-07-16 at 15 05 32

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/19#issuecomment-511832624_

bameyrick:
Will resolve #549

### DIFF
--- a/src/components/duration/_macro.njk
+++ b/src/components/duration/_macro.njk
@@ -17,7 +17,7 @@
 
     {% set fields %}
         {% if params.label %}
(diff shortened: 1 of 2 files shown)


bameyrick:
[ Durations error markup incorrect](https://trello.com/c/74ISgJLP/374-durations-error-markup-incorrect)

PR #552, opened 2019-07-16
Ensure date fields are only wrapped in field-group if there are multiple fields

### LINKED ISSUE #550: When the date component is configured to only show one field, the fields are incorrectly wrapped in a field-group
When the date component is configured to only show one field, the fields are incorrectly wrapped in a `field-group`

Screenshot 2019-07-16 at 15 14 32

_Originally posted by @bameyrick in https://github.com/ONSdigital/design-system/issues/22#issuecomment-511834766_

bameyrick:
Will resolve #550

### DIFF
--- a/src/components/date-input/_macro.njk
+++ b/src/components/date-input/_macro.njk
@@ -26,48 +26,56 @@
             <p class="u-fs-r--b u-mb-xs">{{ params.label }}</p>
         {% endif %}
 
-        <div class="field-group">
-            {% if params.day %}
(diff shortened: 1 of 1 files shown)


bameyrick:
[ When the date component is configured to only show one field, the fields are incorrectly wrapped in a field-group](https://trello.com/c/c4ZZJcwi/376-when-the-date-component-is-configured-to-only-show-one-field-the-fields-are-incorrectly-wrapped-in-a-field-group)

PR #558, opened 2019-07-18
Update the browser support list to match GDS

bameyrick:
- Remove test for browsers we no longer support
- Update browserlist for autoprefixer and webpack


bameyrick:
[Update browser support list to reflect GDS](https://trello.com/c/tZmdMl0w/378-update-browser-support-list-to-reflect-gds)

bameyrick:
We need to upgrade our browserstack plan to automate mobile before this can be merged

PR #563, opened 2019-07-24
Quote and text indent styles

boxadesign:
Look in Styles -> Typography

Styling for `blockquote` and text-indent for addresses

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -12,6 +12,7 @@ $button-border-height: 3px;
   font-family: inherit;
   font-size: 1rem;
   font-weight: 600;
+  line-height: 1.15;
 
   text-rendering: optimizeLegibility;
 
--- a/src/components/quote/_macro.njk
+++ b/src/components/quote/_macro.njk
@@ -1,6 +1,6 @@
 {% macro onsQuote(params) %}
     <blockquote class="quote">
-        <span class="quote__text">{{ params.text | safe }}</span>
+        <span class="quote__text">{{ caller() if caller else params.text | safe }}</span>
         {% if params.ref %}<span class="quote__ref">- {{ params.ref | safe }}</span>{% endif %}
     </blockquote>
 {% endmacro %}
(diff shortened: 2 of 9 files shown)


src/components/button/_button.scss
```diff
@@ -12,6 +12,7 @@ $button-border-height: 3px;
   font-family: inherit;
   font-size: 1rem;
   font-weight: 600;
+  line-height: 1.15;
```
bameyrick: Took me a while to figure out what this was fixing, so might be worth commenting that this fixes the CTA button ore be more specific with the fix so something like:

```
.btn [class*='icon--'] { 
    line-height: 1.15;
}
```

to specifically show this is a fix for buttons with icons.

src/components/text-indent/_macro.njk
```diff
@@ -0,0 +1,5 @@
+{% macro onsTextIndent(params) %}
+    <div class="text-indent">
+        {{ params.text | safe }}
```
bameyrick: Could be good to set this up so it can be utilised using [call](https://mozilla.github.io/nunjucks/templating.html#call)

```suggestion
        {{ caller() if caller else params.text | safe }} 
```

src/components/quote/_macro.njk
```diff
@@ -0,0 +1,6 @@
+{% macro onsQuote(params) %}
+    <blockquote class="quote">
+        <span class="quote__text">{{ params.text | safe }}</span>
```
bameyrick: Could be good to set this up so it can be utilised using [call](https://mozilla.github.io/nunjucks/templating.html#call)

```suggestion
        <span class="quote__text">{{ caller() if caller else params.text | safe }}</span>
```

src/components/button/_button.scss
boxadesign: @bameyrick It doesn't fix the` icon--` class it fixes the `.btn` class as the `line-height` declaration is required for both `button` and `a` implementations. The same `line-height` value is also declared on the `button` element which isn't specific to the `.btn`.

PR #567, opened 2019-07-26
Typeaheads

bameyrick:
- Created typeahead component
- Noticed JS bundle was getting large so updated all components to be conditionally lazy loaded


bameyrick:
[Create typeaheads component](https://trello.com/c/H5FdtNu8/385-create-typeaheads-component)

PR #569, opened 2019-07-30
Update address component to have javascript lookup option

bameyrick:


bameyrick:
[Update address component to have JavaScript lookup option](https://trello.com/c/QMVNCfvi/386-update-address-component-to-have-javascript-lookup-option)

PR #578, opened 2019-08-09
Promo banner component and example

jrbarnes9:
Added a new 'promotional banner' component with a census example.  Component allows for a text and image block with a full bleed themed background.


jrbarnes9:
[Promotional banner](https://trello.com/c/GLLGP03o/393-promotional-banner)

PR #590, opened 2019-09-03
342 cookie consent banner

jrbarnes9:
To review the banner:

- Navigate to /patterns/cookie-settings
- Open the banner example in a new window
- Open the 'application' tab in the inspector and click on cookies under storage and the url for the page. 
- Check that "ons_cookie_policy" gets set with default consent i.e. all are set to true
- Click the "accept cookies" button
- Check it hides
- Refresh the page and check it's still hidden
- Look in the cookies in the inspector and check that "ons_cookie_message_displayed" is set

To review the settings page:

- Navigate to /patterns/cookie-settings
- Open the settings page example in a new window
- Open the 'application' tab in the inspector and click on cookies under storage and the url for the page. 
- Check that "ons_cookie_policy" gets set with default consent i.e. all are set to true
- Toggle some radios on the page and save the page
- Check that the confirmation message is displayed
- Refresh the page and check that the correct radio values are set
- Look in the cookies in the inspector and check that "ons_cookie_policy" gets set with the correct consent values

Any cookie categories that are set to false will delete those cookies that are within that category.


jrbarnes9:
[Cookie consent banner](https://trello.com/c/ZC7Rwqml/342-cookie-consent-banner)

PR #597, opened 2019-09-13
407 transactional footer

jrbarnes9:
Transactional footer for census requires more links than standard, with primary & secondary rows.


jrbarnes9:
[Add second row to transactional footer](https://trello.com/c/bTV4nIZR/407-add-second-row-to-transactional-footer)

PR #606, opened 2019-09-18
button active and hover states

jrbarnes9:
Active and hover states were missing from buttons following the new a11y focus state changes.

## How to review
Check all the hover, active and focus states of all button types.

### DIFF
--- a/src/components/hero/_macro.njk
+++ b/src/components/hero/_macro.njk
@@ -27,7 +27,7 @@
                                 onsButton({
                                     "type": 'button',
                                     "text": params.button.text,
-                                    "classes": 'btn--link u-mb-m',
+                                    "classes": 'u-mb-m',
                                     "url": params.button.url,
                                     "innerClasses": "icon--chevron-right-white"
                                 }) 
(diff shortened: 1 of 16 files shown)


jrbarnes9:
[Buttons: Missing hover and active states](https://trello.com/c/GSenOMQD/410-buttons-missing-hover-and-active-states)

src/components/hero/_macro.njk
```diff
@@ -27,7 +27,7 @@
                                 onsButton({
                                     "type": 'button',
                                     "text": params.button.text,
-                                    "classes": 'u-mb-m',
+                                    "classes": 'btn--link u-mb-m',
                                     "url": params.button.url,
                                     "innerClasses": "icon--chevron-right-white"
```
boxadesign: I think this innerClasses property can be removed here as the btn--link class does the same thing

jrbarnes9:
[Buttons opening in a new window need to warn users](https://trello.com/c/KGPnk9He/413-buttons-opening-in-a-new-window-need-to-warn-users)

PR #611, opened 2019-09-23
new logo focus states for external and internal

jrbarnes9:
The outline on logos on white bg in the header fail WCAG 3:1 contrast ratio requirement for non-text elements (i.e. focus state)


jrbarnes9:
[Logos focus states fail contrast a11y](https://trello.com/c/p47cGcWa/414-logos-focus-states-fail-contrast-a11y)

Issue #629, opened 2019-10-09
Disabled/unready button state


boxadesign:
We need to explore a solution for disabled buttons. Currently we don't show a button if there is no available action. R'Ops currently have a disabled/unready button so we need to explore why/if they need it.

jrbarnes9:
What's the rationale for this in R-Ops then? I remember we all had a discussion about disabled buttons last year, and we deemed them unnecessary and that we should only display a button when the user needs it.

PR #633, opened 2019-10-11
Created a button style and example for disabled button state

boxadesign:
Will resolve https://github.com/ONSdigital/design-system/issues/629

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -20,8 +20,10 @@ $button-border-height: 3px;
   vertical-align: top;
   cursor: pointer;
 
-  &:hover {
-    text-decoration: none;
+  &:hover & {
+    &__inner {
+      background: darken($color-button, 5%);
+    }
   }
 
   &__inner {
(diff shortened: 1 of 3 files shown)


src/components/button/_button.scss
```diff
@@ -44,8 +44,11 @@ $button-border-height: 3px;
   }
 
   // When hovered
-  &:not([class*='btn--ghost']):not([class*='btn--mobile']):not([class*='btn--secondary']):not([class*='btn--print']):hover &__inner {
-    background: darken($color-button, 5%);
+  &:not([class*='btn--ghost']):not([class*='btn--mobile']):not([class*='btn--secondary']):not([class*='btn--print']):not([class*='btn--disabled']):hover
```
rmccar: Is there a better way of doing this rather than listing them all out? Just thinking this is a list we will have to make sure we maintain if we add new button types

boxadesign:
> Perhaps removing the bottom border and changing the text colour would help with the affordance?...
> Screenshot 2019-10-14 at 10 31 55
> 
> I don't think the background colour change alone is enough, especially on mobile where they don't get the disabled cursor on hover.
> 
> Also, using a grey to indicate a disabled button could be an issue considering we use grey for active secondary buttons. E.g. I don't think the cancel button looks particularly disabled in this instance...
> Screenshot 2019-10-14 at 10 42 10

I've updated so the secondary button is not given a border or hover state. I think this overall needs more research but needs to go in as it's the current solution in R'Ops

PR #636, opened 2019-10-15
Disable submit button after click to stop multiple submissions

rmccar:
This is to fix multiple submissions happening by styling the button using the loading style and disabling the button to stop it from being clicked more than once

### DIFF
--- a/src/components/button/_macro.njk
+++ b/src/components/button/_macro.njk
@@ -9,7 +9,7 @@
       {% else %}
         type="{{ params.type if params.type else ('button' if params.print else 'submit') }}"
       {% endif %}
-      class="btn{% if params.classes %} {{ params.classes }}{% endif %}{% if params.url %} btn--link{% endif %}{% if params.disabled %} btn--disabled{% endif %}{% if params.print %} btn--print u-d-no js-print-btn{% endif %}{% if params.type == 'submit' or params.type is undefined %} btn--loader js-button{% endif %}"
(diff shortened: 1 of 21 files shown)


rmccar:
[Temporarily disable submit button to guard against double clicks](https://trello.com/c/CvzQKCMM/3446-temporarily-disable-submit-button-to-guard-against-double-clicks)

boxadesign:
@rmccar There is a log line that should be removed. This will also need unit tests. Other than that it looks good!

src/components/button/_macro.njk
```diff
@@ -9,11 +9,11 @@
       {% else %}
         type="{{ params.type if params.type else ('button' if params.print else 'submit') }}"
       {% endif %}
-      class="btn{% if params.classes %} {{ params.classes }}{% endif %}{% if params.url %} btn--link{% endif %}{% if params.disabled %} btn--disabled{% endif %}{% if params.print %} btn--print u-d-no js-print-btn{% endif %}"
+      class="btn{% if params.classes %} {{ params.classes }}{% endif %}{% if params.url %} btn--link{% endif %}{% if params.disabled %} btn--disabled{% endif %}{% if params.print %} btn--print u-d-no js-print-btn{% endif %}{% if params.type == 'submit' or params.type is undefined %} btn--loader js-button{% endif %}"
```
boxadesign: I don't think we should add this functionality if the type is `undefined`. I think it should be specifically if the type is submit.

Issue #642, opened 2019-10-17
Allow table caption to be visually hidden


boxadesign:
Currently the `caption` will be shown if provided to the `table` component. It may not be required based on the surrounding content of the page but will still be useful to screen reader users who may target the table before getting context from elsewhere on the page.

PR #643, opened 2019-10-17
Updated table to allow visually hidden caption

boxadesign:
Will resolve https://github.com/ONSdigital/design-system/issues/642


src/components/table/_macro.njk
```diff
@@ -16,7 +16,7 @@
                 </thead>
                 <tbody class="table__body">
                     {% for tr in params.trs %}
-                    <tr class="table__row">
+                    <tr class="table__row{{ " table__row--highlight" if tr.highlight }}">
```
rmccar: I think this highlight param needs to be added to the mark down. Other than that looks good

PR #644, opened 2019-10-18
Add second version of loader button

rmccar:
This is to add a second implementation of the loader button. The second version will be a timed version that will only be disabled for 1 second and then will be become enabled again. This is to reduce the double clicking of submit buttons.


src/components/button/button.js
```diff
@@ -1,13 +1,28 @@
...
 
-  submitButton() {
+  loaderButton() {
     this.button.classList.add('is-loading');
     this.button.setAttribute('disabled', true);
   }
+
+  timerButton() {
+    this.button.setAttribute('disabled', true);
+    setTimeout(
```
boxadesign: This could be updated to use es6 syntax e.g. `setTimeout(() => {`

rmccar:
[Temporarily disable submit button to guard against double clicks](https://trello.com/c/CvzQKCMM/3446-temporarily-disable-submit-button-to-guard-against-double-clicks)

src/components/button/button.dom.js
```diff
@@ -1,13 +1,20 @@
...
   if (buttons.length && form.length) {
     const SubmitButton = (await import('./button')).default;
     buttons.forEach(button => {
-      new SubmitButton(button, form[0]);
+      if (button.classList.contains('js-timer')) {
+        submitType = 'timer';
+      } else if (button.classList.contains('js-loader')) {
+        submitType = 'loader';
+      }
+      console.log(submitType);
```
boxadesign: Remove log line.

rmccar: Missed that one! Its fixed

PR #657, opened 2019-11-01
Add search component

boxadesign:


jrbarnes9:
We should keep the focus state the same colour (3px yellow box-shadow) on the error state so users can follow it down the page as they tab.   Only change the input border and outline from black to red.
Screenshot 2019-11-07 at 09 37 24

src/components/input/_macro.njk
```diff
@@ -24,13 +24,15 @@
             class="input input--text input-type__input {{ params.classes }}{{ exclusiveClass }}"
             {% if params.prefix or params.suffix %}title="{{ params.prefix.title if params.prefix }}{{ params.suffix.title if params.suffix }}"{% endif %}
             {% if params.name is defined %}name="{{ params.name }}"{% endif %}
+            {% if params.accessiblePlaceholder is defined %}required{% endif %}
             {% if params.value is defined %}value="{{ params.value }}"{% endif %}
             {% if params.accept is defined %}accept="{{ params.accept }}"{% endif %}
-            {% if params.min is defined %}min="{{ params.min }}"{% endif %}
-            {% if params.max is defined %}max="{{ params.max }}"{% endif %}
```
jrbarnes9: Don't we need to keep these for e.g. date input?

src/components/input/_macro.njk
```diff
@@ -24,13 +24,15 @@
...
+            {% if params.min is defined %}minlength="{{ params.min }}"{% endif %}
+            {% if params.max is defined %}maxlength="{{ params.max }}"{% endif %}
```
jrbarnes9: Think these should be in addition to min and max.

src/components/input/_macro.njk
boxadesign: Yes. I'm not sure why I even changed it.

src/components/input/_macro.njk
boxadesign: `min` covers both dates and numbers whereby `minLength` is only for numbers.

jrbarnes9:
[Search component](https://trello.com/c/HhOv2obw/450-search-component)

PR #669, opened 2019-11-29
CTA button focus styling issues

### LINKED ISSUE #658: CTA button text shouldn't be underlined on hover...
CTA button text shouldn't be underlined on hover...
Screenshot 2019-11-06 at 15 00 54

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/589#issuecomment-550349184_

jrbarnes9:
Resolves #658


jrbarnes9:
[CTA button focus styling issues](https://trello.com/c/Z8zi9Tcf/451-cta-button-focus-styling-issues)

src/scss/base/_global.scss
```diff
@@ -31,7 +31,7 @@ hr {
 
 a {
   color: $color-links;
-  &:hover {
+  &:hover:not(.btn--link):not(.btn--ghost) {
```
boxadesign: Not sure about this. We don't want to couple classes to states in this way I don't think. Would be better overiding the attributes in each class. Could make maintaining difficult if classes are spread around different files.

jrbarnes9: ok

PR #671, opened 2019-12-03
Disable clicks on link btns

boxadesign:


src/components/button/examples/button-link/index.njk
```diff
@@ -2,6 +2,7 @@
 {{
     onsButton({
         "type": 'button',
+        "submitType": "timer",
```
jrbarnes9: Do we need to add this for this example? Or did you just add this for testing?

boxadesign: Just for something to test. Can remove when you have tested

Issue #672, opened 2019-12-06
Add `maxlength` to date inputs


jrbarnes9:
Users have fed back frustrations around the ability to add more than two or four characters in the date inputs.  This is especially frustrating for users who expect auto-tabbing, because they end up entering everything into the first input (which is too small to easily edit/delete)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/22#issuecomment-562507566_

rmccar:
Done

PR #674, opened 2019-12-06
Add maxlength to date input

jrbarnes9:
Users have fed back frustrations around the ability to add more than two or four characters in the date inputs.  This is especially frustrating for users who expect auto-tabbing, because they end up entering everything into the first input (which is too small to easily edit/delete)


jrbarnes9:
[Issue #672: Add `maxlength` to date inputs](https://trello.com/c/6gMTMxvG/464-issue-672-add-maxlength-to-date-inputs)

PR #675, opened 2019-12-10
Add message component

boxadesign:
Created a messages pattern to display message threads in r'ops. Can be viewed:

/patterns/message


src/patterns/message/index.njk
```diff
@@ -0,0 +1,39 @@
...
+## Message - conversation
+Each message can be stacked to display the conversation/thread.
+{{
+    patternlibExample({"path": "patterns/message/examples/conversation/index.njk"})
+}}
+
+## Research on this component
+{% from "components/panel/_macro.njk" import onsPanel %}
+{% call onsPanel() %}
+   If you have conducted any user research using this component, please feed back your findings via the <a href="https://github.com/ONSdigital/design-system/issues/48" target="_blank">Design System forum</a>.
```
jrbarnes9: update link

jrbarnes9:
This the 'close' should probably be a link to reduce the target size.  Maybe stacked underneath.  Think the larger button could be prone to accidental clicks. 
Maybe change send label to 'send message' to add screenreader context? (and consequently increase the primary target size)

boxadesign:
> I think if I was building this I'd want to know what text should precede the text area in the case of writing the first message in the thread (obviously not 'reply') so maybe add a 'first message' example or state it in the documentation somewhere.

You will only reply from this page. There is a create message page which initiates the thread or you receive a message and reply here. I'll put the create message pattern in here as well.

jrbarnes9:
> > I think if I was building this I'd want to know what text should precede the text area in the case of writing the first message in the thread (obviously not 'reply') so maybe add a 'first message' example or state it in the documentation somewhere.
> 
> You will only reply from this page. There is a create message page which initiates the thread or you receive a message and reply here.

Are you adding an example of this pattern then?

boxadesign:
> > > I think if I was building this I'd want to know what text should precede the text area in the case of writing the first message in the thread (obviously not 'reply') so maybe add a 'first message' example or state it in the documentation somewhere.
> > 
> > 
> > You will only reply from this page. There is a create message page which initiates the thread or you receive a message and reply here.
> 
> Are you adding an example of this pattern then?

Wasn't planning on it as it's very open ended. We are also currently changing the create message screen in r'ops which will contain categories/templates etc.

PR #681, opened 2019-12-19
Clear radios

jrbarnes9:
'clear' is a reserved term in Python, so updated 'clear' param to a distinctly unique parameter.


jrbarnes9:
[Always visible 'other' inputs and clear radio selection](https://trello.com/c/eqateaun/459-always-visible-other-inputs-and-clear-radio-selection)

PR #684, opened 2020-01-06
Changed clear radios to clear with js and not reset button type

boxadesign:


src/components/radios/examples/radios-clear/index.njk
```diff
@@ -7,6 +7,7 @@
             "name": "food",
             "clearRadios": {
                 "text": "Clear selection",
+                "name": "clear-radios",
```
jrbarnes9: Do we need to add to the macro options table?

jrbarnes9:
I think we should document the need to add a hidden duplicate save and continue button to the top of the form to ensure the default enter key behaviour actions this button, and not the clear button.

src/components/radios/index.njk
```diff
@@ -54,9 +54,11 @@ This variant of the 'Other' radio input type shows the input instead of waiting
 
 ### Radios with the option to clear
 
-This variant displays a 'Clear selection' button which allows the user to deselect/clear the fieldset so all radios are not `:checked`. The button is displayed after a radio is selected. The `aria-live` attribute is used for use in assistive technologies.
+This variant displays a 'Clear selection' button which allows the user to deselect/clear the fieldset so all radios are not `:checked`. The button is displayed after a radio is selected. The `aria-live` attribute is used for announcing changes in assistive technologies.
 {%call onsPanel() %}
-    <p><strong>Note:</strong> This should not be used if a form contains more than a single radios fieldset. The reset button will clear every input within the form.</p>
+    <strong>Notes:</strong> 
+    <p>This should not be used if a form contains more than a single radios fieldset. The reset button will clear every input within the form.</p>
```
jrbarnes9: Don't need this now we're not using reset button

jrbarnes9:
[Always visible 'other' inputs and clear radio selection](https://trello.com/c/eqateaun/459-always-visible-other-inputs-and-clear-radio-selection)

PR #695, opened 2020-02-11
Added id param to panel error text and updated js for inpagelink

boxadesign:
Issue with errors where the first checkbox or radio was being focussed when an error link is clicked. 
There was no other element to attach an id to.

- Added a param to the error macro to allow id to be added to the error text element. 
- If a matching ID is added to an error panel link it will jump to the error text.


rmccar:
So we can see it working I think the example prototype might need updating

PR #706, opened 2020-02-24
Update CDN link in getting started

rmccar:


rmccar:
[Getting started page displaying wrong CDN path to js](https://trello.com/c/5QWvJqkG/471-getting-started-page-displaying-wrong-cdn-path-to-js)

PR #725, opened 2020-03-02
Fix vulnerability in links opening in new tabs

rmccar:
Adding `rel="noopener"` to all `target="_blank"` links


rmccar:
[Any `target="_blank"` link should also have `rel="noopener"`](https://trello.com/c/VugT5QMX/479-any-targetblank-link-should-also-have-relnoopener)

Issue #759, opened 2020-05-29
Breadcrumbs not rendering correctly at breakpoint 500px and below.


trevorsaint:
Breadcrumbs not rendering correctly at breakpoint 500px and below.

![image](https://user-images.githubusercontent.com/4989027/83241813-4948e580-a193-11ea-9168-9e817932045b.png)

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/issues/43#issuecomment-635861022_

jrbarnes9:
I'm only seeing this issue at exactly 500px, not below.

PR #760, opened 2020-05-29
510 breadcrumbs at 500bp

trevorsaint:
Fixed bug #759 with breadcrumb rendering two arrows at 500px.

This was due to a conflict in breakpoints. Increasing the breakpoint by 1px fixed the issue.

### DIFF
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -1,8 +1,8 @@
 $breadcrumb-chevron-height: 0.65rem;
 
 .breadcrumb {
-  display: flex;
   align-items: center;
+  display: flex;
   padding: 1rem 0;
 
   &__items {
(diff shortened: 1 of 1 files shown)


trevorsaint:
[Issue #759: Breadcrumbs not rendering correctly at breakpoint 500px and below.](https://trello.com/c/kBzqENmY/510-issue-759-breadcrumbs-not-rendering-correctly-at-breakpoint-500px-and-below)

rmccar:
Would a better solution be to calculate S breakpoint + 1px or change the definition of the S breakpoint rather than fixing it to 501px?

trevorsaint:
> Would a better solution be to calculate S breakpoint + 1px or change the definition of the S breakpoint rather than fixing it to 501px?

@rmccar I think we can fix this without the need for changing breakpoints. I will revisit this and see what can be done later today.

jrbarnes9:
> > Would a better solution be to calculate S breakpoint + 1px or change the definition of the S breakpoint rather than fixing it to 501px?
> 
> @rmccar I think we can fix this without the need for changing breakpoints. I will revisit this and see what can be done later today.

I'd imagine overlapping breakpoints will cause issues on other components, so probably best to change the bp.

trevorsaint:
> > > Would a better solution be to calculate S breakpoint + 1px or change the definition of the S breakpoint rather than fixing it to 501px?
> > 
> > 
> > @rmccar I think we can fix this without the need for changing breakpoints. I will revisit this and see what can be done later today.
> 
> I'd imagine overlapping breakpoints will cause issues on other components, so probably best to change the bp.

Agreed lets remove any min - max usage, and stick to a true mobile first approach with `min-width` only.

PR #762, opened 2020-06-02
Fix ONS logo alignment

### LINKED ISSUE #756: ONS logo alignment
When the ONS logo is wrapped in a link tag. The link class 'header__logo-link' is using 'inline-block', which causes extra padding. Changing this from 'inline-block' to 'block' should fix the issue.

**Subtle alignment problem**

image

**Fixed using block instead of inline-block**

image

rmccar:
Change header CSS to inherit block instead of using inline-block to fix the alignment of the ONS logo in headers.

Fixes #756


rmccar:
[Issue #756: ONS logo alignment](https://trello.com/c/R488C83k/507-issue-756-ons-logo-alignment)

jrbarnes9:
Need to tidy up the examples.  We don't need so many.

Reduce down to:
- Default (explain this is external and used for any public facing services)
- Internal (add menu and service links to this example)
- External with lang and save button
- External with description
- External with navigation
- External Welsh
- External branded
-- External Census
-- External Census (welsh) (add menu to this example also)
-- External Census NISRA

boxadesign:
Still requires an additional class/height for NISRA logo

rmccar:
Yep I'm going to work on that and the logo stacking card bit today

rmccar:
[Update header for mobile ONS logo (stacked logo)](https://trello.com/c/x1kgrR4z/505-update-header-for-mobile-ons-logo-stacked-logo)

PR #763, opened 2020-06-02
Move hint text outside of label

### LINKED ISSUE #753: Hint text shouldn't be contained within the label (DAC)
Hint text shouldn't be contained within the label. Place outside and use `aria-describedby` (see: https://design-system.service.gov.uk/components/text-input/#hint-text)

rmccar:
Move hint text outside of label and use `aria-describedby`

Fixes #753


rmccar:
[Issue #753: Hint text shouldn't be contained within the label (DAC)](https://trello.com/c/I2VQAl8m/504-issue-753-hint-text-shouldnt-be-contained-within-the-label-dac)

boxadesign:
Margin requires reducing on the label and adding to the description:

image

Previous:

image

boxadesign:
We could consider moving the description out of the label component and into the input/textarea/select etc as it's not really a label description anymore.

jrbarnes9:
> We could consider moving the description out of the label component and into the input/textarea/select etc as it's not really a label description anymore.

Don't we use the same label + description for radios/checkboxes?

PR #767, opened 2020-06-04
Add rows to textarea parameter table and resize textareas when limit reached

### LINKED ISSUE #735: `rows` param should be changed to a responsive unit of measurement
`rows` param should be changed to a responsive unit of measurement based on the amount of characters that can fit in a text area, so the input can reflect the character limit.  This param needs to be added to the macro options table.

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/139#issuecomment-597169854_

rmccar:
Fixes #735


rmccar:
[Issue #735: `rows` param missing from text area macro options table](https://trello.com/c/xphWXDqK/485-issue-735-rows-param-missing-from-text-area-macro-options-table)

jrbarnes9:
The issue was to add a responsive unit to textareas rather than using `rows`, because the number of rows may be suitable to display the limited characters on desktop, but won't be enough when the width of the textarea is reduced at smaller breakpoints.

PR #769, opened 2020-06-04
Refactor to use js-beautify directly

### LINKED ISSUE #745: HTML in examples badly formatted
For example [Radios](https://ons-design-system.netlify.com/components/radios/). See the 'other' example

rmccar:
This PR is to remove `pretty` which we were using as an intermediary for `js-beautify`. This allows us to set our own configuration for `js-beautify` as the one used by `pretty` was causing issues for indenting around `span` tags when displaying HTML for our examples.

After this PR **NODE VERSION 10** is needed

Fixes #745

### DIFF
--- a/lib/helpers/format-html.js
+++ b/lib/helpers/format-html.js
@@ -1,0 +1,22 @@
+import beautify from 'js-beautify';
+
+export function formatHtml(html, condenseWhiteSpace) {
+  const beautifyDefaults = {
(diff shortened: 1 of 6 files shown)


rmccar:
[Issue #745: HTML in examples badly formatted](https://trello.com/c/UQgcBWos/495-issue-745-html-in-examples-badly-formatted)

PR #775, opened 2020-06-14
Improve input macro

boxadesign:
**Changes:**

- Created sub macros for typeahead, character limit check and search patterns and removed markup and logic from inside the input component to reduce complexity and provide a level of abstraction for each variant for improved maintainability.

- Allowed multiple variants to work together i.e. typeahead and character limit check.

- Breaking change - `onsSearch` component is now implemented as `onsInput` and uses `searchButton` param


src/components/search/index.njk
```diff
@@ -4,7 +4,7 @@ title: Search
 {% from "views/partials/example/_macro.njk" import patternlibExample %}
 {% from "components/panel/_macro.njk" import onsPanel %}
 
-The search component provides an input and button for searching a service.
+The search component provides an input and button for searching a service.
```
rmccar, line 7:
```suggestion
The search component provides an input and button for searching a service.
```

PR #777, opened 2020-06-16
Add data-qa attributes

rmccar:
Adding data-qa attributes to remove chained selectors in runner


rmccar:

PR #778, opened 2020-06-16
Fix for checking if label param exists

rmccar:
This is a fix to check to see if label exists on the params before checking label.description.


rmccar:

PR #780, opened 2020-06-17
Add spacing into macro to ensure attributes are spaced in runner

rmccar:
This PR is to fix an error that was spotted on runner where when a checkbox was set as checked and aria attributes weren't set, the attributes wouldn't be spaced correctly.

### DIFF
--- a/src/components/checkboxes/_checkbox-macro.njk
+++ b/src/components/checkboxes/_checkbox-macro.njk
@@ -10,7 +10,7 @@
             {% if params.name %} name="{{ params.name }}"{% endif %}
             {% if params.checked %} checked{% endif %}
             {% if params.other and not params.other.open %} aria-controls="{{ params.id }}-other-wrap" aria-haspopup="true"{% endif %}
-            {% if params.attributes %} {% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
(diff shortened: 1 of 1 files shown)


rmccar:
[Performance investigation for removing minification of HTML output](https://trello.com/c/6SkOMFNy/3921-performance-investigation-for-removing-minification-of-html-output)

src/components/checkboxes/_checkbox-macro.njk
```diff
@@ -10,7 +10,7 @@
             {% if params.name %} name="{{ params.name }}"{% endif %}
             {% if params.checked %} checked{% endif %}
             {% if params.other and not params.other.open %} aria-controls="{{ params.id }}-other-wrap" aria-haspopup="true"{% endif %}
-            {% if params.attributes %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
+            {% if params.attributes %} {% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
```
jrbarnes9, line 13: I think the space probably needs to be here?...
```suggestion
            {% if params.attributes %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %} {{ attribute }}{% if value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
```

rmccar: I thought about putting it there, but it's only a problem for the first attribute thats why I've put it where I have. But I suppose it is safer putting it where you've suggested

PR #785, opened 2020-06-19
Stop typeahead list clearing on perfect match

rmccar:
PR to stop Typeahead clearing listbox when perfect match is found


rmccar:
[[C] Change SAYT list to be visible when direct match only (S)](https://trello.com/c/3xRkIGh1/3931-c-change-sayt-list-to-be-visible-when-direct-match-only-s)

PR #789, opened 2020-06-19
Bug fix - set id on checkbox labels

rmccar:
There is a bug introduced in #763. Ids are no longer being set on checkbox labels. This PR is to fix that.


rmccar:
[Performance investigation for removing minification of HTML output](https://trello.com/c/6SkOMFNy/3921-performance-investigation-for-removing-minification-of-html-output)

rmccar:

PR #791, opened 2020-06-22
Add ci task files

ajmaddaford:
Adds CI task files to be used for building a release and publishing to a GCP bucket. 

The existing Concourse has pipeline in this repo has not been updated to use the build task as we don't currently have access to the Concourse it's deployed to to test any changes.

### DIFF
--- a/ci/tasks/deploy_gcp.yaml
+++ b/ci/tasks/deploy_gcp.yaml
@@ -25,4 +25,4 @@ run:
       gcloud auth activate-service-account --key-file $GOOGLE_APPLICATION_CREDENTIALS
       gcloud config set project "${PROJECT_ID}"
 
-      gsutil -m cp -r ons-design-system-release-artifact gs://census-eq-cdn/design-system/
+      gsutil -m cp -r ons-design-system-release-artifact/** gs://census-eq-cdn/design-system/
--- a/ci/tasks/deploy_gcp.yaml
+++ b/ci/tasks/deploy_gcp.yaml
@@ -1,0 +1,28 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: google/cloud-sdk
+    tag: alpine
+
+params:
+  SERVICE_ACCOUNT_JSON: ((gcp.service_account_json))
+  PROJECT_ID:
+
+inputs:
+- name: ons-design-system-release-artifact
+
+run:
+  path: bash
+  args:
+    - -exc
+    - |
+      export GOOGLE_APPLICATION_CREDENTIALS=~/gcloud-service-key.json
+      cat >$GOOGLE_APPLICATION_CREDENTIALS <<EOL
+      $SERVICE_ACCOUNT_JSON
+      EOL
+      
+      gcloud auth activate-service-account --key-file $GOOGLE_APPLICATION_CREDENTIALS
+      gcloud config set project "${PROJECT_ID}"
+
+      gsutil -m cp -r ons-design-system-release-artifact gs://census-eq-cdn/design-system/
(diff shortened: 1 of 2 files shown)


ci/tasks/deploy_gcp.yaml
```diff
@@ -0,0 +1,28 @@
...
+  args:
+    - -exc
+    - |
+      export GOOGLE_APPLICATION_CREDENTIALS=~/gcloud-service-key.json
+      cat >$GOOGLE_APPLICATION_CREDENTIALS <<EOL
+      $SERVICE_ACCOUNT_JSON
+      EOL
+      
+      gcloud auth activate-service-account --key-file $GOOGLE_APPLICATION_CREDENTIALS
+      gcloud config set project "${PROJECT_ID}"
```
MebinAbraham, line 26: Is this even needed?. Buckets are globally scoped. I think we should pass in the BUCKET_NAME instead of the project id.

ci/tasks/deploy_gcp.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+      gsutil -m cp -r ons-design-system-release-artifact/** gs://census-eq-cdn/design-system/
```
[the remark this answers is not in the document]
ajmaddaford: The main reason for it being fixed was that we will only have one CDN for released static content. The only reason the development one is being used now is that the production one isn't set up. 

I hadn't thought about it needing to be configurable for testing though. I'll change it to be configurable as the variables file in the pipeline will provide the default.

ci/tasks/deploy_gcp.yaml
ajmaddaford: I hadn't really thought about it, I assumed all gcloud commands required the project to be set. Although the bucket name is global, it still exists within a project. I'll try it without and see what happens.

ci/tasks/deploy_gcp.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - -exc
+    - |
+      export GOOGLE_APPLICATION_CREDENTIALS=~/gcloud-service-key.json
+      cat >$GOOGLE_APPLICATION_CREDENTIALS <<EOL
+      $SERVICE_ACCOUNT_JSON
+      EOL
+      
+      gcloud auth activate-service-account --key-file $GOOGLE_APPLICATION_CREDENTIALS
+
+      gsutil -m cp -r ons-design-system-release-artifact/** gs://census-eq-cdn/design-system/
```
MebinAbraham, line 26: Is the bucket name being introduced as an env var?

ajmaddaford: Yes.

ci/tasks/deploy_gcp.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - -exc
+    - |
+      export GOOGLE_APPLICATION_CREDENTIALS=~/gcloud-service-key.json
+      cat >$GOOGLE_APPLICATION_CREDENTIALS <<EOL
+      $SERVICE_ACCOUNT_JSON
+      EOL
+      
+      gcloud auth activate-service-account --key-file $GOOGLE_APPLICATION_CREDENTIALS
+
+      gsutil -m cp -r ons-design-system-release-artifact/** gs://${BUCKET_NAME}/design-system/
```
MebinAbraham, line 27: By default, the files are [cached for 1hr](https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata#cache-control), is it worth caching for longer even for dev, given we will use these env for stress test etc. Impact should be negligible but for prod, the cache will be much higher than 1 hr.

ajmaddaford: Good spot. This was copied from the original s3 sync and I didn't think to check. Given that this is a deployment of a release, it makes sense to set a long cache time. I have set it to 1 year.

PR #795, opened 2020-06-24
Added new experimental pattern for finding a support centre

trevorsaint:
This pattern allows respondents to find a digital support centre close to them using a postcode.


src/patterns/find-a-support-centre/examples/results/_support-centre.scss
```diff
@@ -0,0 +1,42 @@
...
+  display: block;
+  margin-bottom: 0;
+
+  p:last-child {
+    margin-bottom: 0;
+  }
+}
+
+.support-centre__item:not(:first-child) {
+  border-top: 1px solid #d8d8d8;
```
rmccar, line 26: Can this colour be se using one of the variables in the styles

src/patterns/find-a-support-centre/examples/results/_support-centre.scss
```diff
@@ -0,0 +1,42 @@
...
+  padding-top: 32px;
+}
+
+.support-centre__item:not(:last-child) {
+  padding-bottom: 32px;
+}
+
+.support-centre__distance {
+  background-color: rgba(#00a3a6, 0.2);
```
rmccar, line 35: Can this colour be added to the style as a new colour

src/patterns/find-a-support-centre/examples/results/index.njk
```diff
@@ -0,0 +1,130 @@
...
+            <p>Find us on <a class="icon--external-link" href="https://goo.gl/maps/X48tjZkXA4RaN3LL7" target="_blank" rel="noopener nofollow">Google Maps</a> <br> Telephone: 01142 734 712</p>
+
+            {% set centreDetails %}                          
+            <h2 class="u-fs-r--b u-mb-xs">Opening Times</h2>
+
+            <p>9am to 4pm - Monday to Friday <br> 9am to 5pm - Saturday to Sunday (Census weekend)</p>
+            
+            <h2 class="u-fs-r--b u-mb-xs">Centre facilities</h2>
+
+            <ul class="support-centre__list">
```
rmccar, line 68: Think we should use the list component here

src/patterns/find-a-support-centre/examples/results/index.njk
```diff
@@ -0,0 +1,130 @@
...
+            <p>Find us on <a class="icon--external-link" href="https://goo.gl/maps/SEaVUjW2KY9wDoPE8" target="_blank" rel="noopener nofollow">Google Maps</a> <br> Telephone: 0114 222 2000</p>
+
+            {% set centreDetails %}                            
+            <h2 class="u-fs-r--b u-mb-xs">Opening Times</h2>
+
+            <p>8am to 6pm - Monday to Friday</p>
+            
+            <h2 class="u-fs-r--b u-mb-xs">Centre facilities</h2>
+
+            <ul class="support-centre__list">
```
rmccar, line 107: And use the list component here

src/scss/patternlib.scss
```diff
@@ -6,6 +6,9 @@
 @import '../styles/page-template/examples/block-areas/block-areas';
 @import '../patterns/question/examples/question-anatomy/pl-question-anatomy';
 
+// Find a support centre specific styles (experimental)
```
rmccar, line 9: Does it need to have its own style imported here instead of in the component itself?

src/patterns/find-a-support-centre/examples/results/_support-centre.scss
trevorsaint: Yes this can use the `$color-grey-3`. I will update this.

src/patterns/find-a-support-centre/examples/results/index.njk
trevorsaint: Unfortunately not. The items sit inside a `<ul>`.

So you get the `<ul>` inside a `<ul>` style. Ideally we would assign modifier classes for bullet styling like so:

```
<ul class="list list--bullet">
  <li class="list__item">Item 1</li>
</ul>
```

src/patterns/find-a-support-centre/examples/results/index.njk
trevorsaint: Same situation applies as above. We need to modify the global `<ul>` to be more flexible.

src/scss/patternlib.scss
trevorsaint: It is not a component. More a pattern. It is also experimental, so was unsure where to best place the styles associated.

src/patterns/find-a-support-centre/examples/results/_support-centre.scss
trevorsaint: This is a Census specific tint of a colour and does not seem to be available. Perhaps we can have Census specific colours if we have not already got them. They don’t seem to be available from the documentation.

boxadesign: We do have Census specific colour variables. If you look in `scss/settings/` you'll find a `census.scss` file which has all the census specific variables.

src/patterns/find-a-support-centre/examples/results/index.njk
boxadesign: I'm not sure I'd use a list at the higher level (this is always a much debated thing). If a list item contains multiple elements (headings, para's etc) and include another list that isn't related to its parent `li` I'm not convinced how much semantic meaning it provides. I think the headings provide the content hierarchy alone.

src/patterns/find-a-support-centre/examples/results/_support-centre.scss
boxadesign: I'm not sure we need this stylesheet? It looks like many of these styles could be sorted with utility classes. 

For any styles we do need (the distance element) I think we could possible drop the `support-centre` bit and keep it generic.

src/scss/patternlib.scss
boxadesign: I've left a comment further up about the support centre specific styles. If they are needed then it would become a component without it's own page (index.njk) and then called from your patterns page. Essentially it would just be a folder with a `scss` file in it. 

The `patternlib.scss` file is only for styles specific to the design system website so wouldn't be bundled and deployed to the cdn. 

But as I said in a previous comment I'm not sure this requires it's own stylesheet.

src/patterns/find-a-support-centre/examples/results/_support-centre.scss
trevorsaint: I have removed anything specific to support-centre and used utility classes as you suggested. I have also added the supporting colour for Census (teal) and a tint of 20%. This needs to be revisited at some point.

src/patterns/find-a-support-centre/examples/results/index.njk
trevorsaint: The HTML structure has been changed. Agreed the heading structure alone should cater for this. As a result of this change the list item now renders correctly, although it would be worth looking at the list component styling to make it more flexible for the future.

src/patterns/find-a-support-centre/examples/results/index.njk
```diff
@@ -0,0 +1,130 @@
...
+        <p>Find us on <a class="icon--external-link" href="https://goo.gl/maps/X48tjZkXA4RaN3LL7" target="_blank" rel="noopener nofollow">Google Maps</a> <br> Telephone: 01142 734 712</p>
+
+        {% set centreDetails %}                          
+        <h3 class="u-fs-r--b u-mb-xs">Opening Times</h3>
+
+        <p>9am to 4pm - Monday to Friday <br> 9am to 5pm - Saturday to Sunday (Census weekend)</p>
+        
+        <h3 class="u-fs-r--b u-mb-xs">Centre facilities</h3>
+
+        <ul class="list">
```
rmccar, line 67: Now you've got rid of the higher up list shouldn't we be making use of the list component here and in the one below?

src/components/tag/_tag.scss
```diff
@@ -0,0 +1,9 @@
+@import 'settings/census';
```
boxadesign, line 1: We shouldn't need to do this. If the page template has the theme "census" the variables should be applied automatically. The way the theme styles work is to create a variable in the base `vars/colors` file and provide a default colour for that variable. The variable is then overwritten in the theme scss file.

src/patterns/find-a-support-centre/examples/results/index.njk
```diff
@@ -0,0 +1,130 @@
...
+    <div class="u-pb-m u-pt-m">
+
+        <h2 class="u-fs-l u-mb-xs">Sheffield Central Library</h2>
+        <span class="u-mb-s tag">0.7 miles away</span>
+
+        <p>First Floor, Central Library <br> Surrey Street <br> Sheffield <br> S1 1XZ</p>
+
+        <p>Find us on <a class="icon--external-link" href="https://goo.gl/maps/X48tjZkXA4RaN3LL7" target="_blank" rel="noopener nofollow">Google Maps</a> <br> Telephone: 01142 734 712</p>
+
+        {% set centreDetails %}
```
boxadesign, line 60: Is there any purpose to the centreDetails block? As it's not a template that gets extended I don't think this is doing anything.

src/components/tag/_tag.scss
trevorsaint: I will try this again. For some reason it was not working for me.

src/patterns/find-a-support-centre/examples/results/index.njk
trevorsaint: This has been modified.

src/components/tag/_tag.scss
trevorsaint: This is now working. I added the colour into 'vars/colors' also as you suggested.

src/patterns/find-a-support-centre/examples/results/index.njk
trevorsaint: I missed the list component. I have updated the lists to use the component.

src/patterns/find-a-support-centre/examples/results/index.njk
```diff
@@ -0,0 +1,144 @@
...
+                "chooseLanguage": "Dewiswch iaith",
+                "allLanguages": "Pob iaith",
+                "current": false
+            }
+        ]
+    }
+} %}
+
+{% block main %}
+
```
jrbarnes9, line 48: Consider adding a previous link to this example so navigation back to postal input is clear.

src/components/tag/_tag.scss
```diff
@@ -0,0 +1,7 @@
+.tag {
+  @extend .u-fs-r;
+  @extend .u-p-xxs;
+  background-color: $color-branded-supporting-tint;
```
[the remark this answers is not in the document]
jrbarnes9: Can we create this as a macro and use this new component to mark up this pattern as 'experimental'?

src/patterns/find-a-support-centre/examples/find/index.njk
```diff
@@ -0,0 +1,80 @@
...
+            }
+        ]
+    }
+} %}
+
+{% block main %}
+
+    <h1 class="u-mt-l u-fs-xxl">Find a support centre</h1>
+
+    <div class="col-8@m">
```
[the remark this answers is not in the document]
trevorsaint: Yes I can see that. Will remove.

src/patterns/find-a-support-centre/examples/find/index.njk
```diff
@@ -0,0 +1,80 @@
...
+                    "autocomplete": 'postal-code',
+                    "classes": 'input--w-5 u-mb-m'
+                })
+            }}
+
+            {{
+                onsButton({
+                    "type": 'submit',
+                    "text": 'Find',
+                    "classes": 'btn--small'
```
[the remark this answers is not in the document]
trevorsaint: Agreed I did feel that the button was too big and did not look right.

src/patterns/find-a-support-centre/examples/results/index.njk
```diff
@@ -0,0 +1,144 @@
...
+    <h1 class="u-mt-l u-fs-xxl">Support centres near <em class="highlight">S1 1XN</em></h1>
+
+    <div class="u-pb-m u-pt-m">
+
+        <h2 class="u-fs-l u-mb-xs">Sheffield Central Library</h2>
+        <span class="u-mb-s tag">0.7 miles away</span>
+
+        <p>First Floor, Central Library <br> Surrey Street <br> Sheffield <br> S1 1XZ</p>
+
+        <p>Find us on <a class="icon--external-link" href="https://goo.gl/maps/X48tjZkXA4RaN3LL7" target="_blank" rel="noopener nofollow">Google Maps</a> <br> Telephone: 01142 734 712</p>
```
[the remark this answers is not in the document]
trevorsaint: Lauren did look at the content. But can change as it does see more fitting.

src/patterns/find-a-support-centre/examples/results/index.njk
```diff
@@ -0,0 +1,144 @@
...
+
+        {% call onsCollapsible({
+            "id": "collapsible-1",
+            "title": "Why your answer is important",
```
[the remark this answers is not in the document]
trevorsaint: Yeah just spotted that :(.

src/patterns/find-a-support-centre/examples/results/index.njk
```diff
@@ -0,0 +1,144 @@
...
+            "button": {
+                "close": "Opening times and facilities"
```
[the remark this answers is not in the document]
trevorsaint: Yeah just spotted that :(.

src/components/tag/_tag.scss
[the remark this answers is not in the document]
trevorsaint: Can we revisit this later. We can look at a more reusable pattern. [GDS have done something similar](https://design-system.service.gov.uk/components/tag).

src/patterns/find-a-support-centre/examples/results/index.njk
trevorsaint: This pattern needs work.

src/patterns/find-a-support-centre/examples/results/index.njk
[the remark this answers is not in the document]
trevorsaint: I have adjusted this pattern and it looks much better.

src/patterns/find-a-support-centre/examples/results/index.njk
rmccar: Unresolving I don't think the previous link has been added

src/patterns/find-a-support-centre/examples/find/index.njk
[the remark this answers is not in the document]
rmccar: I think the button guidance still needs updating around this

Issue #783, opened 2020-06-18
List numerals should use a monospaced variant of the font to ensure lists line up...


jrbarnes9:
List numerals should use a monospaced variant of the font to ensure lists line up...
![Screenshot 2020-06-18 at 08 46 55](https://user-images.githubusercontent.com/43346934/84993542-537e5400-b141-11ea-8820-1cd2db183430.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/30#issuecomment-645848135_

Issue #640, opened 2019-10-17
Add monospaced numerals for data tables


jrbarnes9:
The .ttf format of Open Sans can implement `font-variant-numeric: tabular-nums;` which activates the set of figures where numbers are all of the same size, allowing them to be easily aligned in tables. It's required according to the ONS Styleguide : https://style.ons.gov.uk/category/data-visualisation/tables/#fonts

Our version doesn't implement it.

OpenSans (.woff) 
![image](https://user-images.githubusercontent.com/1080383/66906436-9e1e5e00-efff-11e9-8a09-ff8d7a4507ec.png)

Open Sans (.ttf)
![image](https://user-images.githubusercontent.com/1080383/66906477-b2faf180-efff-11e9-84f5-62289c173eee.png)

Looks like our font doesn't contain the monospaced `1` glyph

See what glyphs are missing by dropping our font in here ... https://fontdrop.info/

_Originally posted by @jeremygillies in https://github.com/ONSdigital/design-system/issues/233#issuecomment-542616234_

PR #799, opened 2020-06-26
Update list numerals to use monospace and update error example content

rmccar:
This PR updates the base font for the Design System to use monospacing so that numbered lists can be correctly formatted. It also updates the content of the error lists in examples and as a side effect of all this it fixes an issue around monospaced numbers for data tables.


rmccar:
[Issue #783: List numerals should use a monospaced variant of the font to ensure lists line up...](https://trello.com/c/Ul0wP2Pw/527-issue-783-list-numerals-should-use-a-monospaced-variant-of-the-font-to-ensure-lists-line-up)

rmccar:
[Update error pattern](https://trello.com/c/sZmntAeK/530-update-error-pattern)

rmccar:
[Issue #640: Add monospaced numerals for data tables](https://trello.com/c/xwaDSthy/443-issue-640-add-monospaced-numerals-for-data-tables)

src/components/lists/_macro.njk
```diff
@@ -9,7 +9,7 @@
       {% for item in (params.itemsList if params.itemsList is iterable else params.itemsList.items()) %}
           <li class="list__item {{ item.listclass }}" {% if item.current %}aria-current="true"{% endif %}>
               {% if item.url and item.current != true %}
-                  {% if item.index %}{{ loop.index }}. {% endif %}{% if item.prefix %}<span class="list__prefix" aria-hidden="true">{{ item.prefix }}.</span> {% endif %} <a href="{{ item.url }}" class="list__link {{ item.classes }}" {% if item.index %}data-qa="list-item-{{ loop.index }}"{% endif %} {% if item.target %} target="{{ item.target }}"{% endif %} {% if item.name %} name="{{ item.name }}"{% endif %}>
+                  <span class="list__prefix" aria-hidden="true">{% if item.index %}{{ loop.index }}. {% endif %}{% if item.prefix %}{{ item.prefix }}. {% endif %}</span><a href="{{ item.url }}" class="list__link {{ item.classes }}" {% if item.index %}data-qa="list-item-{{ loop.index }}"{% endif %} {% if item.target %} target="{{ item.target }}"{% endif %} {% if item.name %} name="{{ item.name }}"{% endif %}>
```
boxadesign, line 12: We are outputting the prefix before checking for the prefix param...is this right?

rmccar: Yeah thats probably not right. I just wanted to make sure that if it is set using the `{{ loop.index }}` then it is also included in a span and has the correct classes. Ill rework this

src/scss/base/_fonts.scss
```diff
@@ -1,4 +1,4 @@
-@include font-face(OpenSans, '#{$static}/fonts/opensans-regular', $weight: 400, $style: null, $exts: woff2 woff);
-@include font-face(OpenSans, '#{$static}/fonts/opensans-bold', $weight: 600, $style: null, $exts: woff2 woff);
+@include font-face(OpenSans, '#{$static}/fonts/opensans-regular', $weight: 200, $style: null, $exts: ttf);
```
jrbarnes9, line 1: I think we probably want to keep both woff and ttf formats and only apply to monospaced numerals format to ordered lists and table cells?

rmccar:
> Can you update error validation documentation `The error `panel__header` must display the number of errors on the page _if there is more than one_.`?

I'm not sure I like the "if there is more than one" being bold but ill push it and see what you think

PR #804, opened 2020-06-29
Add more metadata

boxadesign:
Added some more SEO related properties to the base template


src/styles/page-template/_template.njk
```diff
@@ -22,7 +24,7 @@
 {% else %}
     {# Development #}
     {% set assetsURL = "" %}
-{% endif %}
+{% endif %}
```
rmccar, line 27:
```suggestion
{% endif %}
```

PR #806, opened 2020-06-30
Make CDN URL configurable

MebinAbraham:
Runner requires the ability to specify its own CDN URL, so this PR introduces a CDN object which can contain the URL for the CDN and also the version of the CDN (previously known as just CDN in config)

### How to review 
Ensure you can pass your own CDN and that when no CDN is provided, the default ONS CDN is used.


src/styles/page-template/_template.njk
```diff
@@ -12,7 +12,8 @@
 
 {% if pageConfig.cdn or release_version %}
     {# Production #}
-    {% set assetsURL = "https://cdn.ons.gov.uk/sdc/design-system/" + (pageConfig.cdn or release_version) %}
+    {% set cdn_url = pageConfig.cdn.url or "https://cdn.ons.gov.uk/sdc/design-system/" %}
+    {% set assetsURL = cdn_url + (pageConfig.cdn.version or release_version) %}
```
MebinAbraham, line 16: @rmccar @boxadesign Are you aware of any services that are using `pageConfig.cdn` at the moment for the `version`?

boxadesign: Yes so this would be a breaking change which we can handle ok.

ajmaddaford: When would we want to use a different version of the CDN assets to the template `release_version`?

MebinAbraham: We wouldn't but others might for whatever reason. I'm just supporting what was already supported. Previously the var was just called `cdn` which is misleading.

MebinAbraham: @boxadesign just to confirm, other services make use of the `cdn version`?. For any given version of templates you load, the `release_version` would match. Would you happened to know of the use case for this?

boxadesign: So the `release_version` gets set if you use the design system components via the zip that gets generated on release. You can of course use the DS is a few other ways i.e. NPM or manually and you can set the cdn version via the `pageConfig`

src/styles/page-template/_template.njk
```diff
@@ -14,7 +14,8 @@
...
```
ajmaddaford, line 18: This will break if `pageConfig.cdn.url` doesn't have a trailing slash. We should remove the trailing slash from the default cdn url and add a slash before the version. That way, if someone does provide a trailing slash with a cdn url it will carry on working (with a `//` in the url).

MebinAbraham: I did think about that but that they would be more to do with knowing what you need to pass. But I'm happing to prefix with `/` just so that it works either way.

MebinAbraham: Ad discussed, it didn't work with `//` in the URL, therefore I added a small check to add or not add a trailing slash if required.

src/styles/page-template/_template.njk
```diff
@@ -14,7 +14,9 @@
 
 {% if pageConfig.cdn or release_version %}
     {# Production #}
-    {% set assetsURL = "https://cdn.ons.gov.uk/sdc/design-system/" + (pageConfig.cdn or release_version) %}
+    {% set cdn_url = pageConfig.cdn.url or "https://cdn.ons.gov.uk/sdc/design-system" %}
+    {% set trailing_slash = "/" if cdn_url | last != "/" else "" %}
+    {% set assetsURL = cdn_url + trailing_slash + (pageConfig.cdn.version or release_version) %}
```
ajmaddaford, line 19: Minor, the `/` isn't trailing as we're using it between two strings, and I think flipping the logic to be positive makes it clearer.
```suggestion
    {% set slash = "" if cdn_url | last == "/" else "/" %}
    {% set assetsURL = cdn_url + slash + (pageConfig.cdn.version or release_version) %}
```

PR #808, opened 2020-07-02
Add content attributes to collapsibles

rmccar:
This is to allow content attributes to be set on collapsibles. This is to allow `data-qa` attributes to be set on the content in runner to allow selectors to use them for tests

### How to review 
Tests pass, can set an attribute on a collapsible


rmccar:

PR #813, opened 2020-07-07
Add Summary docs and move rowTitleAttributes

rmccar:
To add docs for summary attributes and move rowTitleAttributes to be at the same level as rowTitle


rmccar:

Issue #779, opened 2020-06-17
We should remove the 'current page' from the breadcrumb


trevorsaint:
We should remove the 'current page' from the breadcrumb. It adds no real value other than visual distraction.

**Problem**
![image](https://user-images.githubusercontent.com/4989027/84881773-3cc6f700-b086-11ea-8b86-0b33cf06711b.png)

**Improvement**
![image](https://user-images.githubusercontent.com/4989027/84881877-6122d380-b086-11ea-87d5-06fd82ef0915.png)

GDS have removed the current page.
https://design-system.service.gov.uk/components/breadcrumbs/

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/issues/43#issuecomment-645268115_

PR #815, opened 2020-07-07
Remove current from breadcrumbs component

rmccar:
Remove current from breadcrumbs component


rmccar:
[Issue #779: We should remove the 'current page' from the breadcrumb](https://trello.com/c/1fCqMnGU/525-issue-779-we-should-remove-the-current-page-from-the-breadcrumb)

PR #821, opened 2020-07-08
Tidy up sass and docs for breadcrumbs

rmccar:
- Removed redundant css
- Added guidance on when to use breadcrumbs
- Fixed bug on mobiles that was showing the last but one link, not the last one
- Updated style
- Changed desktop version to display left chevron before link if only one link

### How to review 
- Tests pass
- Check mobile version shows correct link

### DIFF
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -60,5 +60,9 @@ $breadcrumb-chevron-height: 0.65rem;
   &__link {
     color: $color-text;
     text-decoration: underline;
+
+    &:hover {
+      text-decoration: none;
+    }
   }
 }
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -58,7 +58,7 @@ $breadcrumb-chevron-height: 0.65rem;
   }
 
   &__link {
-    color: #222;
+    color: $color-text;
     text-decoration: underline;
   }
 }
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -49,6 +49,7 @@ $breadcrumb-chevron-height: 0.65rem;
   }
 
   &__link {
-    text-decoration: none;
+    color: #222;
+    text-decoration: underline;
   }
 }
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -22,7 +22,7 @@ $breadcrumb-chevron-height: 0.65rem;
     &:before {
       @include icon('chevron-left', $breadcrumb-chevron-height, $breadcrumb-chevron-height);
       content: '';
-      margin: 0 0.2rem 0 0;
+      margin: 0 0.2rem 0 -0.14rem;
       vertical-align: middle;
     }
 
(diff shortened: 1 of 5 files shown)


jrbarnes9:
[Update breadcrumb](https://trello.com/c/wtfNwyE3/556-update-breadcrumb)

src/components/breadcrumb/_breadcrumb.scss
```diff
@@ -31,28 +31,34 @@ $breadcrumb-chevron-height: 0.65rem;
...
     }
-
-    &--current {
-      color: $color-text-light;
-    }
   }
 
   &__link {
-    text-decoration: none;
+    color: #222;
```
jrbarnes9, line 61: Use variable: 
```suggestion
    color: $color-text;
```

src/components/breadcrumb/index.njk
```diff
@@ -11,7 +11,8 @@ A breadcrumb trail consists of a list of links to the parent pages of the curren
 }}
 
 ## When to use this component
-Use the breadcrumbs component when you need to help users understand and move between the multiple levels of a website.
+Use the breadcrumbs component when you need to help users understand and move between the multiple levels of a website. We recomend not using it on
+pages that are only one level deep in the page hierarchy. As we would like to avoid 'home' being the only link, when it’s directly under a menu that also has ‘home’.
```
jrbarnes9, line 15:
```suggestion
Use the breadcrumbs component when you need to help users understand and move between the multiple levels of a website.
```
I think we'll have to remove this comment about not displaying it when home is in the nav - it's a trade-off but I think the convention and consistency of always retaining the breadcrumb is more important than the lack of value in this small case

src/components/breadcrumb/_breadcrumb.scss
```diff
@@ -31,28 +31,34 @@ $breadcrumb-chevron-height: 0.65rem;
...
+    text-decoration: underline;
```
jrbarnes9, line 62: Removing the underline on hover is needed for a11y (shouldn't rely on colour change alone) but we could leave for a full sweep of this in: https://trello.com/c/mA0eWaT8/545-issue-803-dont-use-colour-as-the-only-indicator-of-hover-states

src/components/breadcrumb/_breadcrumb.scss
```diff
@@ -31,28 +31,34 @@ $breadcrumb-chevron-height: 0.65rem;
         content: none;
       }
 
-      &:not(:nth-last-child(2)) {
+      &:not(:nth-last-child(1)) {
         display: inline-block;
       }
 
+      &:first-child:nth-last-child(1) {
```
jrbarnes9, line 38: This chevron needs to be left-aligned flush with edge of content block

src/components/breadcrumb/_breadcrumb.scss
jrbarnes9: ...or probably a new variable for breadcrumbs?

trevorsaint: I think the breadcrumb is always going to be the text colour, so it makes sense to use the existing text variable.

rmccar:
> Looking at the HTML It might be worth setting the font scale in the component itself rather than using utility classes for each item. We don’t really want to allow control of the breadcrumb font scale.
> 
> Also why are IDs being applied for each item?
> 
> **It might be nice adding a little negative margin to bring the alignment back. I appreciate it is only a few pixels, but its the small details :)**
> image
> 
> Negative 2px should do the trick.
> image

Ok I've fixed the margin and utility classes issues. The IDs I'll leave in. I asked, and it looks like they might be needed for craft

Issue #827, opened 2020-07-10
We need to remove the underline when the pager is in focus.


trevorsaint:
We need to remove the underline when the pager is in focus.
![image](https://user-images.githubusercontent.com/4989027/87150686-74eadf80-c2aa-11ea-8bc1-39306b16890f.png)

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/issues/70#issuecomment-656630921_

PR #828, opened 2020-07-10
Fixed focus state styling for pager

trevorsaint:
I have removed the underline of the label on focus as it is not required. It is also in keeping with how focus works with other links.

### How to review 
From this
image

To this
image


jrbarnes9:
[Issue #827: We need to remove the underline when the pager is in focus.](https://trello.com/c/d2l4DMpO/562-issue-827-we-need-to-remove-the-underline-when-the-pager-is-in-focus)

PR #829, opened 2020-07-10
Add ability to set icons before panel content

rmccar:
This PR is to add the ability to set icons before panel content. Mainly for the need to add a tick icon to the success panel.

### How to review 
- Tests pass
- Check panels look good with and without icon set


rmccar:
[Add green check icon to success panel as an option](https://trello.com/c/e0x66T4B/547-add-green-check-icon-to-success-panel-as-an-option)

PR #660, opened 2019-11-22
Added clear radios functionality and visible ins

boxadesign:
https://deploy-preview-660--ons-design-system.netlify.com/components/radios/examples/radios-clear/

PR #836, opened 2020-07-15
Add 'is defined' to input component label

### LINKED ISSUE #687: Input fields with type=hidden currently require a label to be provided
**Describe the bug**
onsInput can be set to have a type of hidden, but the macro also then requires a label. A label is not needed for hidden fields, so it should be possible to not specify one.

**To Reproduce**

1. Add an onsInput call to code
2. Set type to hidden
3. Do not specify a label
4. Rendering errors with jinja2.exceptions.UndefinedError: ‘dict object’ has no attribute ‘label’

**Expected behavior**
A hidden input field should be created without needing to provide a value for label

**Additional context**
The use of type=hidden is not currently documented. Can this be added to the documentation.

rmccar:
Fixes #687 Jinja error when an input type is set to hidden and label not provided

### How to review 
- Make sure input component works as expected
- Test in a environment running Jinja that you can set a hidden input without a label


rmccar:
[Issue #687: Input fields with type=hidden currently require a label to be provided](https://trello.com/c/bqkAqZxY/566-issue-687-input-fields-with-typehidden-currently-require-a-label-to-be-provided)

src/components/input/_macro.njk
```diff
@@ -47,7 +47,7 @@
         {% endif %}
     {% endset %}
     {% set field %}
-        {% if params.label or not (params.prefix or params.suffix or type == 'password') %}
+        {% if params.label is defined %}
```
boxadesign, line 50: Surprised we can remove the other conditions. I've had a look on the netlify deployment and it looks ok!?

rmccar: Yeah I was too, I'm not sure about removing them because they had to have been put there for a reason. But it was the only way I could find of fixing the Jinja error. Just thought I'd create a PR to see what you thought. I'll see if I can find out why they were added.

rmccar: https://github.com/ONSdigital/design-system/pull/660/commits/cbbf1e50f2b8a04eb0a113440b8648d8bd2a6aa7 so the password bit was added in this commit. Ive tested the other stuff in that PR without the password bit and it seems to all work ok

rmccar: https://github.com/ONSdigital/design-system/pull/503/commits/84d1b82f40183e80a9da679fc0f4a943a399eb1b the prefix suffix stuff was added in this one. I've tested prefixed and suffixed inputs and they seem to work with this change too

rmccar: Maybe if you do a test yourself too and see if you can find any problems with it?

PR #839, opened 2020-07-15
Add 'is defined' to all components

rmccar:
Errors have been happening when using some components in a Jinja environment. The answer to this is usually to add "is defined" to if statements checking for params. So this PR is to add them in all places to prevent these errors in future.

### How to review 
- Tests Pass
- Check that all components and patterns are still working correctly


src/components/collapsible/_macro.njk
```diff
@@ -3,21 +3,21 @@
...
-        {% if params.group %} data-group="{{ params.group }}"{% endif %}
-        {% if params.attributes %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
-        {% if params.saveState %} data-save-state="true"{% endif %}
+        class="collapsible js-collapsible{% if params.isAccordion is defined %} collapsible--accordion{% endif %}{% if params.classes is defined %} {{ params.classes }}{% endif %}"
+        {% if params.button is defined and params.button.close is defined %} data-btn-close="{{ params.button.close }}"{% endif %}
+        {% if params.group is defined %} data-group="{{ params.group }}"{% endif %}
+        {% if params.attributes is defined %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value is defined %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
+        {% if params.saveState is defined and params.saveState %} data-save-state="true"{% endif %}
```
jrbarnes9, line 10: duplicate param?
```suggestion
        {% if params.saveState is defined %} data-save-state="true"{% endif %}
```

src/components/field/_macro.njk
```diff
@@ -1,25 +1,25 @@
...
+                class="field{% if params.classes is defined %} {{ params.classes }}{% endif %}"
+                {% if params.attributes is defined %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value is defined %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
             >
                 {{ caller() }}
             </p>
         {% endif %}
     {% endset %}
 
-    {% if params.error %}
+    {% if params.error and params.error is defined %}
```
jrbarnes9, line 22: Duplicate param?
```suggestion
    {% if params.error is defined %}
```

rmccar: Nah I intentionally did that, its because "is defined" tests if its defined and "if params.error" just tests if its truthy

src/components/collapsible/_macro.njk
rmccar: Same as comment below

src/components/field/_macro.njk
rmccar: After having a chat with Matt, maybe this isn't needed

rmccar: Remembered why I put these in now. Tests fail without them haha. So I've put them back in

rmccar:
[Add 'is defined' to all components](https://trello.com/c/bm47QMKh/602-add-is-defined-to-all-components)

PR #847, opened 2020-07-21
Refactor to use inline SVGs for icons

### LINKED ISSUE #835: Introduce inline SVGs
I think we need to revisit the way we use icons. We should be introduce inline SVGs.

We are using the pseudo after attribute with a background SVG. Zoom is fine. However for something like focus, we are unable to change the colour attribute and have limitations. An inline SVG would provide greater flexibility. Colour could be inherited, we could also change the icon colour without changing the image.

**Resting state**
image

**Focus state**
image

The icon should be black on focus.

rmccar:
This is to change icons to be set using inline SVGs.
### How to review 
- Tests pass
- Components that use icons still work as expected


rmccar:
[Issue #835: Introduce inline SVGs](https://trello.com/c/SVJv66uG/568-issue-835-introduce-inline-svgs)

rmccar:
[Add inline SVGs across the Design System](https://trello.com/c/6Nlj7lgz/614-add-inline-svgs-across-the-design-system)

PR #857, opened 2020-07-28
Allow mutually exclusive input with charCheck 

rmccar:
This is a fix requested by the runner team to allow them to use a mutually exclusive field with an input field that has a charCheckLimit. This PR will also change textfield to also use the charCheckLimit macro like input to make it more consistent.

### How to review 
- Test to see that an input field with a charCheckLimit and mutually exclusive input behaves as expected
- Test to see if all other components that use charCheckLimit (Input, Textarea, Search and Mutually exclusive) still behave correctly
- Tests pass


rmccar:
[Bug - max length on text field breaking mutually exclusive option](https://trello.com/c/xpLb4XjO/4039-bug-max-length-on-text-field-breaking-mutually-exclusive-option)

src/components/input/_macro.njk
```diff
@@ -110,18 +110,23 @@
...
-        }) %}
-            {{ field | safe }}
-        {% endcall %}
-    {% elif params.mutuallyExclusive %}
+    {% if params.charCheck is defined %}
+        {% set charCheckField %}
+            {% call onsCharLimit({
+                "id": params.id ~ "-check",
+                "limit": params.charCheck.limit,
+                "type": "limit",
```
boxadesign, line 118: Think you can get rid of this?

rmccar: That was put in because Input was using the charCheck macro but textarea was doing something similar but not using the macro. To standerdise the way charCheck was implemented I managed to get them both using the charCheck macro. But there were parts of the macro that wasn't needed for the textarea implementation.
So I used that variable to differentiate between using it on an input or textarea. Do you think we should go back to the way textareas were working originally? Or I could rename the variable? It was just to standerdise the way charChecks were being used

boxadesign: Right ok that sounds good. Does the textarea use it's own JS still or does it use the charcheck?

rmccar: Nah it still uses its own JS, maybe I should move the charcheck stuff out of input into its own folder and intergrate it with the textarea JS

iwootten: On textarea/mutually exclusive components limits are defined using the maxlength, charCountSingular and charCountPlural keys. 

Is there a reason why input can't use the same keys without the charCheck object? Otherwise we have to work round this difference in runner's templates which is why I think we may have this bug at the moment.

rmccar: Because input and textarea are two different components I think getting both to use the charCheck macro rather than having them each use their own implementation like textarea currently does makes sense to me as it will standardise the way character limits are handled.
The reason why we have the bug at the moment is because adding a char limit and mutually exclusive to inputs is currently mutually exclusive in itself. So you can either add one or the other. The changes I've made in this PR allow you to use both.

rmccar: To make it a bit easier for you to implement in runner. Ive got Input and Textarea both using `limit` now instead of one using `maxlength` and one using `limit`. I tried looking into getting the other ones the same too but there needs to be a charCountSingular and a charCountOverLimitSingular param because of the need for messages for over and under the char count as in [this](http://localhost:3030/components/search/examples/search-character-check) search example. I feel like this is the best solution to fix the bug and maintain as much consistency as possible.

rmccar:
[Update Runner to latest version of the Design System to fix the mutual exclusive/charCheck bug](https://trello.com/c/EJ1H8GmL/4045-update-runner-to-latest-version-of-the-design-system-to-fix-the-mutual-exclusive-charcheck-bug)

PR #862, opened 2020-07-31
Fix accessible placeholder

boxadesign:
The accessible placeholder (/components/search) wasn't working, possibly due to lots of work being done on inputs/labels lately.

### How to review 
Navigate to /components/search and look at the 2nd example. You should see the placeholder. Click the input and it should hide. Type some text and click outside the input and it should not display.


jrbarnes9:
Button needs padding when stacked 
Screenshot 2020-08-03 at 09 10 11

boxadesign:
> Button needs padding when stacked
> Screenshot 2020-08-03 at 09 10 11

Done

PR #869, opened 2020-08-03
Resources for partners

trevorsaint:
Added experimental pattern to allow users to filter and sort downloadable resources.


boxadesign:
Would it be possible to use components inside this i.e checkboxes etc?

rmccar:
Yeah I agree, it makes sense to use the other components

trevorsaint:
> Would it be possible to use components inside this i.e checkboxes etc?

I have been able to do this in part using the **onsCheckboxes** component.  I had to use the dontWrap in a fieldset feature due to the lack of flexibility with the fieldset. It needed to include additional markup to show active filters and selection.

jrbarnes9:
Minor content suggestion, but I think the filter headings should either all be singular or all plural.  I'd go for singular: Audience, Type, Language. Then the 'all' selection playback labels should all be plural: 'All audiences', 'All types', 'All languages'.  Actually, I'm not sure how beneficial these playback labels are in this case, where you can easily see all checkboxes.  This would be only beneficial for long lists of filters or if the filter blocks were collapsed in accordions.

Also not sure if 'Recruitment' is for recruiters or job-seekers?

jrbarnes9:
Could you maybe stack the 'results found' and 'sort by' elements under the smallest breakpoint, to avoid this wrapping?
Screenshot 2020-08-10 at 16 42 48

trevorsaint:
> Could you maybe stack the 'results found' and 'sort by' elements under the smallest breakpoint, to avoid this wrapping?
> Screenshot 2020-08-10 at 16 42 48

I have fixed this. I used the small breakpoint as you suggested for adjustment.

trevorsaint:
I have made those content adjustments (thanks). We’ve tested this pattern and it seems to work Jon. I think for now leave it in. We can test some more at a later date. Playback does seem to be a pattern that is used. 

On a previous project we built up a tag list for filters selected. GDS also do this kind of thing, labelled as facet tags. Perhaps we can discuss this in more detail at a design review.

src/patterns/download-resources/index.njk
```diff
@@ -0,0 +1,13 @@
...
+experimental: true
+---
+{% from "views/partials/example/_macro.njk" import patternlibExample %}
+
+Allow users to filter and sort downloadable resources.
+
+## Example
+{{
+    patternlibExample({ "path": "patterns/download-resources/examples/download-resources/index.njk" })
+}}
```
rmccar, line 13: "Research on this component" section needs to be added to fit with the style of other patterns and components. I asked you to take out the panel import because it wasn't being used but you'll need to bring it back in for this

rmccar, line 13:
```suggestion
}}

## Research on this component
{% call onsPanel() %}
    If you have conducted any user research using this component, please feed back your findings via the
    {{
        onsExternalLink({
            "url": "https://github.com/ONSdigital/design-system/issues/45",
            "linkText": "Design System forum"
        })
    }}
{% endcall %}

## Design System forum
{{
    onsExternalLink({
        "url": "https://github.com/ONSdigital/design-system/issues/45",
        "linkText": "Discuss 'Download Resources' on GitHub"
    })
}}
```

Youll have to create an issue and change the number on the links in this suggestion

src/patterns/download-resources/index.njk
```diff
@@ -0,0 +1,13 @@
+---
+title: Download resources
+group: Help users to...
+experimental: true
+---
+{% from "views/partials/example/_macro.njk" import patternlibExample %}
+
```
rmccar, line 7:
```suggestion
{% from "components/panel/_macro.njk" import onsPanel %}

```

src/patterns/download-resources/index.njk
trevorsaint: Alright will take a look at this.

src/patterns/download-resources/index.njk
trevorsaint: This has been added.

src/patterns/download-resources/index.njk
trevorsaint: This has been added.

PR #870, opened 2020-08-05
Fixes for changes brought in by is defined

rmccar:
Some fixes for bugs brought in by the "is defined" PR

### How to review 
- Check code
- Tests pass


rmccar:
> So you did need it in the end!

Haha yeah theres a few places where params are always set, even if they contain nothing

rmccar:
["Is defined" bugs](https://trello.com/c/uCxkZBT3/610-is-defined-bugs)

PR #872, opened 2020-08-05
Added new component 'timeline'

trevorsaint:
Added a new timeline component. 

This pattern has already been tested with a number of Local Authorities and has proved to be useful.

timeline

### DIFF
--- a/src/components/timeline/_macro.njk
+++ b/src/components/timeline/_macro.njk
@@ -1,11 +1,29 @@
 {% macro onsTimeline(params) %}
 
+    {% from "components/lists/_macro.njk" import onsList %}
+
     <div class="timeline">
 
         {% for item in params.items %}
         <div class="timeline__item">
+            
             <h2 class="timeline__heading">{{ item.heading }}</h2>
-            {{ item.content | safe }}
+            
+            {% if item.itemsList is defined %}
+            
+                {{
+                    onsList({
+                        "classes": 'list--bare',
+                        "itemsList": item.itemsList
+                    })
+                }}
+
+            {% else %}
+
+                {{ item.content | safe }}
+
+            {% endif %}
+
         </div>
         {% endfor %}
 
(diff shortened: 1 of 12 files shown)


src/components/timeline/_macro.njk
```diff
@@ -0,0 +1,14 @@
+{% macro onsTimeline(params) %}
+
+    <div class="timeline">
+
+        {% for item in params.items %}
+        <div class="timeline__item">
+            <h2 class="timeline__heading">{{ item.heading }}</h2>
+            {{ caller() if caller else item.content | safe }}
```
boxadesign, line 8: I don't think (?) `caller()` will work inside the loop in this way. You would need to `call` the timeline component multiple times for each item. If this is the case you would just need to keep `item.content | safe`

boxadesign: The `caller()` is mainly useful for including other components inside the content of the main component i.e. a list component instance inside an error component.

trevorsaint: Thanks Matt. Will update.

src/components/timeline/examples/timeline/index.njk
```diff
@@ -0,0 +1,28 @@
+{% from "components/timeline/_macro.njk" import onsTimeline %}
+
+{{
+    onsTimeline({
+    "items": [
+            {
+                "heading": 'September to October 2020',
+                "content": '<ul class=\"list--bare\"><li class=\"list__item\">Meet your Community Engagement Manager and set up regular catch up meetings</li><li class=\"list__item\">Local Authority Toolkit available online</li></ul>'
```
rmccar, line 8: I think we would want to reuse the list component here. Other than that looks good

trevorsaint: I have updated the timeline to use the onsList as optional. Thanks.

PR #873, opened 2020-08-05
Fix is defined bug on Label and Feedback

rmccar:
Another bug from the "is defined" PR

### How to review 
- Check code
- Tests pass
- Label and Feedback component looks as expected


rmccar:
["Is defined" bugs](https://trello.com/c/uCxkZBT3/610-is-defined-bugs)

PR #880, opened 2020-08-07
Add external link macro for inline SVGs

### LINKED ISSUE #835: Introduce inline SVGs

rmccar:
Fixes #835 
Adds inline svgs for external links and adds design system forum missing links

### How to review 
- Tests pass
- View code
- Check all pages and links display correctly
- Check that link icon is correct colour when focused on


rmccar:
[Issue #835: Introduce inline SVGs for external links](https://trello.com/c/SVJv66uG/568-issue-835-introduce-inline-svgs-for-external-links)

PR #882, opened 2020-08-07
Use inline SVGs for buttons

rmccar:
This PR changes buttons to use inline SVGs and also changes how print and exit buttons are defined to make it more simple for the user. The "loader" button has not been affected by this change and will be updated in a later PR

### How to review 
- Check all button types look and behave as expected
- Tests pass


rmccar:
[Refactor buttons to use inline SVGs](https://trello.com/c/i9iYDhPG/616-refactor-buttons-to-use-inline-svgs)

src/components/button/_macro.njk
```diff
@@ -2,22 +2,48 @@
...
       {% if params.url is defined and params.newWindow is defined %}target="_blank" rel="noopener"{% endif %}
       {% if params.attributes is defined %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{attribute}}="{{value}}" {% endfor %}{% endif %}
       >
-      <span class="btn__inner{% if params.innerClasses is defined %} {{ params.innerClasses }}{% endif %}{% if params.submitType != 'loader' %}{% if params.newWindow is defined %} icon--external-link {% elif params.exit is defined %} icon--exit {% endif %}{% endif %}">{{- params.html | safe if params.html else params.text -}}</span>
+      <span class="btn__inner{% if params.innerClasses is defined %} {{ params.innerClasses }}{% endif %}">
+        {% if iconPosition == "before" %}<svg><use xlink:href="/img/icons/icons--{{ iconType }}.svg#{{ iconType }}"></use></svg>{% endif %}
+        {{- params.html | safe if params.html is defined else params.text -}}
+        {% if iconPosition == "after" %}<svg><use xlink:href="/img/icons/icons--{{ iconType }}.svg#{{ iconType }}"></use></svg>{% endif %}
```
boxadesign, line 45: We were discussing on the DS stand up about making the path to icons configurable (so you could have a brand pack of icons). Thinking we could set the path at the `pageConfig` level and have a default.

rmccar: Yeah ok ill sort that

boxadesign:
I think it would be cleaner to check for `params.iconsPath` as a ternary i.e. `params.iconsPath ? params.iconsPath else '/img/foo`.

I understand the approach of setting the path in the header but something doesn't feel right as it's a component on it's own. I don't have a solution but I think it should also available in the `pageConfig` base template as most services use the DS in that way.

rmccar:
> I think it would be cleaner to check for `params.iconsPath` as a ternary i.e. `params.iconsPath ? params.iconsPath else '/img/foo`.
> 
> I understand the approach of setting the path in the header but something doesn't feel right as it's a component on it's own. I don't have a solution but I think it should also available in the `pageConfig` base template as most services use the DS in that way.

Moved this piece of work to a new branch

PR #886, opened 2020-08-10
Use inline svgs for panels and UAC

rmccar:
Updated Panel component to use inline SVGs. Also added the "bare" panel type to allow icons to easily be used next to a plain paragraph of text and used it in the UAC example for the security paragraph with the lock icon.

### How to review 
- View panel and UAC examples and make sure they look as they should


rmccar:
[Refactor Panels and UAC to use inline SVGs](https://trello.com/c/EvQ07Mki/619-refactor-panels-and-uac-to-use-inline-svgs)

src/components/panel/_macro.njk
```diff
@@ -1,43 +1,48 @@
...
+    {% if params.iconsPath is defined and params.iconsPath %}
+        {% set iconsPath = params.iconsPath %}
+    {% else %}
+        {% set iconsPath = "/img/icons/" %}
+    {% endif %}
 
     <div
-        class="panel{{ typeClass }}{{ iconClass }}{{ simpleClass }}{{ spaciousClass }}{{ classes }}"
-        {% if params is defined and params.attributes is defined %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{attribute}}="{{value}}" {% endfor %}{% endif %}{% if params is defined and params.id is defined %} id="{{params.id}}"{% endif %}
+        class="{% if params.type != bare %}panel{% endif %}{{ typeClass }}{{ iconClass }}{{ simpleClass }}{{ spaciousClass }}{{ classes }}"
```
boxadesign, line 27: Shouldn't the `panel` class be added in all scenarios as the parent class?

rmccar: Yeah true, the reason why I did that was because when using the `bare` type on the UAC pattern I kept getting a big gap because it was a gap added if you have one panel under another but ill try override that

PR #887, opened 2020-08-10
Updated the downloads pattern to match new look and feel. Added image…

trevorsaint:
Updated the look and feel of the **downloads component** to match new design. Added example thumbnails and PDF documents.
download-example


src/components/downloads/_downloads.scss
```diff
@@ -1,35 +1,37 @@
 .download {
```
rmccar, line 1: In this file I would include `.download__image`, `.download__image-link`, `.download__meta` and `.download__excerpt` inside the `.download` brackets and then replace the "download" in each of them with "&". to fit in with the style of other SCSS files we have which use Sass. Other than that it looks good

boxadesign:
Agree with Rich's comment other than that looks good!

src/components/downloads/_downloads.scss
trevorsaint: Nested. Sure no problem.

trevorsaint:
I am thinking it might be best to allow flexibility in the heading level.

I am using a H3 in this instance because it made sense for Resource for Partners. If being used elsewhere a H3 may not be suitable. What are your thoughts?

PR #892, opened 2020-08-11
Use inline SVGs for Lists

rmccar:
Refactor Lists to use inline SVGs

### How to review 
- Check lists work as expected and all examples look as expected


rmccar:
[Use inline SVGs for lists](https://trello.com/c/MTW36mZU/620-use-inline-svgs-for-lists)

PR #906, opened 2020-08-12
Use inline SVGs for Panels and UAC

rmccar:
Updated Panel component to use inline SVGs. Also added the "bare" panel type to allow icons to easily be used next to a plain paragraph of text and used it in the UAC example for the security paragraph with the lock icon.

### How to review
View panel and UAC examples and make sure they look as they should


rmccar:
[Refactor Panels and UAC to use inline SVGs](https://trello.com/c/EvQ07Mki/619-refactor-panels-and-uac-to-use-inline-svgs)

PR #908, opened 2020-08-12
Use inline SVGs for Buttons

rmccar:
This PR changes buttons to use inline SVGs and also changes how print and exit buttons are defined to make it more simple for the user. The "loader" button has not been affected by this change and will be updated in a later PR

### How to review
Check all button types look and behave as expected


rmccar:
[Refactor Buttons to use inline SVGs](https://trello.com/c/i9iYDhPG/616-refactor-buttons-to-use-inline-svgs)

boxadesign:
I'm probably missing something but if all svg code is now held with the svg macro why do we still have individual svg files?

rmccar:
> I'm probably missing something but if all svg code is now held with the svg macro why do we still have individual svg files?

Yeah, I'm going to do a mop up PR to remove them

rmccar:
[Refactor buttons to use inline SVGs](https://trello.com/c/pTZ4ZoXs/4062-refactor-buttons-to-use-inline-svgs)

PR #910, opened 2020-08-13
Add SVG macro

rmccar:
Added icon macro to allow us to use inline SVGs without using the `<use>` tag that doesn't work in Microsoft browsers

### How to review 
Try using macro and check that icon is returned. Any incorrect sizes will be fixed when integrating into components.


rmccar:
[Introduce SVG macro to solve IE and Edge problems](https://trello.com/c/x4201K4i/623-introduce-svg-macro-to-solve-ie-and-edge-problems)

PR #920, opened 2020-08-14
Use inline SVGs for Quotes

rmccar:
Refactor Quotes to use inline SVGs

### How to review
Check Quotes work as expected and example looks as expected


rmccar:
[Use inline SVGs for Quotes](https://trello.com/c/OiykyJu1/628-use-inline-svgs-for-quotes)

PR #921, opened 2020-08-14
Use inline SVG for Sortable Tables

rmccar:
Refactor Sortable Tables to use inline SVGs

### How to review
Check Sortable Tables work as expected and example looks as expected


rmccar:
[Use inline SVGs for Sortable Table](https://trello.com/c/0sUvrBVH/625-use-inline-svgs-for-sortable-table)

PR #927, opened 2020-08-18
Use inline SVG for Loader Buttons

rmccar:
Updates the Loader Button to use inline SVGs

### How to review 
Check that Loader Button works as expected


rmccar:
[Use inline SVGs for Loader Button](https://trello.com/c/zvj6NETi/627-use-inline-svgs-for-loader-button)

PR #928, opened 2020-08-18
Use inline SVGs for Breadcrumbs

rmccar:
Updates Breadcrumbs to use inline SVGs

### How to review
Check Breadcrumbs work as expected


rmccar:
[Use inline SVGs for Breadcrumbs](https://trello.com/c/uY7cj2Yj/626-use-inline-svgs-for-breadcrumbs)

boxadesign:
This fine but you can achieve this with pure css.

rmccar:
> This fine but you can achieve this with pure css.

Would doing it that way have any extra advantages? Just wondering, because I think doing it this way would be more consistent with the other components

boxadesign (approved):

PR #932, opened 2020-08-18
Changed image path for hero banner

trevorsaint:
A simple image path change. Removed the hero folder from the img directory.


rmccar:
Shouldn't we also get rid of the small and large folders?

trevorsaint:
> Shouldn't we also get rid of the small and large folders?

I need to determine the large and small images in some way. Craft creates folders for the image manipulations '_large' and '_small' etc... Which is why I specify the folders. The image name is usually the same.

PR #937, opened 2020-08-20
Remove underline from links on hover

### LINKED ISSUE #803: Don't use colour as the only indicator of hover states
We shouldn't use colour as the only indicator of hover states on links, accordions and buttons etc 

https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/233#issuecomment-651021883_

## Proposal
Remove underline of links and accordions when hovered

rmccar:
Fixes #803 
Also tides up global CSS and CSS for Collapsibles

### How to review 
Check all links remove underline on hover

### DIFF
--- a/src/components/collapsible/_collapsible.scss
+++ b/src/components/collapsible/_collapsible.scss
@@ -15,12 +15,8 @@ $collapsible-caret-right-spacing: 0.4rem;
       display: none;
     }
 
-    &__title {
-      transform: translateY(-1px);
-    }
-
     &:focus {
-      & .collapsible__heading__title {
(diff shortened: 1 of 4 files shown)


rmccar:
[Issue #803: Don't use colour as the only indicator of hover states](https://trello.com/c/mA0eWaT8/545-issue-803-dont-use-colour-as-the-only-indicator-of-hover-states)

boxadesign (approved):

PR #939, opened 2020-08-20
Use inline SVG for Feedback

rmccar:
Updates Feedback to use inline SVGs

### How to review
Check Feedback works as expected


rmccar:
[Use inline SVGs for Feedback](https://trello.com/c/frMaIXKI/633-use-inline-svgs-for-feedback)

boxadesign:
I think this was already an issue but the left alignment is off 
image

rmccar:
> I think this was already an issue but the left alignment is off
> image

Jon, said not to worry about this because the footer will be changing soon anyway

PR #941, opened 2020-08-20
Fix Inline Lists

rmccar:
Inline Lists are currently broken, they are displaying with one link per line as usual lists do. This PR will fix this

### How to review 
Check Inline Lists now display correctly


rmccar:
[Fix for Inline lists not displaying correctly](https://trello.com/c/BgDu2R5K/634-fix-for-inline-lists-not-displaying-correctly)

boxadesign (approved):

PR #943, opened 2020-08-25
Updated footer link colour. Changed OGL colour to an accessible grey

trevorsaint:
All links in the footer are to be a dark grey. Black on hover as per the design updates.

OGL colour was updated to be using an accessible grey.

**Current**
image

**Updated**
image


rmccar (approved):

jrbarnes9:
[Rationalise and update footers](https://trello.com/c/sIePEsjy/421-rationalise-and-update-footers)

PR #947, opened 2020-09-01
Update Autosuggest to remove lang code

rmccar:
This PR updates Mocha to v8.1.3 and Fuse to v6.4.1 and refactors the Autosuggest component to no longer use the language code. It will now accept data in an array format rather than a JSON object format.

### How to review 
Test that Autosuggest works as expected with new data format


rmccar:
Due to a bug in any version of fuse above 3.6.1 this is no longer possible

PR #957, opened 2020-09-03
Fix tabbing after clicking Skip to content link

### LINKED ISSUE #852: Skip to main content not working as expected.
This doesn't seem to be working as expected.

Skip to main content link is scrolling to #main-content, but not ignoring the tab order of the focusable elements before it.

So when you hit tab key afterwards, it still puts focus on the first element of the page, not the first element within #main-content

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/59#issuecomment-663470961_

rmccar:
Fixes #852 

### How to review 
Add a skip to content link and a couple of input fields with a paragraph of main content between them to a page and check the tabbing order after clicking the skip to content link. When tabbing after clicking the link it should go focus on the input below the content, not the one above.


src/components/skip-to-content/skip-to-content.js
```diff
@@ -2,6 +2,7 @@ export default function skipToMain(link) {
   const id = link.getAttribute('href').replace('#', '');
   link.addEventListener('click', event => {
     event.preventDefault();
-    document.getElementById(id).scrollIntoView();
+    document.getElementById(id).tabIndex = 0;
+    document.getElementById(id).focus();
```
boxadesign, line 6: What happens if there is no input on the page? `scrollIntoView` works on any element with the provided `id` but `focus` relies on a focusable element being the target.

boxadesign: Sorry, I see you are giving a tab order to the element to allow focus.

rmccar:
[Issue #852: Skip to main content not working as expected.](https://trello.com/c/Y5NgNiSP/580-issue-852-skip-to-main-content-not-working-as-expected)

jrbarnes9 (changes requested):
Need to exclude `page__main` from the focus styling. We don't want the border showing.

Screenshot 2020-09-09 at 10 17 38

jrbarnes9 (approved):

PR #965, opened 2020-09-07
Fix bottom error margin for Date fields

rmccar:
When Date fields are wrapped with an Error Panel the bottom margin isn't large enough for it to conform with other components

### How to review 
Check a Date component has the correct amount of space under it when wrapped in an Error Panel.

### DIFF
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -14,7 +14,7 @@
     width: 100%;
 
     &:last-child {
-      margin-bottom: 0rem;
+      margin-bottom: 0;
     }
 
     @include mq(m) {
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -14,7 +14,7 @@
     width: 100%;
 
     &:last-child {
-      margin-bottom: 0;
+      margin-bottom: 1rem;
     }
 
     @include mq(m) {
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -14,7 +14,7 @@
     width: 100%;
 
     &:last-child {
-      margin-bottom: 1rem;
+      margin-bottom: 0rem;
     }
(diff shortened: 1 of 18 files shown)


rmccar:
[Error padding](https://trello.com/c/mlhgsjud/613-error-padding)

jrbarnes9:
Seems to be inconsistent across many input types: https://deploy-preview-965--ons-design-system.netlify.app/patterns/error-validation/examples/errors-proto/errors/?day=&month=&year=

src/components/input/_macro.njk
```diff
@@ -70,7 +71,7 @@
 
         {% if params.prefix is defined or params.suffix is defined %}
             {% if params.prefix is defined %}
-                {% set prefixClass = " input-type--prefix" %}
+                {% set prefixClass = " input-type--prefix u-mb-s" %}
```
jrbarnes9, line 74: Could probably use `@extend .u-mb-s;` in the sass rather than setting and adding it to the template?

src/components/checkboxes/_checkboxes.scss
```diff
@@ -14,7 +14,7 @@
     width: 100%;
 
     &:last-child {
-      margin-bottom: 0;
+      margin-bottom: 1rem;
```
jrbarnes9, line 17: Adds too much space after answer options now (see below).  I think the panel padding itself should probably control the spacing.  Could we do something that targets the last element of any type within a panel to remove any bottom margin? We want 1rem spacing all around the error panel.  
Screenshot 2020-09-10 at 11 36 09

jrbarnes9 (changes requested):

src/components/checkboxes/_checkboxes.scss
rmccar: Ok yeah I can have a look at that. I only did it this way because I saw that all the other components had a 1rem bottom margin plus the panel padding so was just keeping it consistent. Ill see if I can do that and remove it for all in panels

jrbarnes9 (approved):

Issue #953, opened 2020-09-03
Increase `page__main` margin-top to 2.5rem (over 740) and 1.5rem (under 740).


jrbarnes9:
Increase `page__main` margin-top to 2.5rem (over 740) and 1.5rem (under 740).

Spacing is currently too tight.

<img width="886" alt="Screenshot 2020-09-03 at 08 49 09" src="https://user-images.githubusercontent.com/43346934/92086627-5bee3e80-edc2-11ea-9e2c-f30de97f920a.png">

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/195#issuecomment-686318489_

PR #993, opened 2020-09-16
Increase page__main margin

rmccar:
Fixes: #953 

This is a **BREAKING CHANGE** as it will affect the alignment of elements on some pages. Particularly pages with a related links sidebar or pages where utility classes have been used for vertical spacing.

### How to review 
Check margins above and below page width 740px


rmccar:
[Issue #953: Increase `page__main` margin-top to 2.5rem (over 740) and 1.5rem (under 740).](https://trello.com/c/mVHxdBNH/644-issue-953-increase-pagemain-margin-top-to-25rem-over-740-and-15rem-under-740)

boxadesign (dismissed):

jrbarnes9 (changes requested):
It looks like this is going to cause a problem for runner.  The problem is we don't specify where breadcrumb navigation should sit in the page template.  Runner adds it's own utility classes to control spacing, and because their 'previous' links are within page__main, they'll be pushed too far down.  We need to rethink this and breadcrumbs...

jrbarnes9 (approved):

PR #1007, opened 2020-09-21
Add census colour palette

rmccar:
This adds the Census colour palette to the Design System. It also includes changes to the onsColours macro to allow gradients to be added to colour palettes.

### How to review 
Check palette contains the correct colours and looks as expected


rmccar:
[Update Census colour palette](https://trello.com/c/Tk4JofA5/629-update-census-colour-palette)

boxadesign:
Couple issues here. The main one is the renaming of `branded` colour vars to `purple` etc. The mechanism here is to set a default `branded` colour set (which is grey) and then services can override with their own colours. The various components scss will call `element: $colour-branded` and the colour value will be provided by each service.

boxadesign (approved):

Issue #1011, opened 2020-09-22
The prefix option in lists no longer works after #892


Paul-Joel:
The prefix option in lists no longer works after #892 (released in v22.0.0)
The issue seems to be that 'prefixText' has been introduced to replace 'prefix', but 'prefix' still exists, and the code checks for 'prefix', then writes out 'prefixText'
The prefix setting is in the Census help pages, and is now broken.
It should also have been marked as a breaking change, as Craft was using 'prefix', which no longer displays the value.
There is an earlier possible issue as well that if a prefix is specified, you additionally have to manually declare 'list--bare' to stop it writing out both the automatic bullets and the prefix - this used to be added automatically

_Originally posted by @Paul-Joel in https://github.com/ONSdigital/design-system/issues/233#issuecomment-696678294_

rmccar:
Im putting a fix out for this now, to keep it backwards compatible ill remove 'prefixText' and just leave 'prefix'. As for the 'list--bare' functionality I have been through many versions of the List macro and haven't been able to find any functionally implemented that adds the 'list--bare' class automatically.

PR #1013, opened 2020-09-22
Fix List prefix issue

rmccar:
Fixes: #1011 

### How to review 
Make sure lists with prefixes work as expected


rmccar:
[Issue #1011: The prefix option in lists no longer works after #892](https://trello.com/c/TskTp1ih/671-issue-1011-the-prefix-option-in-lists-no-longer-works-after-892)

boxadesign (approved):

PR #1019, opened 2020-09-23
Add 'and var' to 'is defined'

rmccar:
On request from the runner team this PR is to standardise if statements across the Design System. Adding "and var" after "is defined" has solved some issues here and there when using in the jinja environment. As these problems were only partly solved because they were added as and when there was and issue, this PR is to add them everywhere for all components and patterns.

### How to review 
Check all components and patterns display correctly and work as expected.


rmccar:
[Update DS to have "and var" everywhere "is defined" is used](https://trello.com/c/6dU7pBIT/669-update-ds-to-have-and-var-everywhere-is-defined-is-used)

boxadesign (approved):
Checked all patterns and components and no problems observed.

boxadesign (approved):

boxadesign (approved):

Issue #333, opened 2019-04-29
Mutually exclusive getElementLabelText 


saedwards:
'getElementLabelText' method knows too much about the structure and type of nodes. For CSS, BEM is used to avoid this and should use Javascript-specific classes instead to be consistent.

bameyrick:
Relates to #29

PR #1029, opened 2020-09-24
Mutually Exclusive getElementLabelText refactor

rmccar:
Fixes: #333 
'getElementLabelText' method knows too much about the structure and type of nodes. For CSS, BEM is used to avoid this and should use Javascript-specific classes instead to be consistent.

This PR will fix this by adding a new `js-exclusive-checkbox-label` and `js-exclusive-group-item-label` classes, and then using them to select the checkboxes equivalent labels.

### How to review 
Test all mutually exclusive components work correctly


rmccar:
[mutually-exclusive.js](https://trello.com/c/FHzo1tds/238-mutually-exclusivejs)

boxadesign (approved):

Issue #1008, opened 2020-09-21
Related links not using census css for hovering on links


rmccar:
**Describe the bug**
Related links not using census css for hovering on links. Links should be black on hover not dark blue.
The census style is being set but patternlib.css is overriding it.

PR #1043, opened 2020-09-28
Fix Related Links not using Census style and tidy css

rmccar:
Fixes: #1008 
This removes the base.css being imported twice into DS examples because it has already been imported before in the main.css for the page. When census.css was being used instead it would then overwrite the census style by still importing the base.css again.
It also removes some whitespace and tidies up some variables to make the naming style consistent with others

### How to review 
Because the change in this affects all examples they all need to be checked to make sure they look and work correctly


rmccar:
[Issue #1008: Related links not using census css for hovering on links](https://trello.com/c/P0PrngrP/668-issue-1008-related-links-not-using-census-css-for-hovering-on-links)

boxadesign (approved):
Checked all pages and looks good to me!

PR #1044, opened 2020-09-28
Fix autosuggest macro options for Jinja rendering

ajmaddaford:
Checks that `options` key exists before accessing properties, otherwise Jinja fails to render when `options` is not provided.

Currently the [pull request](https://github.com/ONSdigital/eq-questionnaire-runner/pull/309) to update eq-questionnaire-runner to v25.0.0 of the design system is failing because of this.

### How to review 
Test the macro in eq-questionniare-runner with the test_textfield_suggestions schema.

### DIFF
--- a/src/components/input/_autosuggest-macro.njk
+++ b/src/components/input/_autosuggest-macro.njk
@@ -21,7 +21,7 @@
         data-error-select="{{ params.errorMessageSelect }}"
         data-error-api="{{ params.errorMessageAPI }}"
         data-error-api-link-text="{{ params.errorMessageAPILinkText }}"
-        {% if params.options %}
+        {% if params.options is defined and params.options %}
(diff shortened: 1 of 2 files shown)


src/components/input/_autosuggest-macro.njk
```diff
@@ -21,9 +21,11 @@
         data-error-select="{{ params.errorMessageSelect }}"
         data-error-api="{{ params.errorMessageAPI }}"
         data-error-api-link-text="{{ params.errorMessageAPILinkText }}"
+        {% if params.options %}
```
rmccar, line 24: I think we also need to check that this is defined too

rmccar (changes requested):

src/components/input/_autosuggest-macro.njk
ajmaddaford: Added.

rmccar (approved):

Issue #826, opened 2020-07-10
Highlight write-in when input fails validation


jrbarnes9:
Highlight write-in when input fails validation
![Screenshot 2020-07-10 at 10 18 20](https://user-images.githubusercontent.com/43346934/87138701-b8d3e980-c296-11ea-9506-7edeb5260c1b.png)

![Screenshot 2020-07-10 at 10 19 09](https://user-images.githubusercontent.com/43346934/87138758-d0ab6d80-c296-11ea-9606-f58f5b28728b.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/30#issuecomment-656577491_

armstrongb:
Is there something 'more accessible' that can be done as the highlight of the input alone may not be enough for colourblind users?

PR #1046, opened 2020-09-29
Highlight Checkbox Radio write-in on error

rmccar:
Fixes: #826 

### How to review 
Test errors prototype and check radio and checkbox textbox gets red highlighting

### DIFF
--- a/src/components/checkboxes/_checkbox.scss
+++ b/src/components/checkboxes/_checkbox.scss
@@ -86,6 +86,12 @@ $checkbox-label-vertical-padding: 0.75rem;
     box-shadow: 0 0 0 1px $color-input;
   }
 
+  .panel--error &__input:checked ~ &__other > .input--text:not(:focus) {
+    border: 1px solid $color-red;
+    outline: 1px solid $color-red;
+    outline-offset: -2px;
+  }
+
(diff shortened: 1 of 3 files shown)


rmccar:
[Issue #826: Highlight write-in when input fails validation](https://trello.com/c/J5LSmbN1/560-issue-826-highlight-write-in-when-input-fails-validation)

boxadesign:
Is this expected?

boxadesign:
Is this expected behaviour?

captured

rmccar:
> Is this expected behaviour?
> 
> captured

I figured you would only want the red highlight if that radio was currently selected. @jrbarnes9, what would you expect it to act like in that situation?

jrbarnes9:
The use case is when the user has inputted an invalid answer into the write-in.  So when the page is reloaded with the error, the input border should be red, and remain red until the format has been submitted again.

We don't change the error state on-the-fly anywhere else (i.e. when. user has subsquently selected a radio) so we should retain the error message and status here also, until the form is submitted.

While the write-in validation error message persists e.g. 'Enter a number less than 20' (because the user has entered above the max), the write-in border should remain red.

boxadesign (approved):

Issue #990, opened 2020-09-16
Collapsible chevron doesn't left align with other page elements when collapsed


jrbarnes9:
Collapsible chevron doesn't left align with other page elements when collapsed

<img width="582" alt="Screenshot 2020-09-16 at 11 59 56" src="https://user-images.githubusercontent.com/43346934/93328932-55c17e80-f814-11ea-8d4e-a9c394cf1db8.png">

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/148#issuecomment-693332811_

PR #1050, opened 2020-09-29
Align Collapsible chevron with edge of page

rmccar:
Fixes: #990 

### How to review 
Check chevron aligns with left side of page

### DIFF
--- a/src/components/collapsible/_collapsible.scss
+++ b/src/components/collapsible/_collapsible.scss
@@ -1,4 +1,4 @@
-$collapsible-caret-width: 1.1rem;
+$collapsible-caret-width: 1.5rem;
 
 .collapsible {
   &--initialised & {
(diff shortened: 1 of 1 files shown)


rmccar:
[Issue #990: Collapsible chevron doesn't left align with other page elements when collapsed](https://trello.com/c/7kLVCY6p/662-issue-990-collapsible-chevron-doesnt-left-align-with-other-page-elements-when-collapsed)

boxadesign:
Issue with active state spacing...

image

image

rmccar:
> Issue with active state spacing...
> 
> image
> 
> image

Yeah so I've spoken with @jrbarnes9 about this. The closed state needed to be aligned with the side of the page but that would mean that the chevron on the open state would creep outside of the focus highlighting and wouldn't be aligned with the side of the page. The other solution would be for the text to move to the right when opening to keep that spacing but I think that would look odd. The spacing is a bit tight though when it is open. We could make the icon smaller when open is another option? But that would be odd too.

boxadesign:
Couldn't the close state have the same spacing as bulleted lists...

image

boxadesign (approved):

PR #1059, opened 2020-10-02
Added social icons for Census2021

trevorsaint:
Added social icons for use with the icons list.

**Required for Census2021**

On another point. 
It may be best to prefix IDs with 'icon-xyz'. Just to avoid any duplicate ID usage that may exist as our icons grow.

For example I have used the following:

```
icon-facebook
icon-twitter
icon-instagram
```

image


rmccar (changes requested):
So that these can be used inline and be compatible with the list component for the social links list, they need to be added to the icons macro and not as individual files

trevorsaint:
> So that these can be used inline and be compatible with the list component for the social links list, they need to be added to the icons macro and not as individual files

Will look at that now. Thanks for the heads up.

rmccar (approved):

PR #1060, opened 2020-10-02
Changed the approach to nonce value configuration

### LINKED ISSUE #1054: Ability to provide non-Talisman csp nonce for inline script calls in the main template
Is there a way to provide a 'nonce' for the inline script calls without using Talisman? The existing if statements in the layout/_template.njk only allow for the talisman call to add a nonce as far as I can tell.

_Originally posted by @Paul-Joel in https://github.com/ONSdigital/design-system/issues/195#issuecomment-702078873_

Need to be able to add a non-Talisman nonce value to any DS driven script tags

boxadesign:
Fixes #1054


rmccar (dismissed):

Paul-Joel (approved):
LGTM

src/styles/page-template/index.njk
```diff
@@ -126,6 +126,7 @@ Variables can be set with:
 | assetsURL     | string                                              | false                             | The base url where to find CSS/JS/images etc. e.g. `http://localhost:3030`                   |
 | theme         | string                                              | false                             | The name of the css file e.g. "census" if using a theme. Defaults to "main".                 |
 | title         | string                                              | true                              | The contents for the `<title>` element                                                       |
+| cspNonce      | string                                              | false                             | Configuration for `nonce` security                                                      |
```
ajmaddaford, line 129: As the csp is abbreviated, it would be clearer if this was something like "The Content Security Policy (CSP) nonce value"

rmccar (approved):

ajmaddaford (approved):

PR #1061, opened 2020-10-02
Fix issue with webpack chunking of js files

boxadesign:
Fix issue when referencing `main.js` from cdn and receiving chunkfile load error.


boxadesign:
The issue is solved using the global `ONS_assets_base_URL` and defining the appropriate path.

PR #1069, opened 2020-10-06
Update Feedback pattern

rmccar:
To update the feedback pattern to the latest designs

### How to review 
Test that the feedback pattern looks and works how it should


rmccar:
[Update feedback form](https://trello.com/c/3zlHwsEw/396-update-feedback-form)

boxadesign:
I think this should now be a component instead of a pattern and the pattern is the form.

rmccar:
@boxadesign Yeah I almost made it a component because I thought the same. Ill make that change

boxadesign (approved):

PR #1070, opened 2020-10-06
Added flexibility to the articles for a featured item

trevorsaint:
Enhanced the articles component to allow for a featured posting.

Updated styles, and adjusted macro to cater for this change. I also added some documentation for image sizes for both standard and retina screens.

**See example**

Screenshot 2020-10-06 at 16 34 40


boxadesign:
Looks good but missing a macro options file for the component.

trevorsaint:
> Looks good but missing a macro options file for the component.

Thanks @boxadesign will take a look at that.

boxadesign (approved):

PR #1073, opened 2020-10-08
Remove redundant Breadcrumb if

rmccar:
To remove the if around breadcrumbs that hides the current page for breadcrumbs. A better solution to this is to not specify the current page in the breadcrumbs themselves.

This will be a **BREAKING CHANGE** because now current breadcrumbs will be shown on the page if specified.

### How to review 
Check breadcrumbs work as expected


rmccar:
[Remove redundant breadcrumb if](https://trello.com/c/dWzsJKbl/711-remove-redundant-breadcrumb-if)

jrbarnes9 (approved):

PR #1075, opened 2020-10-09
Fix for focus on thumbnail. Added documentation and guidance on use

trevorsaint:
- Added macro options and documentation
- Fixed a visual bug with focus style on image thumbnail

**Previously like this**
Screenshot 2020-10-09 at 11 48 34

**Now like this**
Screenshot 2020-10-09 at 11 48 43

### DIFF
--- a/src/components/downloads/_downloads.scss
+++ b/src/components/downloads/_downloads.scss
@@ -21,7 +21,9 @@
 
     &:focus {
       border-color: $color-document-border-focus;
+      box-shadow: none;
       outline: 4px solid $color-focus !important;
+      outline-offset: 0;
     }
   }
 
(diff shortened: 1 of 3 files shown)


rmccar (changes requested):
Looks good although I think It needs more documentation. If you look at other pages some sections are missing and the multiple downloads example should probably come further down the page. But I this isn't priority at the moment then just add the documentation panel back in

trevorsaint:
> Looks good although I think It needs more documentation. If you look at other pages some sections are missing and the multiple downloads example should probably come further down the page. But I this isn't priority at the moment then just add the documentation panel back in

I have put 'requires documentation' back in. That is not a priority at this stage, but the bug fix is.

rmccar (approved):

Issue #1103, opened 2020-10-19
Inline lists don’t play nicely with icons


trevorsaint:
Inline lists don’t play nicely with icons.

<img width="601" alt="Screenshot 2020-10-19 at 09 43 00" src="https://user-images.githubusercontent.com/4989027/96422167-9a637f80-11ef-11eb-9e9c-0d9a8164ed77.png">

We seem to be specifying `.svg-icon` size specifically for use within inline lists. We don’t need this. It is going to override icon scale as you can see in the example. applying a modifier in the case won’t have an effect due to specificity.

<img width="346" alt="Screenshot 2020-10-19 at 09 44 27" src="https://user-images.githubusercontent.com/4989027/96422321-c67f0080-11ef-11eb-9105-813d1b9da3fe.png">

**Removing the attribute** works just fine as the example below shows. The gap however is far too big between inline list items.
<img width="592" alt="Screenshot 2020-10-19 at 09 44 42" src="https://user-images.githubusercontent.com/4989027/96422667-38574a00-11f0-11eb-895b-2538a061a89e.png">

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/issue_comments/711868310_

rmccar:
I can put a PR out to fix the icon size issue but the gap between the inline list items is the same as all other inline lists. Is it not best to have them standardised across the service?

rmccar:
![](https://github.trello.services/images/mini-trello-icon.png) [Issue #1103: Inline lists don’t play nicely with icons](https://trello.com/c/RGYoPZO7/741-issue-1103-inline-lists-dont-play-nicely-with-icons)

trevorsaint:
> I can put a PR out to fix the icon size issue but the gap between the inline list items is the same as all other inline lists. Is it not best to have them standardised across the service?

I totally understand. We will need some flexibility in some use cases obviously, because that gap is way too big!

rmccar:
> > I can put a PR out to fix the icon size issue but the gap between the inline list items is the same as all other inline lists. Is it not best to have them standardised across the service?
> 
> I totally understand. We will need some flexibility in some use cases obviously, because that gap is way too big!

@jrbarnes9 is probably the best person to speak to about that

PR #1104, opened 2020-10-19
Fix inline list icon sizes

rmccar:
Fixes: #1103 

### How to review 
Check inline lists with icons now respect the icon size specified


rmccar:
[Issue #1103: Inline lists don’t play nicely with icons](https://trello.com/c/RGYoPZO7/741-issue-1103-inline-lists-dont-play-nicely-with-icons)

trevorsaint (approved):
Great

PR #1110, opened 2020-10-20
Remove Back Link from Feedback thank you example and add Done button

rmccar:
Remove Back Link and add done button to Feedback thank you example as it will not be on the page when implemented in runner

### How to review 
Check Feedback examples look and work as they should


boxadesign (dismissed):

rmccar:
Not planning on removing back link from the pattern at the moment

boxadesign (approved):

PR #1113, opened 2020-10-21
Update Footer params

rmccar:
This PR is to update the Footer docs as some params were missing. I have also removed the `usage` object to make it more conformant with other components and easier to document and removed the HTML param because there is no requirement for that at this time

### How to review 
Check that all params for the footer are documented


trevorsaint (approved):
Looks fine to me @rmccar

PR #1116, opened 2020-10-21
Updated visible writein tabindex

boxadesign:
DAC raised an issue with the visible write in which exists on both radio and checkbox components. Using tabbing to navigate in and out of the fieldset, the visible input would receive focus (and check the radio/checkbox). This made it impossible to not check that field for keyboard users.

This is a JS enhancement. Upon the component loading a negative `tabindex` value is given to the input. When it's associated radio/checkbox receives focus the `tabindex` is changed to `0`. Upon the field blurring, the `tabindex` is again changed to `-1`.
 
### How to review 
View the visible write in radio example. Tab into the component and select the first radio (using spacebar), then tab and you should skip out of the component (not focus inside the input which is what currently happens).

### DIFF
--- a/src/components/radios/check-radios.js
+++ b/src/components/radios/check-radios.js
@@ -4,10 +4,21 @@ export default class CheckRadios {
     this.openOther = openOther;
     this.input = this.openOther.querySelector('.input');
 
+    this.setInputBlurAttributes();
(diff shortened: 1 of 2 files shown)


rmccar:
I think we also need to add this to checkboxes

rmccar (approved):

jrbarnes9 (approved):

PR #1117, opened 2020-10-22
Remove BEM depth lint rule warning

rmccar:
This is to remove the sass lint rule for BEM depth. We have in some components the need to name elements with a BEM depth of more than 1 so these warnings are getting in the way.

### How to review 
Check BEM warnings are now gone. One component this occurs on is summary.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -4,7 +4,7 @@ options:
   formatter: stylish
   merge-default-rules: false
 rules:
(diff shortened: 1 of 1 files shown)


rmccar:
[Remove BEM lint rule warning](https://trello.com/c/tzLLlgAM/746-remove-bem-lint-rule-warning)

boxadesign (approved):

PR #1120, opened 2020-10-22
DAC - autosuggest fixes

boxadesign:

From DAC report:

- Updated aria live alerts to work as expected and at the appropriate time.
- Updated keyboard events so that it works in ie11.

Breaking changes:

New `autosuggest` properties:

- `autosuggest.ariaGroupedResults`
- `autosuggest.groupCount`

See macro option docs for examples.

### How to review 
Test the component with voiceover.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -4,7 +4,7 @@ options:
   formatter: stylish
   merge-default-rules: false
 rules:
-  bem-depth: 1
+  bem-depth: 0
   border-zero:
     - 1
     - convention: zero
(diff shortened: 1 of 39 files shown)


rmccar:
Tested the countries autosuggest example with voiceover and it doesn't seem to read out the country options

boxadesign:
> Tested the countries autosuggest example with voiceover and it doesn't seem to read out the country options

I just tested and it works for me??

rmccar (dismissed):
Works for me now

rmccar (approved):

Issue #1094, opened 2020-10-14
List items not clearing correctly


trevorsaint:
List items not clearing correctly.

Rather than using floats for the icons we should consider using flexbox. If this is not possible due to browser compatibility limitations, we will need to add a clear fix for the list items.

**See the issue below**
<img width="536" alt="Screenshot 2020-10-14 at 11 52 13" src="https://user-images.githubusercontent.com/4989027/95979753-0884fc80-0e14-11eb-97be-8d08c153f2cb.png">

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/issue_comments/708326612_

PR #1130, opened 2020-10-27
Added clearfix for icon lists

boxadesign:
Fixes https://github.com/ONSdigital/design-system/issues/1094


trevorsaint (changes requested):
Hi @boxadesign could we not add the clearfix to the `.list__item` class rather than a utility class?

We also need a vertical align on the `.list__link` even if for social icons to ensure text sits in the middle.

Social icons should be using the xl scale. In the design they are 2rem (36px), so we may need an xxl.

boxadesign:
> Hi @boxadesign could we not add the clearfix to the `.list__item` class rather than a utility class?
> 
> We also need a vertical align on the `.list__link` even if for social icons to ensure text sits in the middle.
> 
> Social icons should be using the xl scale. In the design they are 2rem (36px), so we may need an xxl.

I think i'll just introduce flex box. This was a quick fix tbh.

trevorsaint (changes requested):
Hi Matt sorry to be a pain.

You will need to apply the same flex approach for the `inline--list`. We have an inline list version of the social icons and applying this class causes alignment issues.

In your PR
Screenshot 2020-10-28 at 12 10 59

With inline-flex applied to the`list__item`.
Screenshot 2020-10-28 at 12 11 18

Very subtle agreed, but this would ensure the text is vertically aligned.

boxadesign:
> Hi Matt sorry to be a pain.
> 
> You will need to apply the same flex approach for the `inline--list`. We have an inline list version of the social icons and applying this class causes alignment issues.
> 
> In your PR
> Screenshot 2020-10-28 at 12 10 59
> 
> With inline-flex applied to the`list__item`.
> Screenshot 2020-10-28 at 12 11 18
> 
> Very subtle agreed, but this would ensure the text is vertically aligned.

No problem Trevor. Do we have that as an example in the DS anywhere?

trevorsaint:
> > Hi Matt sorry to be a pain.
> > You will need to apply the same flex approach for the `inline--list`. We have an inline list version of the social icons and applying this class causes alignment issues.
> > In your PR
> > Screenshot 2020-10-28 at 12 10 59
> > With inline-flex applied to the`list__item`.
> > Screenshot 2020-10-28 at 12 11 18
> > Very subtle agreed, but this would ensure the text is vertically aligned.
> 
> No problem Trevor. Do we have that as an example in the DS anywhere?

We don’t have an example of use in the design system,  however it will be used on the Census2021 website. 

**For example:** 
https://deploy-preview-118--sdc-prototypes.netlify.app/prototypes/census-website-engage/home/
https://deploy-preview-118--sdc-prototypes.netlify.app/prototypes/census-website-engage/ni/home/

jrbarnes9 (approved):

trevorsaint (changes requested):
Me again and sorry.

**In an inline use case:** we have margin bottom on both the li’s and the ul totalling 36px bottom margin. I don’t think this should be added by default. I would expect that to be added as utility classes if required. (in my case not).

For some reason we also have margin right on each link, so obviously the gap between each item inline is more than double what it needs to be.

**How it should render**
Screenshot 2020-10-30 at 15 11 30

trevorsaint (changes requested):
Unfortunately on small devices, it does not have sufficient spacing. See example:

Screenshot 2020-11-03 at 13 29 53

trevorsaint (approved):
Great Rich. Looks good.

Issue #1128, opened 2020-10-27
Reduce top padding on the link container that follows a summary


jrbarnes9:
Reduce top padding on the link container that follows a summary 

Should be `u-pt-s` not `u-pt-m`

From this...
![Screenshot 2020-10-27 at 08 28 58](https://user-images.githubusercontent.com/43346934/97276176-c48bf180-182e-11eb-8fea-4ade08d85093.png)

To this...
![Screenshot 2020-10-27 at 08 29 11](https://user-images.githubusercontent.com/43346934/97276203-cb1a6900-182e-11eb-95b0-89e19a6b6c2a.png)

This relates to the work done to fix this issue...
> Inconsistent borders in household collector summaries...
> <img alt="Screenshot 2020-01-22 at 16 42 49" width="712" src="https://user-images.githubusercontent.com/43346934/72914519-b3a9cf00-3d36-11ea-9dd4-ac5c76d285da.png">
> ...proposing to always display a bottom border on a summary if a link follows...
> <img alt="Screenshot 2020-01-22 at 16 47 21" width="742" src="https://user-images.githubusercontent.com/43346934/72914681-f2d82000-3d36-11ea-871f-c9328fed66fe.png">
> <img alt="Screenshot 2020-01-22 at 16 47 37" width="717" src="https://user-images.githubusercontent.com/43346934/72914682-f2d82000-3d36-11ea-99f3-245d3460b9e7.png">

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/84#issuecomment-717076430_

PR #1134, opened 2020-10-28
Reduce top padding on links after summary

rmccar:
Fixes: #1128 

### How to review 
Check padding is now correct


rmccar:
[Issue #1128: Reduce top padding on the link container that follows a summary](https://trello.com/c/BqepVICM/755-issue-1128-reduce-top-padding-on-the-link-container-that-follows-a-summary)

jrbarnes9 (commented):
If you add `margin-bottom: -1px;` to `summary` css, we can avoid the double-border when there's only row in the summary

rmccar:
> margin-bottom: -1px;

I was going to add that to the summary PR but yeah ill add it here

jrbarnes9:
> > margin-bottom: -1px;
> 
> I was going to add that to the summary PR but yeah ill add it here

Is it possible to add this only if there is only 1 in the loop?

rmccar:
> > > margin-bottom: -1px;
> > 
> > 
> > I was going to add that to the summary PR but yeah ill add it here
> 
> Is it possible to add this only if there is only 1 in the loop?

Yeah should be able to do that

rmccar:
> > > margin-bottom: -1px;
> > 
> > 
> > I was going to add that to the summary PR but yeah ill add it here
> 
> Is it possible to add this only if there is only 1 in the loop?

Did look at doing this but the border needed to be adjusted on the parent and not the item itself

boxadesign (approved):

PR #1139, opened 2020-10-30
Move Question outside of Relationships

rmccar:
This is a change to a previous breaking change to make screen readers read out the question description first to make more sense to the user. This change was difficult to implement in runner so this change is to reverse the breaking part of that PR (moving question inside of relationships) and just advise the user to make sure to use `readDescriptionFirst` on relationships wrapped in questions.

### How to review 
- Check docs and code make sense
- Examples work as they should
- Description is read out first on relationship examples

### DIFF
--- a/src/patterns/relationships/index.njk
+++ b/src/patterns/relationships/index.njk
@@ -8,6 +8,8 @@ group: Ask users for...
 
 Help users define relationships between household memebers.
 
+If this is used within a the question macro its recommended that `readDescriptionFirst` in the question params is set to true on so that the description is read first by screenreaders which will make more sense to the user.
+
 {{
     patternlibExample({"path": "components/relationships/examples/relationships/index.njk"})
 }}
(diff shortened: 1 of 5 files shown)


rmccar:
[Move question back out of relationships and update docs to advise readDescriptionFirst set to true](https://trello.com/c/fRNRGzd4/762-move-question-back-out-of-relationships-and-update-docs-to-advise-readdescriptionfirst-set-to-true)

src/patterns/relationships/index.njk
```diff
@@ -8,6 +8,8 @@ group: Ask users for...
 
 Help users define relationships between household memebers.
 
+If this is used within a the question macro its recommended that `readDescriptionFirst` in the question params is set to true on so that the description is read first by screenreaders which will make more sense to the user.
```
jrbarnes9, line 11: Move this below the example...

```suggestion
## When to use this pattern 
This pattern should be used when you need to collect the relationship between two people.

## How to use this pattern 
Call `onsRelationships` within `onsQuestion`, setting `readDescriptionFirst: true` so the description is read before the incomplete heading sentence by screen readers, to add context.

The pattern is enhanced with javascript, to complete the sentence in the heading and `playback` when the user selects a radio option.
```

jrbarnes9 (approved):

PR #1144, opened 2020-11-02
Fix extra spaces added around titles when escaping HTML

rmccar:
To fix a failing runner test because when testing escaping HTML for the address pattern it was adding spaces either side of the question title.

Also fixes some indentation in the address macro

### How to review 
Check question example and see that in the HTML the whitespace around the question title text is now gone

### DIFF
--- a/src/components/address/_macro.njk
+++ b/src/components/address/_macro.njk
@@ -18,67 +18,67 @@
                     }}
                 {% endif %}
                 {% if params.line1 is defined and params.line1 %}
-                {{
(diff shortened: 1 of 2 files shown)


rmccar:
[Fix to remove whitespace around question titles for runner tests](https://trello.com/c/dHgwmsMJ/766-fix-to-remove-whitespace-around-question-titles-for-runner-tests)

jrbarnes9 (approved):

Issue #1129, opened 2020-10-27
Remove top and bottom padding from `panel--bare` so spacing is dictated by bottom margins


jrbarnes9:
Remove top and bottom padding from `panel--bare` so spacing is dictated by bottom margins

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/49#issuecomment-717178082_

PR #1145, opened 2020-11-03
Panel spacing improvements and renaming `simple` class to `no-title`

rmccar:
Fixes: #1129 

Also renames the simple class to no-title to make it more descriptive

### How to review 
- Test all types of panels to see that they behave correctly
- Check new bare example is correct


rmccar:
[Issue #1129: Remove top and bottom padding from `panel--bare` so spacing is always dictated by bottom margins](https://trello.com/c/EpU9Wycd/757-issue-1129-remove-top-and-bottom-padding-from-panel-bare-so-spacing-is-always-dictated-by-bottom-margins)

boxadesign (approved):

PR #1147, opened 2020-11-03
Update Radio select docs

rmccar:
This is to add a line to the Radio select example that explains to the user they need to select the "pepperoni" option for the select to become visible.

### How to review 
Check docs make sense


rmccar:
[Add more docs for select radio](https://trello.com/c/XdQlmpaY/767-add-more-docs-for-select-radio)

boxadesign (approved):

PR #1154, opened 2020-11-04
Add ability to set target on copyright link

rmccar:
Add ability to set target on copyright link so that it can be set to open in a new tab

### How to review 
Check that target gets set on the link correctly


rmccar:
[Add functionality for target attribute to be set on the copyright declaration](https://trello.com/c/1nk4XbRh/773-add-functionality-for-target-attribute-to-be-set-on-the-copyright-declaration)

src/components/footer/_macro.njk
```diff
@@ -221,7 +221,7 @@
                 {% if params.copyrightDeclaration is defined and params.copyrightDeclaration %}
                 <div class="grid">
                     <div class="grid__col u-mt-m">
-                        <p class="u-fs-s u-mb-no">&copy; {{ params.copyrightDeclaration.copyright }} <br> {{ params.copyrightDeclaration.text }} <a href="{{ params.copyrightDeclaration.url }}">{{ params.copyrightDeclaration.link }}</a>.</p>
+                        <p class="u-fs-s u-mb-no">&copy; {{ params.copyrightDeclaration.copyright }} <br> {{ params.copyrightDeclaration.text }} <a href="{{ params.copyrightDeclaration.url }}" target="{{ params.copyrightDeclaration.target }}">{{ params.copyrightDeclaration.link }}</a>.</p>
```
jrbarnes9, line 224: Does this need an `if` statement in case target is not defined?

jrbarnes9: Will need `rel="noopener"` if `target` is defined, so could bundle both attributes with `params.newWindow` like we do in button component...
```
if params.newWindow is defined and params.newWindow %}target="_blank" rel="noopener"{% endif %}
```

rmccar: yep ill sort that out

boxadesign (approved):

PR #1157, opened 2020-11-06
Add table headers to Summaries

rmccar:
This is a issue from the DAC report to make summaries more accessible by adding visually hidden table headers

### How to review 
See that table headers have been added and work well with screen readers


rmccar:
[HIGH: Summary tables - hard to understand without table headers (GDS use <dl>) (pg.14 & 16) >> change summaries to definition lists](https://trello.com/c/ipTkykTl/715-high-summary-tables-hard-to-understand-without-table-headers-gds-use-dl-pg14-16-change-summaries-to-definition-lists)

src/components/summary/examples/hub/index.njk
```diff
@@ -6,6 +6,7 @@
             {
                 "groups": [
                     {
+                    "headers":["Name of section or person", "Section progress", "Access section"],
```
boxadesign, line 9: Can runner provide just values and not keys? Maybe something to check.

rmccar: Should be able to I would think but ill check with the runner guys

rmccar: Spoken to the runner guys and they're happy with this implementation

boxadesign (approved):

PR #1158, opened 2020-11-09
Fix classes param for checkbox selects

rmccar:
Bug in checkbox selects where classes aren't being passed because of typo

### How to review 
Check that checkbox selects work correctly with classes being passed correctly


rmccar:
[Checkbox select class bug](https://trello.com/c/mQKL1S5N/777-checkbox-select-class-bug)

boxadesign (approved):

PR #1160, opened 2020-11-09
Add link to summary

rmccar:
Add an 'add' link to summaries and allows setting of a placeholder if there are no rows in the summary

This PR contains a **BREAKING CHANGE** for summaries. This is because links that come after summaries such as add links will no longer be defined outside of the summary.

### How to review 
- Check that add link works as the previous external add link does
- Check placeholder works
- Check that the correct border is or is not added with and without a link and if there is 0, 1 or >1 number of rows


rmccar:
[Add an 'add' link to summaries](https://trello.com/c/soXwE2B6/775-add-an-add-link-to-summaries)

src/components/summary/examples/summary-household/index.njk
```diff
@@ -62,13 +62,14 @@
...
+                        "url": "#0",
+                        "dataQa": "add-item-link"
+                    }
                 }
             ]
         }
     ]
 }) }}
-
-<div class="u-mb-l u-pt-s u-bt">
```
jrbarnes9, line 72: I think this div should have a class, then apply top padding in the css like we do for the summary rows.  Then the service can use utility classes on `summary` to add the desired bottom margin to the component

rmccar: I don't think it needs a class to do that now its been integrated into the component but I have moved the mb to the `summary` and set it in the example which I think will achieve the same result

src/components/summary/_macro.njk
```diff
@@ -120,7 +120,7 @@
                     {% endif %}
 
                     {% if group.summaryLink is defined and group.summaryLink %}
-                        <div class="u-mb-l u-pt-s{% if group.placeholderText is not defined and group.rows is defined and group.rows %} u-bt{% endif %}">
+                        <div class="u-pt-s{% if group.placeholderText is not defined and group.rows is defined and group.rows %} u-bt{% endif %}">
```
jrbarnes9, line 123: I think this padding and border shouldn't use util classes, and instead should be css, so I would add a class for this link (we may want to style it in future) now that we know it's specific use e.g. `summary__add-link`

src/components/summary/_summary.scss
```diff
@@ -3,6 +3,7 @@ $summary-col-spacing: 1rem;
 $hub-row-spacing: 1.3rem;
 
 .summary {
+  @extend .u-mb-l;
```
jrbarnes9: Sorry, I meant we should leave the summary bottom margin down to the user to add as they wish with a utility class, as we do with most elements.  They need to control spacing between components.

rmccar: Yeah I realised that and have fixed it

src/components/summary/examples/summary-household/index.njk
jrbarnes9: Don't we usually try to make classes for component elements?  Just makes it more future proof, allowing us to make styling changes in the future.

rmccar: Yeah makes sense

src/components/summary/_macro.njk
```diff
@@ -20,100 +20,110 @@
...
+                                    {% endfor %}
+                                </tbody>
+                            {% endfor %}
+                        </table>
+                    {% elif group.placeholderText is defined and group.placeholderText %}
+                        {{ group.placeholderText }}
+                    {% endif %}
+
+                    {% if group.summaryLink is defined and group.summaryLink %}
+                        <div class="summary__add-link{% if group.placeholderText is not defined and group.rows is defined and group.rows %} u-bt{% endif %}">
```
jrbarnes9, line 123: rather than the negative bottom margin we put on `.summary__items` to avoid the double border, could we just put something like if group.rows > 1 for that border class, so it’s not added if there’s only one row in the summary which has it’s own border?

jrbarnes9 (approved):

Issue #1127, opened 2020-10-27
Footer should use a fluid layout, allowing links to wrap in a row


jrbarnes9:
Footer should use a fluid layout, allowing links to wrap in a row

From this...
![Screenshot 2020-10-27 at 08 12 31](https://user-images.githubusercontent.com/43346934/97274706-ee441900-182c-11eb-97d9-f7f7ce3d8ec0.png)

To this...
![Screenshot 2020-10-27 at 08 18 39](https://user-images.githubusercontent.com/43346934/97274784-09af2400-182d-11eb-8a37-dc482a414d4c.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/172#issuecomment-717070182_

trevorsaint:
If we remove the `flex: 0 0 auto` from the `grid__col`, our list items flow as you would expect them too.

If we can’t remove this property globally, we could have a modifier of sorts specific to footer list items.

Some adjustments need to be made with spacing and padding etc. But this will allow our items to flow naturally.

**Resulting in:**

<img width="440" alt="Screenshot 2020-10-27 at 11 21 53" src="https://user-images.githubusercontent.com/4989027/97295303-04f76980-1847-11eb-8222-a79712d7bd85.png">

PR #1161, opened 2020-11-09
Modified footer to allow for responsive links. Modified row usage to …

trevorsaint:
closes: https://github.com/ONSdigital/design-system/issues/1127

- Simplified macro
- Split row usage and added `legal` to simplify use
- List links are now fluid and don’t require breakpoint classes
- Documentation added

This will potentially be a breaking change. Instead of using two rows, we will need to add legal to include things like:

- Cookies
- Accessibility statement
- Privacy and data protection
- Terms and conditions

We can still add many rows into the footer but this spits our legal links out, which allows more layout control.

**Before:**
Screenshot 2020-11-09 at 16 47 11

**After:**
Screenshot 2020-11-09 at 16 46 47


rmccar:
[Issue #1127: Footer should use a fluid layout, allowing links to wrap in a row](https://trello.com/c/k6JGrwhf/756-issue-1127-footer-should-use-a-fluid-layout-allowing-links-to-wrap-in-a-row)

src/scss/utilities/_grid.scss
```diff
@@ -121,7 +125,7 @@
 .grid__col {
   display: inline-block;
   box-sizing: border-box;
-  flex: 0 0 auto;
+  // flex: 0 0 auto;
```
rmccar, line 128: Probably just want to remove this rather than comment it out

rmccar:
image

Theres an issue with the footer warning

src/scss/utilities/_grid.scss
trevorsaint: Agreed.

rmccar (approved):

PR #1166, opened 2020-11-11
Change single breadcrumb breakpoint

rmccar:
Breadcrumbs switching to single happens after language header removal and sign out button moving to footer. This will fix this issue so all things change at the same point.

Also fixes some indentation

### How to review 
Check the customised page template example and make the viewport smaller to see changes

### DIFF
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -36,7 +36,7 @@ $breadcrumb-chevron-height: 0.65rem;
       }
     }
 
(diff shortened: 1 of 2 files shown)


rmccar:
[Breadcrumbs switching to single happens after language header removal and sign out button moving to footer](https://trello.com/c/I43LdFYe/779-breadcrumbs-switching-to-single-happens-after-language-header-removal-and-sign-out-button-moving-to-footer)

rmccar:
Decided this was working as expected as it is

PR #1176, opened 2020-11-19
Allow setting of grid col number

rmccar:
Allow setting of grid col number in the pageConfig. This change was due to it needing to be able to be overwritten for the census website

### How to review 
Check `pageColNumber` can be set and it defaults to 8 if not set

### DIFF
--- a/src/styles/page-template/_template.njk
+++ b/src/styles/page-template/_template.njk
@@ -11,7 +11,7 @@
     {% set otherLanguageISOCode = otherLanguage.ISOCode %}
 {% endif %}
 
-{% set pageColNumber = "8" %}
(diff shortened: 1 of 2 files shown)


rmccar:
[Add ability to set col widths in page template](https://trello.com/c/6PYnKmDw/789-add-ability-to-set-col-widths-in-page-template)

boxadesign:
I don't think this should be set at this level, as in defined in `pageConfig`. This would mean every template that extends this would have it set.

I think we need to create a variable and set it in the page_template and then this can be overriden within templates that extend it. This would need to be tested as there may be an inheritance issue at the point the value gets set.

boxadesign (approved):

Issue #1170, opened 2020-11-12
Selected DS menu item should be in focus after page reload


rmccar:
Selected DS menu item should be in focus after page reload. This would improve the experience for the user as if the user was clicking through the list of components in the menu they would no longer have to keep scrolling down to get back to where they were. It would also help if they had mis-clicked and clicked the component next to the one they wanted.

PR #1186, opened 2020-11-23
Sub navigation scroll current page into view

rmccar:
Fixes: #1170 

### How to review 
See that when selecting a menu item that normally needs to be scrolled to


rmccar:
[Issue #1170: Selected DS menu item should be in focus after page reload](https://trello.com/c/DMDS4UCi/783-issue-1170-selected-ds-menu-item-should-be-in-focus-after-page-reload)

boxadesign (approved):

PR #1188, opened 2020-11-25
Fix issue with incorrect index value for prev/next url generation

boxadesign:
Fix issue with incorrect index value for prev/next url generation

### How to review 
Check prev/next url's are correct

### DIFF
--- a/src/components/pagination/examples/pagination-no-range-indicator/index.njk
+++ b/src/components/pagination/examples/pagination-no-range-indicator/index.njk
@@ -7,32 +7,32 @@
         "hideRangeIndicator": true,
         "pages": [
             {
-                "url": "#"
+                "url": "#page1"
             },
             {
-                "url": "#",
+                "url": "#page2"
+            },
+            {
+                "url": "#page3"
+            },
+            {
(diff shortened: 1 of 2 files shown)


src/components/pagination/examples/pagination-no-range-indicator/index.njk
```diff
@@ -7,32 +7,32 @@
...
             {
-                "url": "#"
+                "url": "#page8"
             },
             {
-                "url": "#"
+                "url": "#page9"
             },
             {
                 "url": "#"
```
rmccar, line 38: Do we not want to set the page number urls for all the nodes? Other than that looks good

boxadesign: Yeah we can do. It was more to test that next/prev links rendered the correct url. Obviously the url's don't go anywhere.

rmccar: Yeah just makes sense to me if you're going to do it for some you may as well do it for all of them

rmccar (approved):

Issue #1194, opened 2020-11-27
Remove unnecessary padding from `panel--warn`


jrbarnes9:
Remove unnecessary padding from `panel--warn`

<img width="975" alt="Screenshot 2020-11-27 at 08 24 55" src="https://user-images.githubusercontent.com/43346934/100427492-70bd3400-308a-11eb-8791-091756e6f518.png">

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/49#issuecomment-734711044_

PR #1196, opened 2020-11-27
Remove padding on warning panel

rmccar:
Fixes: #1194 

This is a **BREAKING CHANGE** for UAC. The utility classes `u-mb-l` and `u-mt-l` will now need to be passed into the panel in the collapsible for that pattern like the change in this PR for the UAC example.

### How to review 
- Check padding has been removed
- Check everything that uses the panel still looks correct


rmccar:
[Issue #1194: Remove unnecessary padding from `panel--warn`](https://trello.com/c/gXgZqwfo/799-issue-1194-remove-unnecessary-padding-from-panel-warn)

boxadesign (approved):

Issue #1173, opened 2020-11-17
Add the ability to also set suffixIcon as well as `prefixIcon`


jrbarnes9:
Add the ability to also set suffixIcon as well as `prefixIcon`

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/233#issuecomment-728848912_

PR #1197, opened 2020-11-27
Add suffix option to lists and fix link icon colour hover

rmccar:
Fixes: #1173 
- Updates case of "forum" in external links to be more consistent
- Adds suffix option to list items.
- Updated css so that icons in icon links now change colour on hover

### How to review 
- Check suffix option works for lists
- Check icons in icon links change to correct colour on hover


rmccar:
[Issue #1173: Add the ability to also set `suffixIcon` as well as `prefixIcon`](https://trello.com/c/7r2PIrqB/785-issue-1173-add-the-ability-to-also-set-suffixicon-as-well-as-prefixicon)

jrbarnes9 (commented):
I think a better example for the prefixed list would be the 'H1, H2' etc that we use for the census paper questions - as it's something you can't achieve with `list-style-type`.

I can't think of a good example for the suffixed list.

src/styles/typography/index.njk
```diff
@@ -141,6 +141,13 @@ The prefix list allows each list item to contain a string as a prefix to the ite
     patternlibExample({"path": "styles/typography/lists/examples/list-prefix/index.njk", "componentFolderPath": "components/lists"})
 }}
 
+#### Suffixed list
+The suffix list works the same as the prefix but just comes after the list item and the full stop isnt included.
```
jrbarnes9, line 145:
```suggestion
The suffix list works the same as the prefix but just comes after the list item and the full stop isn’t included.
```

jrbarnes9 (approved):

jrbarnes9 (approved):

Issue #1177, opened 2020-11-19
Ensure input width class numerals match number of widest character


jrbarnes9:
Ensure input width class numerals match number of widest character

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/163#issuecomment-730369568_

PR #1205, opened 2020-12-04
Add number width class to input

rmccar:
Fixes: #1177 

### How to review 
Check text field width using both `input--w-<number>` and `input-number--w-<number>` are correct width by entering the same number of characters into the field. Using '4's for the number one and 'W's for the character one.

### DIFF
--- a/src/components/input/_input.scss
+++ b/src/components/input/_input.scss
@@ -57,6 +57,9 @@
 // Text input widths
 @include input-width('input--w-{x}');
 
+// Number input widths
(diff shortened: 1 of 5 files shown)


rmccar:
[Issue #1177: Ensure input width class numerals match number of widest character](https://trello.com/c/tpPe2265/790-issue-1177-ensure-input-width-class-numerals-match-number-of-widest-character)

boxadesign (approved):

PR #1206, opened 2020-12-07
Fix button icon colour

rmccar:
This PR is to fix the problem with icons when added to buttons always being black.
image (1)

It also reduces the whitespace around button text in the HTML and fixes an error in the button param docs

### How to review 
Add icons to regular and secondary buttons and make sure they are the correct colour


rmccar:
[Fix button icon colour](https://trello.com/c/a9nvywZJ/806-fix-button-icon-colour)

boxadesign (approved):

PR #1207, opened 2020-12-07
Fix footer BEM and footer warning spacing

rmccar:
This is to update the footer css to conform with BEM rules and to update the footer warning panel to have the correct spacing between the text and icon.

It also removes some redundant css and makes sure all utility classes are set last.

### How to review 
- Check code makes sense
- Check footer warning panel spacing matches warning panel spacing
- Check all footers still look as expected


rmccar:
[Fix extra footer border and BEM](https://trello.com/c/C02ZUr3W/807-fix-extra-footer-border-and-bem)

boxadesign (approved):

PR #1211, opened 2020-12-07
Fix footer on census page template example

rmccar:
The census page template example wasn't configuring the footer correctly which was adding an extra border. This PR fixes that.

### How to review 
Check footer looks as expected in census page template example


rmccar:
[Fix extra border in footer in census page example](https://trello.com/c/9eUIlYmH/808-fix-extra-border-in-footer-in-census-page-example)

boxadesign (approved):

PR #1214, opened 2020-12-07
Download updates

trevorsaint:
Removed filename to simplify component for Craft usage.

We have smallSrc and largeSrc. There is no need to specify the filename separately.


src/components/downloads/_macro.njk
```diff
@@ -4,10 +4,10 @@
 
         <div class="download__image" aria-hidden="true">
             <a class="download__image-link" href="{{ download.url }}" tabindex="-1">
-                {% if download.thumbnail is defined and download.thumbnail and download.thumbnail.filename %}
-                    <img srcset="{{ download.thumbnail.smallSrc }}{{ download.thumbnail.filename }} 1x, {{ download.thumbnail.largeSrc }}{{ download.thumbnail.filename }} 2x" src="{{ download.thumbnail.smallSrc }}{{ download.thumbnail.filename }}" alt="{{ download.thumbnail.alt }}">
+                {% if download.thumbnail is defined and download.thumbnail and download.thumbnail.smallSrc %}
```
rmccar, line 7:
```suggestion
                {% if download.thumbnail is defined and download.thumbnail and download.thumbnail.smallSrc is defined and download.thumbnail.smallSrc and download.thumbnail.largeSrc is defined and download.thumbnail.largeSrc %}
```

Id add this to be safe, otherwise looks good

rmccar (approved):

PR #1219, opened 2020-12-08
Panel variant for census 2021 respond phase

trevorsaint:
As part of the respond phase for census 2021, we use a modified version of the warning panel.

As such I have added a new variant.

It can be seen in use here: https://census-website-respond.netlify.app/end-to-end/home/

This can be contained in a fixed with container and the background extends beyond its container. 
I needed to add overflow-y: hidden to the body class for this to work correctly.

Screenshot 2020-12-08 at 14 47 04

### DIFF
--- a/src/components/panel/_panel.scss
+++ b/src/components/panel/_panel.scss
@@ -17,7 +17,7 @@
   position: relative;
   border-radius: 0;
 
-  /* Removes inherited bottom margin to make whitespace inside panel equal */
+  // Removes inherited bottom margin to make whitespace inside panel equal
   > *:last-child {
     margin-bottom: 0;
   }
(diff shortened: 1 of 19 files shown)


src/components/panel/_panel.scss
```diff
@@ -80,6 +80,24 @@
     }
   }
 
+  &--warn-ondark {
```
rmccar, line 83: Can this not be styled in the same way the footer warning is? I think we will be able to reuse more code if it can

trevorsaint: Let me take a look at how that is is being used Rich.

src/components/panel/examples/warn-branded/index.njk
```diff
@@ -0,0 +1,14 @@
+---
+"theme": census
```
rmccar, line 2:
```suggestion
"theme": census
"fullWidth": true
```
This will make the example fill the width of the page which I think will make it look more how you want it to

rmccar (approved):

Issue #1230, opened 2020-12-10
Alt tags should not be on source tags only img tags


trevorsaint:
Looking at the source code.

Alt tags should NOT be on the source tag. Only the img tag.

<img width="1093" alt="Screenshot 2020-12-10 at 16 26 51" src="https://user-images.githubusercontent.com/4989027/101799873-8a3b9280-3b04-11eb-93e7-57f6076cdaab.png">

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/discussions/160#discussioncomment-187228_

PR #1241, opened 2020-12-14
Remove alt attribute from source tag

rmccar:
Fixes: #1230 

### How to review 
Check header still looks and works as it should


rmccar:
[Issue #1230: Alt tags should not be on source tags only img tags](https://trello.com/c/4wmO5awn/819-issue-1230-alt-tags-should-not-be-on-source-tags-only-img-tags)

trevorsaint (approved):
Looks good to me.

PR #1248, opened 2020-12-16
Fixed header by adding display block on the logo svg that removes ext…

trevorsaint:
Closes https://github.com/ONSdigital/design-system/issues/1245

### DIFF
--- a/src/components/header/_header.scss
+++ b/src/components/header/_header.scss
@@ -67,6 +67,7 @@
     color: $color-white;
 
     &-link {
+      display: block;
       &:focus {
         outline: 4px solid $color-focus;
       }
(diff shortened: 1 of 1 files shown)


rmccar (changes requested):
This was removed because it no longer allows/shows tabbing focus onto the census logo when it has a link added to it. If this needs to be realigned we need to do it in another way.

trevorsaint:
> This was removed because it no longer allows/shows tabbing focus onto the census logo when it has a link added to it. If this needs to be realigned we need to do it in another way.

Interesting. Let me look into this.

rmccar (approved):

PR #1252, opened 2020-12-17
Fix dsExample logic double negative for focus on error panels

rmccar:
Error panel inputs are not focusing when clicking the link in the error panel header this fixes the double negative logic to stop this happening

### How to review 
Check that in the error prototype the error panels now focus on the inputs within them when the header links are clicked


rmccar:
[Error panels not focusing on input when header links clicked](https://trello.com/c/e65DOV9M/830-error-panels-not-focusing-on-input-when-header-links-clicked)

boxadesign (approved):

Issue #1231, opened 2020-12-10
Empty ID’s break W3C validation


trevorsaint:
I am not sure why we have ID’s on the breadcrumbs. But having empty IDs on the links will cause the page to break W3C validation.

<img width="816" alt="Screenshot 2020-12-10 at 16 32 24" src="https://user-images.githubusercontent.com/4989027/101800686-6fb5e900-3b05-11eb-97de-969f51ac6e08.png">

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/discussions/43#discussioncomment-187255_

PR #1253, opened 2020-12-17
Remove unneeded id from breadcrumbs

rmccar:
Fixes: #1231

This is a **BREAKING CHANGE** for Breadcrumbs in Runner because ids will no longer be able to be used in tests

### How to review 
Check breadcrumbs still work as expected


rmccar:
[Issue #1231: Empty ID’s break W3C validation](https://trello.com/c/FUZZkZH0/818-issue-1231-empty-ids-break-w3c-validation)

boxadesign:
You could put an `if` around the `id` to prevent this being breaking or you would need to provide a `data-` (inside an `if`) so runner can still test the link.

rmccar:
> You could put an `if` around the `id` to prevent this being breaking or you would need to provide a `data-` (inside an `if`) so runner can still test the link.

Yeah that could be a good idea, I was just thinking as you said that they should really be using data attributes and this would mean they would have to use them. But thinking about it now, maybe a non breaking change would be good. So ill put the `if` round it

boxadesign (approved):

PR #1257, opened 2020-12-18
Remove striptags due to runner vulnerability

rmccar:
`striptags` causes a vulnerability in runner as they are already escaping tags but us calling `striptags` on it as well unescapes it. I don't think `striptags` is needed anyway because html strings shouldn't be provided in the aria label

### How to review 
Check aira-labels on summary actions work correctly without `striptags`

### DIFF
--- a/src/components/summary/_macro.njk
+++ b/src/components/summary/_macro.njk
@@ -104,7 +104,7 @@
                                                         <a
                                                             href="{{ action.url }}"
                                                             class="summary__button"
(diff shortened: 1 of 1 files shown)


rmccar:
[Remove `striptags` for Runner vulnerability](https://trello.com/c/saxVwiT9/831-remove-striptags-for-runner-vulnerability)

boxadesign (approved):

Issue #1266, opened 2020-12-30
The width of the census logo should be the same regardless of mobile menu


jrbarnes9:
The width of the census logo should be the same regardless of mobile menu
![Screenshot 2020-12-30 at 11 24 47](https://user-images.githubusercontent.com/43346934/103348694-b0fd4100-4a92-11eb-86d0-82bf7f54abe6.png)
![Screenshot 2020-12-30 at 11 28 51](https://user-images.githubusercontent.com/43346934/103348700-b35f9b00-4a92-11eb-8207-d13896d096dd.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/160#discussioncomment-252451_

PR #1267, opened 2021-01-04
Header logo tweaks

rmccar:
Fixes: #1266 

### How to review 
- Check that when the window is at the xs breakpont (300px) the census header logo is the same size with or without the menu button
- Check that the stacked ONS header logo is the same size for both Welsh and English


rmccar:
[The width of the census logo should be the same regardless of presence of mobile menu](https://trello.com/c/hdsY39hr/838-the-width-of-the-census-logo-should-be-the-same-regardless-of-presence-of-mobile-menu)

trevorsaint (commented):
I am wondering if using 227.83 is going to cause rendering bugs. Perhaps even mess with the sharpness of the images. Perhaps it is best to use whole values like 228, even if some white space exists in the SVG. Just an idea.

boxadesign (approved):

Issue #1260, opened 2020-12-22
The vertical position of the census logo, and the height of the header block are different between a default header, and header with sign out button


jrbarnes9:
The vertical position of the census logo, and the height of the header block are different between a default header, and header with sign out button

https://user-images.githubusercontent.com/43346934/102880668-99dfa300-4443-11eb-9e25-81d5c434b760.mov

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/160#discussioncomment-234141_

PR #1272, opened 2021-01-06
Fix header height without signout button

rmccar:
Fixes: #1260 

### How to review 
Check that the header is the same height with and without signout button


rmccar:
[Issue #1260: The vertical position of the census logo, and the height of the header block are different between default header, and header with sign out button](https://trello.com/c/gM2qbhau/834-issue-1260-the-vertical-position-of-the-census-logo-and-the-height-of-the-header-block-are-different-between-default-header-and)

boxadesign (approved):

PR #1273, opened 2021-01-06
Fixed bug with joined class names. Modified whtespace between classes

trevorsaint:
Fixed bug with how social icons are rendered due to no space between class names

**Bug**
Screenshot 2021-01-06 at 15 15 56

**Fixed**
Screenshot 2021-01-06 at 15 16 07

Altered the macro to ensure correct spaces are used for each type. See output for each list type used.

### DIFF
--- a/src/components/lists/_macro.njk
+++ b/src/components/lists/_macro.njk
@@ -17,7 +17,7 @@
         {% set listEl = 'ul' %}
     {% endif %}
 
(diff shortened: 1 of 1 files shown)


boxadesign (approved):

jrbarnes9:
[Social icons not rendering correctly due to no space between classes](https://trello.com/c/b96Brecn/1000-social-icons-not-rendering-correctly-due-to-no-space-between-classes)

Issue #1280, opened 2021-01-08
Error examples displaying suffix number


jrbarnes9:
<img width="811" alt="Screenshot 2021-01-08 at 15 53 56" src="https://user-images.githubusercontent.com/43346934/104035343-ee1bae80-51c9-11eb-9ab9-13bb27dd2c4e.png">

PR #1282, opened 2021-01-11
Remove list index from suffix

rmccar:
Fixes: #1280

### How to review 
Check list indexes are no longer displayed using prefix and suffix and are just on the prefix now


rmccar:
[Issue #1280: Error examples displaying suffix number](https://trello.com/c/Gt3IspVI/857-issue-1280-error-examples-displaying-suffix-number)

jrbarnes9:
No error summary panels should be using `list__prefix`.  Found in Access code pattern error example

boxadesign (dismissed):

rmccar:
> No error summary panels should be using `list__prefix`. Found in Access code pattern error example

I don't think the issue is that they're using `list__prefix` because all that does now is set the font to monospaced. But I have removed `list--bare` from UAC and Feedback examples which will fix the spacing issue

boxadesign (approved):

PR #1285, opened 2021-01-11
Rework form submission for timer/loader button

boxadesign:
An issue was raised in the RAS team that when multiple timer buttons existed on the page that the disabling didn't work. This was due to the event being coupled to the form submission, which lost the scope of the button that fired the event.


rmccar:
The loader/timer button DS examples don't seem to work after this change

boxadesign:
> The loader/timer button DS examples don't seem to work after this change

How are you testing? What are you expecting to happen?

rmccar:
> > The loader/timer button DS examples don't seem to work after this change
> 
> How are you testing? What are you expecting to happen?

So if you go to the buttons page and scroll down to the loader and timer buttons. Click them in live and they behave how they would on a page, stopping you clicking one for a few seconds or activating the spinner. On this branch they look like they just refresh the example

rmccar (dismissed):

rmccar (approved):

jrbarnes9:
https://trello.com/c/CELPaoK0/848-fix-timer-loader-button

PR #1286, opened 2021-01-11
Added a consistent margin bottom for each list item to improve legibility

trevorsaint:
List item margin bottom has been added to improve legibility.

Myself and Jon felt a value of 9px Or 0.5rem was a value that would look good in all cases with the exception of the spacious modifier.

See all list variations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "webpack": "^4.26.0",
     "webpack-bundle-analyzer": "^3.0.3",
     "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "^3.11.0",
+    "webpack-dev-server": "^3.11.1",
     "webpack-fix-style-only-entries": "^0.0.5",
     "webpack-livereload-plugin": "^2.1.1",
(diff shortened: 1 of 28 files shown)


rmccar:
Did you test increasing the margin bottom on the li tag in _typeography.scss instead? I know that could affect a lot of other stuff but as we want this change on all list items it could be a better solution if it doesn't affect anything negatively

trevorsaint:
> Did you test increasing the margin bottom on the li tag in _typeography.scss instead? I know that could affect a lot of other stuff but as we want this change on all list items it could be a better solution if it doesn't affect anything negatively

Yes I can see how that looks @rmccar. We do want this globally applied for all list items.

rmccar (approved):

jrbarnes9:
[Usability issue reading bare lists when items wrap due to spacing and line height](https://trello.com/c/MVAYjiwX/1002-usability-issue-reading-bare-lists-when-items-wrap-due-to-spacing-and-line-height)

PR #1294, opened 2021-01-13
Download component - Correctly place comma if required and updates to styling when used inside page__body

### LINKED ISSUE #1291: Incorrect margins for download component heading
I have spotted a bug with this component when used inside `page__body`. The margin is inherited causing incorrect margins.

**See example:**

Screenshot 2021-01-12 at 17 32 06

The top image is how it should look. The bottom image is how it currently looks.

It might be worth adding a title class for the heading to control how it should be rendered.

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/discussions/716#discussioncomment-277191_

trevorsaint:
Updated download component to reset the heading margin top due to being used inside `page__body`.

This fixes #1291

**Enhancements added**

If some meta tags don’t exist, we see comma’s being used incorrectly. This fixes that problem. I have also added this pattern to the visually hidden text in the heading.

I have tested in various scenarios and it works fine. I have also added some consistency tweaks to the examples.

### DIFF
--- a/src/components/downloads/_macro.njk
+++ b/src/components/downloads/_macro.njk
@@ -1,5 +1,7 @@
 {% macro onsDownloads(params) %}
+
     {% for download in (params.downloads if params.downloads is iterable else params.downloads.items()) %}
+
     <div class="download{{ ' ' + params.classes if params.classes else '' }}">
 
         <div class="download__image" aria-hidden="true">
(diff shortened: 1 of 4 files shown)


rmccar (approved):

rmccar:
[Issue #1291: Incorrect margins for download component heading](https://trello.com/c/m1ry6n6X/870-issue-1291-incorrect-margins-for-download-component-heading)

Issue #1278, opened 2021-01-08
Set page contents to be displayed right-to-left using css `direction: rtl`


jrbarnes9:
Set page contents to be displayed right-to-left using css `direction: rtl`

Originally required for the Craft Census website 'guide' template to be displayed rtl for rtl language guidance

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/195#discussioncomment-269386_

trevorsaint:
This is a little more complicated that simply adding a CSS attribute `direction: rtl`.

We are using push classes on the grid, which obviously won’t work if you reverse the items. It would need to be a pull.

Perhaps we can have a content layout which creates the layout without the push and pull classes that does exactly the same thing, but a simple modifier class would do what we need. For example:

```
<div class="content-layout content-layout--rtl">
  <div class="content-layout__main"></div>
  <div class="content-layout__related"></div>
</div>
```

We won’t have push or pull classes, so a simple toggle of 'content-layout--rtl' would do the trick.

Might be worth a wider discussion before implementing this.

jrbarnes9:
Could we not just switch out `.push-1@m` with `.pull-1@m`, and add `grid--reverse` for RTL templates?

trevorsaint:
Here is another idea.

Our container will have a `.rtl` class.

Each component will have to be updated to be aware of RTL support. As the example below demonstrates:

```
.rtl .panel {
   direction: rtl;
}

.rtl .list {
   direction: rtl;
}
```

boxadesign:
Thinking at a high level I would consider creating a separate stylesheet for this that works in the same way as our theme overrides (i.e. census.scss).

rmccar:
I think that Matts solution is probably the simplest way of doing it and stops us having to make changes to every component

trevorsaint:
Ok, so we create a `rtl.scss` stylesheet that handles the styling overrides.

PR #1296, opened 2021-01-13
Rtl

trevorsaint:
- I have added an example of a right to left template
- I have added a new pageConfig property `rtl`. If this exists it wraps the template with a div with a class name 'rtl'

This then allows for overriding components and grid properties. I may have missed some overrides, but these will be best spotted when used in Craft and updated accordingly.


src/styles/page-template/_template.njk
```diff
@@ -136,13 +136,15 @@
                                 }}
                             {% endif %}
                             {% block preMain %}{% endblock %}
+                            {% if pageConfig.rtl is defined and pageConfig.rtl %}<div class="rtl">{% endif %}
```
boxadesign, line 139: Census website templates replace the higher level `pageContent` block. This is due to a legacy issue where the grid was set at `grid-col-8`. This has now been reworked in the DS so that main grid col can be set using `pageColNumber` but to bring this into the generator would require reworking the templates so we replace the `main` block instead.

If this isn't something we want to do now you could add the class to the `body` tag.

boxadesign (approved):

jrbarnes9:
[Issue #1278: Set page contents to be displayed right-to-left using css `direction: rtl`](https://trello.com/c/Z3LgBl58/854-issue-1278-set-page-contents-to-be-displayed-right-to-left-using-css-direction-rtl)

PR #1301, opened 2021-01-18
Added retina support and documentation to support component

trevorsaint:
Updated initially for CRAFT usage with `srcSmall` and `srcLarge` properties.

Added retina flexibility for image.

It now uses the same approach as used in Article, Hero, Downloads components in providing a `smallSrc` and a `largeSrc`.

Component will fallback if none specified to the old `url` approach.

Added documentation and macro options with a few style adjustments.

See how it looks :).

### DIFF
--- a/src/components/images/_images.scss
+++ b/src/components/images/_images.scss
@@ -1,12 +1,15 @@
 .figure {
   margin: 0;
+  text-align: center;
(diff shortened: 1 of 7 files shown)


boxadesign (approved):
lgtm

PR #1310, opened 2021-01-20
Fix error panels being focusable and tabbable

rmccar:
DAC issue: make error summary panels be focusable but not tabbable

### How to review 
Check that error summary panels are focused on page load but not tabbable

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -1,4 +1,3 @@
 export default function errorPanel(panel) {
-  panel.setAttribute('tabIndex', -1);
-  panel.focus();
+  panel.scrollIntoView();
 }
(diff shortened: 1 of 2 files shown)


rmccar:
[Focus Order (A)](https://trello.com/c/S703foJP/897-focus-order-a)

src/components/panel/panel.js
```diff
@@ -1,4 +1,4 @@
 export default function errorPanel(panel) {
-  panel.setAttribute('tabIndex', 0);
+  panel.setAttribute('tabIndex', -1);
```
boxadesign, line 2: We can just remove this line as the panel doesn't have an index value by default.

rmccar: Doesnt work if I remove that line because the panel is no longer focusable. So Ive removed it and used `scrollIntoView() `instead

boxadesign (approved):

PR #1311, opened 2021-01-20
Autosuggest - Additional aria-label on results <ul>

boxadesign:
DAC review suggested an additional `aria-label` is required on the autosuggest results `<ul>` element.

BREAKING CHANGE - A new parameter `ariaResultsLabel` needs to be added to `onsAutosuggest` implementations. The value should be a string that is contextual to the current scenario i.e.  `'ariaResultsLabel': 'Passport suggestions'`.

### How to review 
Check the parameter is rendered as an `aria-label` html attribute.


rmccar (approved):

jrbarnes9:
[Fix ARIA for autosuggest](https://trello.com/c/HLn9KgPp/901-fix-aria-for-autosuggest)

PR #1313, opened 2021-01-21
Replace highlight with prismjs

rmccar:
Due to the security issue in highlightjs v9. I investigated into upgrading to v10 but v10 doesnt support ie11. So this PR is to replace highlightjs with prismjs which does support ie11.

As part of this I needed to rename the `tag` css class used in the find a support centre pattern

### How to review 
- Check that code blocks in example tabs look correct and code is styled
- Check that find a support centre pattern still looks as it should


rmccar:
[Replace highlightjs with prismjs](https://trello.com/c/JBxkI38T/889-replace-highlightjs-with-prismjs)

boxadesign:
@rmccar Looks like it all works now. I'd maybe switch the font to monospace.

rmccar:
> @rmccar Looks like it all works now. I'd maybe switch the font to monospace.

I think it is a monospace font its using the font family `Consolas,Monaco,Andale Mono,Ubuntu Mono,monospace`

boxadesign (approved):

jrbarnes9:
[Replace highlightjs with prismjs](https://trello.com/c/JBxkI38T/889-replace-highlightjs-with-prismjs)

PR #1314, opened 2021-01-21
Added more flexibility for related content and combined related links

### LINKED ISSUE #1312: Need to differentiate types of related content by sub headings
We need to combine related content and related links.

We also need to extend the functionality, similar to how we have multiple blocks in the footer. The same sort of thing will happen for related content blocks.

**See image as an example**

Screenshot 2021-01-21 at 10 04 51

trevorsaint:
This component should be used in favour of related links. 

- Added more control and flexibility for related content
- Added documentation and macro options

We should be able to remove the related links component.

**Includes multiple row option**

Screenshot 2021-01-21 at 10 04 51

Closes #1312


boxadesign:
Looks good but all examples seem to have the census theme. Maybe good to have a default

trevorsaint:
> Looks good but all examples seem to have the census theme. Maybe good to have a default

Ok @boxadesign I will get one added.

trevorsaint:
[Consider merging related content and related links components](https://trello.com/c/GmW2dsVo/865-consider-merging-related-content-and-related-links-components)

jrbarnes9 (commented):
So can we remove 'related links' component now? It is possible only the Census website using it.  Maybe leave it in as we don't know what other services are using is, but we should at least remove component from documentation, so no one uses it going forward.  Perhaps put in release notes that this new macro should replace it? Then remove component macro later?

trevorsaint:
> So can we remove 'related links' component now? It is possible only the Census website using it. Maybe leave it in as we don't know what other services are using is, but we should at least remove component from documentation, so no one uses it going forward. Perhaps put in release notes that this new macro should replace it? Then remove component macro later?

@jrbarnes9 I did want to remove the related links, but wanted to leave it in just in case it was being used elsewhere. In regards to release notes. Good idea. Both @boxadesign and @rmccar do the releases.

src/components/related-content/_macro.njk
```diff
@@ -1,15 +1,43 @@
 {% macro onsRelatedContent(params) %}
 
+{% from "components/lists/_macro.njk" import onsList %}
+
 {% if params is defined and params and params.classes is defined and params.classes %}
     {% set classes = ' ' + params.classes %}
 {% endif %}
 
   <div class="related-content{{ classes }}">
```
jrbarnes9, line 9: I think this could probably be improved using `<aside>` ? 
```suggestion
  <aside class="related-content{{ classes }}" role="complementary">
```
See: https://design-system.service.gov.uk/patterns/start-pages/

trevorsaint: @jrbarnes9 Good suggestion. I can update this, but this needs to sit outside of `main`, so the main template will need updating to allow for this. Currently the template wraps everything inside `main`

boxadesign (approved):

boxadesign (approved):

PR #1317, opened 2021-01-21
Added property to all SVGs (IE bug fix DAC)

### LINKED ISSUE #1315: SVG images are focusable in Internet Explorer
SVG images are focusable in Internet Explorer.

We need to add `focusable="false"` to all of our SVGs.

This will ensure they are not in tab sequence.

They also need to be hidden from screen reader users using `aria-hidden` or given a descriptive text.

Screenshot 2021-01-21 at 15 11 18

trevorsaint:
## DAC Accessibility finding

Added `focusable="false"` to all SVGs. In Internet Explorer all SVGs are focusable. This property ensures they are not in the tab sequence.

I have added `aria-hidden` property for the social icons specifically. Other icons require a little more thought and could include descriptive text perhaps.

Closes #1315

### DIFF
--- a/src/components/icons/_macro.njk
+++ b/src/components/icons/_macro.njk
@@ -1,62 +1,62 @@
 {% macro onsIcon(params) %}
     {% if params.icon == "check" %}
-        <svg class="{% if params.iconSize %}svg-icon svg-icon--{{ params.iconSize }}{% else %}svg-icon{% endif %}" viewBox="0 0 13 10" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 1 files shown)


rmccar (dismissed):

rmccar:
Don't know if its worth adding a comment to the icons file as this fix is only for ie11. Im happy either way

rmccar (approved):

boxadesign (approved):

jrbarnes9:
[Remove svgs from focus order](https://trello.com/c/loy0sDV4/915-remove-svgs-from-focus-order)

PR #1318, opened 2021-01-21
Upgrade karma to v6

rmccar:
This is to fix a vulnerability in the dependancy socket.io which is a dependancy of karma.

Updating to v6 caused the error "Some of your tests did a full page reload!" when running locally. By removing the `clearContext: false,` flag I was able to avoid this.

Also in the radio tests there were errors in the console because there was more than one other input textbox with the same id. So I have added ids to these to fix this.

### How to review
- Changes make sense
- Tests pass


rmccar:
[Upgrade Karma to v6.0.1 for vulnerability](https://trello.com/c/0DnfkmCZ/905-upgrade-karma-to-v601-for-vulnerability)

boxadesign (approved):

PR #1320, opened 2021-01-25
Added DAC suggestions to warning panel for accessibility reasons

trevorsaint:
Warning panel has been updated to meet DAC accessibility requirements

Previous style attributes had not conveyed the important information for screen readers. A visual user would have understood the message whereas a screen reader would have read it out like any other paragraph in the site with no emphasis.

The use of strong has been added as well as screen reader only assistive purpose 'Warning' span.

**DAC suggested pattern below**

Screenshot 2021-01-25 at 13 20 24

This has been applied using our own visually hidden class.


trevorsaint:
[Warning panel updates](https://trello.com/c/ptdTbqye/924-warning-panel-updates)

jrbarnes9:
[Warning panel icon has no text alternative](https://trello.com/c/iUDedDMh/916-warning-panel-icon-has-no-text-alternative)

src/components/panel/_macro.njk
```diff
@@ -30,6 +30,8 @@
 
             {% if params is defined and params and params.type == "warn" or params.type == "warn-branded" %}
                 <span class="panel__icon" aria-hidden="true">!</span>
+                <strong>
+                <span class="u-vh">Warning</span>
```
jrbarnes9, line 34: I think this should be a param with a default.

```suggestion
                <span class="u-vh">{{ params.iconText | default("Warning: ") }}</span>
```

trevorsaint: I can sort that out today @jrbarnes9

boxadesign (approved):

PR #1321, opened 2021-01-25
Removed some aria attributes causing accessibility problems (DAC)

trevorsaint:
From the **DAC report** a few items were causing big issues for screen readers, and as such have been removed.

Each landmark region was marked `role="navigation"` causing confusion when navigating the content on the page. 

Removing both `aria-label` and `role="navigation"` resulted in a more efficient experience for a screen reader user.


trevorsaint:
[Card component: Multiple navigation Landmarks (A)](https://trello.com/c/oy61czef/918-card-component-multiple-navigation-landmarks-a)

boxadesign (approved):

PR #1326, opened 2021-01-26
added additional is defined param to list macro

boxadesign:
Issue with the way we run a condition for a parameter. Works ok with nunjucks but when brought into a python app with jinja2 it throws an error. Requires an additional `is defined`.

https://trello.com/c/OEa7M5LI


jrbarnes9:
[Fix list macro to work with jinja2](https://trello.com/c/OEa7M5LI/1006-fix-list-macro-to-work-with-jinja2)

jrbarnes9 (approved):

Issue #1322, opened 2021-01-25
Download thumbnail images shouldn't require `alt` text, as they are decorative


jrbarnes9:
Download thumbnail images shouldn't require `alt` text, as they are decorative

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/716#discussioncomment-308753_

PR #1328, opened 2021-01-26
Removed alt tag property from document download thumbnail

trevorsaint:
Thumbnail is decorative and does not convey important information to screen readers and has been assigned a null alt attribute to enable screen reading software to disregard it.

**Ref:** P16 in the DAC Accessibility report.

### DIFF
--- a/src/components/downloads/_macro-options.md
+++ b/src/components/downloads/_macro-options.md
@@ -11,11 +11,10 @@
 
 ## Thumbnail
 
-| Name     | Type   | Required | Description                                                 |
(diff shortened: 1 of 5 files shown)


boxadesign (approved):

jrbarnes9:
[Issue #1322: Download thumbnail and similar images shouldn't require `alt` text, as they are decorative](https://trello.com/c/UBfpYKhN/993-issue-1322-download-thumbnail-and-similar-images-shouldnt-require-alt-text-as-they-are-decorative)

PR #1338, opened 2021-01-28
Left align for image and caption. Updated docs and macro options

trevorsaint:
Image and caption is now left aligned.

Added some information about when not to add an alt tag (decorative image).

Alt tag is not required.


jrbarnes9 (approved):

jrbarnes9:
[Figure images and captions should be left-aligned](https://trello.com/c/yhBSS0ej/1015-figure-images-and-captions-should-be-left-aligned)

PR #1344, opened 2021-02-01
Error summary - Set title tag default to h2 (DAC)

boxadesign:
From the  DAC review we now set the panel header title tag to H1 if the panel type is error.

### How to review 
Check to see the h1 renders in the error summary example.

https://trello.com/c/JgPlD8xa


jrbarnes9:
Just need to remove all the `titleTag` params in the correct errors pattern

jrbarnes9 (approved):

PR #1354, opened 2021-02-04
1025 skip link tabindex

jrbarnes9:
The js solution for the skip to content link, preventing the need to append `#main-content` to the url (which persists when moving to next page in eQ), was adding `tabindex="0"` to the `<main>` element, which meant it was then part of the tab order.  So when the user tabs back up to the header, the focus indicator is lost when `<main>` is in focus.

#### Changes:

- Changed js `tabindex` value to `-1` so it's focusable by JS, but not put in tab order
- Updated custom page template example, moving phase banner to a different block, to ensure skip links are in the correct position
- Add TOC to RTL page template guide

### How to review 
Check skip to content link works as expected, and user can tab back up without losing focus indicator

### DIFF
--- a/src/components/skip-to-content/skip-to-content.js
+++ b/src/components/skip-to-content/skip-to-content.js
@@ -2,7 +2,7 @@ export default function skipToMain(link) {
   const id = link.getAttribute('href').replace('#', '');
   link.addEventListener('click', event => {
     event.preventDefault();
-    document.getElementById(id).tabIndex = 0;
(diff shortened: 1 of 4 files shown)


jrbarnes9:
[#1354 Change tabindex value in skip link js](https://trello.com/c/tZ48NFxn/1025-1354-change-tabindex-value-in-skip-link-js)

boxadesign (approved):

PR #1359, opened 2021-02-08
Update axios dep for security vulnerability

rmccar:
`axios` which is a dependancy of `@percy/agent` which is a dependancy of `@percy/script` needs to be updated to v0.21.1. The latest version of `@percy/agent` (v0.28.6) has this vulnerability fixed so we need to upgrade to that. But `@percy/script` hasn't been upgraded to the latest version of `@percy/agent` yet. So I have used resolutions to force it to use the latest version of `@percy/agent`

### How to review
- Change makes sense
- Tests pass
- Run `yarn audit` to see that vulnerability is gone

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "@babel/preset-env": "^7.1.6",
     "@babel/register": "^7.0.0",
     "@babel/runtime": "^7.2.0",
+    "@percy/agent": "^0.28.6",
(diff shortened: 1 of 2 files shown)


rmccar:
[Security vulnerability axios](https://trello.com/c/0O7fy3hy/1032-security-vulnerability-axios)

boxadesign (approved):

PR #1375, opened 2021-02-17
Fix dsExample param on error panels

rmccar:
This is to fix the logic around the `dsExample` param on error panels. When set to true it is designed to not autofocus on the error panels when used in the Design System examples. It was no longer stopping the autofocus, this will fix that.

### How to review 
- Check that on the panels page the error panels aren't autofocused
- Check to see that they are autofocused if the `dsExample` param is removed


rmccar:
[Error panel: Include `autofocus` as part of `dsExample` param to prevent focussing panel in documentation](https://trello.com/c/XZ9OrLZ5/1027-error-panel-include-autofocus-as-part-of-dsexample-param-to-prevent-focussing-panel-in-documentation)

boxadesign (approved):

jrbarnes9 (approved):

PR #1404, opened 2021-02-23
Added guide pattern page

boxadesign:
Added a new pattern for a guide.
### How to review 
Check examples and documentation.

https://trello.com/c/71iQLE9m


src/patterns/guide/index.njk
```diff
@@ -0,0 +1,66 @@
...
+
+View the nunjucks output below which shows the components required to create the guide pattern.
+
+The first page of a guide serves as the "parent" to the other pages. In the table of contents it should be linked as "Overview". 
+For accessibility the title `h1` should contain an additional visually hidden `span` that provides the additional "Overview" text e.g.
+
+`<h1 class="u-mb-l">`<br>
+&nbsp;&nbsp;`<span class="u-fs-xxl u-db">About the census </span>`<br>
+&nbsp;&nbsp;`<span class="u-vh">: Overview</span>`<br>
+`</h1>`<br><br>
```
rmccar, line 30: Shouldn't we style these codeblocks like other codeblocks for continuity? Like the one on the "Get Started" page

boxadesign: I've just attempted to use the same markup, but it doesn't work. It ignores certain tags and I think it needs a convoluted class name combination to get it to work. I'd rather come up with a better solution.

rmccar (dismissed):

rmccar:
#1423 Has merged so we can use that in the code highlighting now

rmccar (approved):

jrbarnes9 (approved):

PR #1412, opened 2021-03-02
Replace sass-lint with stylelint

rmccar:
Sass-lint is deprecated, no longer maintained and uses a dependency that has a security issue (minimalist). This dependancy was updated in other packages we have but the version with the fix is incompatible with sass-lint

This PR updates the DS to use stylelint, sets out the rules for our css, removes sass-lint and fixes our css to fit in with the rules from stylelint.

### How to review 
- Check there are no changes to the way components look or work
- Review stylelint rules and check that we are happy with the changes
- See that when running `npm audit` there is now no vulnerabilities


rmccar:
[Replace sass-lint](https://trello.com/c/Kg6lltIV/1035-replace-sass-lint)

boxadesign (dismissed):

boxadesign (approved):

jrbarnes9 (approved):

boxadesign (approved):

jrbarnes9 (approved):

PR #1423, opened 2021-03-05
Add code highlight component

rmccar:
This adds the code highlight component to the DS. It allows us to highlight code more easily outside of the tabs in DS examples because we can take advantage of the parsing that happens when the code is passed as a param in nunjucks and gives a consistent way to highlight code across the DS.

### How to review 
All tabs in examples still highlight code correctly and code highlighted in the content of the "Get Started" page is formatted in the same way.


rmccar:
[Find better solution to code highlighting than the one used on `get started` page](https://trello.com/c/MJXwMqh9/1068-find-better-solution-to-code-highlighting-than-the-one-used-on-get-started-page)

boxadesign (approved):

jrbarnes9 (approved):

Issue #1426, opened 2021-03-08
Promo banner should use wider grid col if no image


jrbarnes9:
Promo banner should use wider grid col if no image

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/730#discussioncomment-451796_

PR #1427, opened 2021-03-09
Expand promo banner text grid col when no image

jrbarnes9:
Updated promo banner text only grid col from 6 to 7

### How to review 
Check promo banner variants


jrbarnes9:
[Issue #1426: Promo banner should use wider grid col if no image](https://trello.com/c/zf7Rb3Fq/1086-issue-1426-promo-banner-should-use-wider-grid-col-if-no-image)

boxadesign (approved):

rmccar (approved):

PR #1428, opened 2021-03-10
Updated link as button behaviour for accessibility

boxadesign:

Accessibility changes -
- A link styled as a button must allow the `spacebar` to trigger the link as well as `enter`.
- Changed the "reset filter" styling on download resources to a secondary button.
https://trello.com/c/tpVeyfl3/1083-ensure-link-and-button-behaviours-match-their-roles

### Breaking change
Will require the download resources template to include additional classes on the reset button/link. See diff for classes.


jrbarnes9 (dismissed):

jrbarnes9:
Does it need a percy check?

rmccar:
I think we should add some more to the documentation around buttons that are styled links. Theres not really much in there. I don't mind if you want to add that as a separate card or add some docs in this PR. One other thing I noticed was the ghost button is a styled link but doesn't have the `btn--link` class. Other than that it looks good

jrbarnes9 (approved):

rmccar (approved):

rmccar (approved):

jrbarnes9 (approved):

PR #1425, opened 2021-03-08
Removed min-width from checkbox__item if being used inside filters

trevorsaint:
Fixes #1424

PR #1432, opened 2021-03-11
Download resources remove min width census

rmccar:
Same as #1425 for census branch


rmccar:
[Remove min width from checkbox when being used inside filters](https://trello.com/c/aE3sdVP0/1088-remove-min-width-from-checkbox-when-being-used-inside-filters)

trevorsaint (dismissed):

PR #1437, opened 2021-03-11
Added an alternative version for call to actions 'Announcement'

trevorsaint:
Call to action component has been extended to allow for single or multiple announcement.

BlackBanner

**This pattern is pretty much identical to what GDS have done for GOV.UK that can be seen here:** 
https://www.gov.uk

**Trello card can be found here:**
https://trello.com/c/Z2LaG5a3/47-blackbanner

### DIFF
--- a/src/components/call-to-action/_call-to-action.scss
+++ b/src/components/call-to-action/_call-to-action.scss
@@ -1,8 +1,57 @@
 .call-to-action {
+  background-color: $cta-bg;
   padding: 0.85rem 0;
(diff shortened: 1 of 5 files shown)


trevorsaint:
[Royal Death - Black Banner](https://trello.com/c/yFYjAW7q/1090-royal-death-black-banner)

trevorsaint:
We are still unsure of this pattern at the moment and may ditch it for a different approach! We are currently in talks about it due to limited information.

PR #1441, opened 2021-03-15
Added new panel for important announcement

trevorsaint:
Added a new panel similar to the branded warning panel.

Screenshot 2021-03-15 at 13 26 31

Screenshot 2021-03-15 at 13 27 05


trevorsaint:
[Royal Death - Black Banner](https://trello.com/c/yFYjAW7q/1090-royal-death-black-banner)

jrbarnes9 (commented):
Just need to add the new type to the macro options table

trevorsaint:
> Just need to add the new type to the macro options table

Missed that one. Thanks will do now.

src/components/panel/examples/announcement/index.njk
```diff
@@ -0,0 +1,18 @@
...
+{% call onsPanel({
+        "type": 'announcement'
+    })
+%}
+
+<p class="u-mb-no">National lockdown: stay at home</p>
+
+<div class="u-fs-r">
+    <p>Coronavirus (COVID-19) remains a serious threat across the country. <a href="#0">Find out more</a></p>
+</div>
```
jrbarnes9, line 16: This will probably have to use WYSIWYG in Craft cms, so we wouldn't want to have to manually add html markup.  I'd suggest remove the bold css from the announcement style type so it can be added (if required) using `strong` tags in the editor.  

The first line would probably have to be a heading? So might want to control font size of `h2 ` in announcement type with css

trevorsaint: When we know what the content is we can revisit this component.

jrbarnes9 (approved):

jrbarnes9 (dismissed):

src/components/panel/_macro.njk
```diff
@@ -14,12 +14,20 @@
         {% set typeClass = ' panel--info' %}
     {% endif %}
 
+    {% if params is defined and params.type == "warn-branded" %}
+        {% set containerClass = 'census-warning' %}
+    {% endif %}
+
+    {% if params is defined and params.type == "announcement" %}
+        {% set containerClass = 'census-announcement' %}
```
boxadesign, line 22: I don't think the class should reference census as it could be used across any service

boxadesign (approved):

rmccar (approved):

PR #1444, opened 2021-03-18
Replace nunjucks-loader to remove deprecated warning

rmccar:
This PR is to remove the deprecated warning in our builds and to do this we need to migrate to `simple-nunjucks-loader`

As part of this, this PR also updates webpack to the newest version of v4

### How to review 
Tests pass and warning is gone in logs


rmccar:
[Nunjucks-loader causes warning when running yarn test](https://trello.com/c/76rAbwiX/1093-nunjucks-loader-causes-warning-when-running-yarn-test)

PR #1445, opened 2021-03-19
Add templates.zip github action

rmccar:
Added github action to add templates.zip to release and remove the current process from the AWS pipeline.

### How to review 
- Can be tested locally using https://github.com/nektos/act
- Code looks correct


rmccar:
[Github actions to build design-system artefacts and replace travis](https://trello.com/c/vLc2I5EL/800-github-actions-to-build-design-system-artefacts-and-replace-travis)

rmccar:
[Add github action to add templates.zip to release](https://trello.com/c/5zYC0Xrx/1095-add-github-action-to-add-templateszip-to-release)

ci/concourse.yml
```diff
@@ -121,9 +121,3 @@ jobs:
             title: Concourse Build $BUILD_ID
```
MebinAbraham, line 121: You should remove the zip command from the build task.

.github/workflows/zip-templates-on-release.yml
```diff
@@ -0,0 +1,32 @@
...
+      - run: mkdir ../$GITHUB_RELEASE_TAG
+      - run: cd ../
+      - run: zip -r templates/templates.zip templates/*
+      - run: cp -R build/* ../$GITHUB_RELEASE_TAG
+      - uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: templates/templates.zip
+          tag: $GITHUB_RELEASE_TAG
+          overwrite: true
```
MebinAbraham, line 32: I think it would be clearer as 3 jobs (one to install dependencies, one to build, another to zip and upload the templates zip). Each job can have multiple steps to break them up.

.github/workflows/zip-templates-on-release.yml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  zip-templates:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - run: npm install -g yarn
+      - run: yarn install
```
MebinAbraham, line 14: Multiple run commands can be combined like:
```sh
- run: |
	npm install -g yarn
	yarn install
```

.github/workflows/zip-templates-on-release.yml
```diff
@@ -0,0 +1,32 @@
...
+      - run: sudo apt-get update && sudo apt-get install -y --allow-unauthenticated zip
+      - run: curl -s https://api.github.com/repos/ONSdigital/design-system/releases/latest | jq -r ".tarball_url" | wget -i - -O source.tar.gz
+      - run: mkdir design-system
+      - run: tar -xzf source.tar.gz -C design-system --strip-components=1
+      - run: cd design-system
+      - run: echo "GITHUB_RELEASE_TAG=$(curl -s https://api.github.com/repos/ONSdigital/design-system/releases/latest | jq -r ".tag_name")" >> $GITHUB_ENV
```
MebinAbraham, line 20: You don't need to hit the GitHub API for the release version. This release event has multiple payload properties that can be used: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release
The `GITHUB_REF` env var will give you the tag.

MebinAbraham:
Can you squash the commits?

.github/workflows/zip-templates-on-release.yml
rmccar: Now I've got this working again and cut down the amount of commands needed, Im not sure its worth splitting it into 3 jobs. Because the way I would split it would mean the last 2 jobs would just be one command each. So what I've done instead is split it down into named steps

.github/workflows/zip-templates-and-add-to-release.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Install dependencies
+        run: |
+          npm install -g yarn
+          sudo apt-get update && sudo apt-get install -y --allow-unauthenticated zip
+          yarn install
+      - name: Download latest release
+        run: |
+          curl -s https://api.github.com/repos/ONSdigital/design-system/releases/latest | jq -r ".tarball_url" | wget -i - -O source.tar.gz
+          mkdir design-system
+          tar -xzf source.tar.gz -C design-system --strip-components=1
```
ajmaddaford, line 22: Do we need to do this if the `checkout` action has checked out the repo commit for the release?

.github/workflows/zip-templates-and-add-to-release.yml
```diff
@@ -0,0 +1,33 @@
...
+  release:
+    types: [published]
+jobs:
+  add-templates-to-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
ajmaddaford, line 12: This may not be needed as Node version 14 is installed in the ubuntu image being used - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#language-and-runtime

.github/workflows/zip-templates-and-add-to-release.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Install dependencies
+        run: |
+          npm install -g yarn
+          sudo apt-get update && sudo apt-get install -y --allow-unauthenticated zip
```
ajmaddaford, line 16: This may not be needed as it looks like yarn and zip should already be installed for `ubuntu-latest` - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

.github/workflows/zip-templates-and-add-to-release.yml
rmccar: I tried removing this and got a permissions error when running it without this

.github/workflows/zip-templates-and-add-to-release.yml
rmccar: When running without either of these lines the relevant command fails because it's not installed. Im not sure why exactly if they are supposed to be installed in that version ubuntu. Probably a similar reason as the node comment above

.github/workflows/zip-templates-and-add-to-release.yml
rmccar: Yeah you're right, don't need to do it. Ive sorted that and removed the "Download latest release" step

.github/workflows/zip-templates-and-add-to-release.yml
ajmaddaford: I have tried it without these installs and it works. Tested in my fork at https://github.com/ajmaddaford/design-system/runs/2390030841?check_suite_focus=true.

.github/workflows/zip-templates-and-add-to-release.yml
```diff
@@ -0,0 +1,29 @@
...
+          yarn install
+      - name: Build templates
+        run: |
+          cd src
+          RELEASE_VERSION=$GITHUB_REF yarn cdn-bundle
```
ajmaddaford, line 21: `GITHUB_REF` will be `ref/tags/<version>`, this needs to be changed to:
```suggestion
          RELEASE_VERSION=${{ github.event.release.tag_name }} yarn cdn-bundle
```

.github/workflows/zip-templates-and-add-to-release.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Zip templates
+        run: zip -r templates.zip templates/*
+      - uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: templates.zip
+          tag: $GITHUB_REF
+          overwrite: true
```
ajmaddaford, line 29: We add a file to a release in the schemas repo using `softprops/action-gh-release` (https://github.com/ONSdigital/eq-questionnaire-schemas/blob/master/.github/workflows/release.yml#L58-L63). We should use the same action here for consistency:
```suggestion
      - name: Release
        uses: softprops/action-gh-release@v1
        with:
          files: templates.zip
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

ajmaddaford (changes requested):

.github/workflows/zip-templates-and-add-to-release.yml
```diff
@@ -0,0 +1,27 @@
+name: zip-templates-and-add-to-release
```
ajmaddaford, line 1: In other repos we use use the filename `release.yml` and the name `Release`. We should consider doing the same for this repo for consistency. It would also mean we could add other jobs to this file in future without needing to change the name.

.github/workflows/zip-templates-and-add-to-release.yml
rmccar: Seems to work if I push changes. I think it wasn't working for me because I tested it using act. It fails if I run it with that

ajmaddaford (approved):

MebinAbraham (approved):
Actions working as expected on a fork.

PR #1458, opened 2021-03-30
Remove sticky header nav and make DS responsive

jrbarnes9:
The DS isn't currently responsive, so can't be used on a mobile device.  

### How to review 
Review the following changes...
- Used grid and `wide` container across the site layout so homepage layout doesn't differ to page layout
- Removed the 'sticky' header and contents nav
- Added `grid--reverse` so contents nav can be displayed below content when stacked on mobile (as an MVP and no js solution - [new ticket raised](https://trello.com/c/xmbspduF/1130-add-sub-navigation-to-mobile-menu) to combine subnav with mobile menu)
- Added mobile-only sub-heading to sub nav e.g. "More in 'Styles'"
- Added breadcrumbs for mobile
- Added `group` sub-heading to `H1`
- Added `topLevel` variable to frontmatter so top level pages can display a 'contents' sub heading and 'home' breadcrumb on mobile
- Linked up title and masthead logo
- Removed the subnavigation js class as `scrollIntoView()` no longer required


lib/nunjucks-html-loader.js
```diff
@@ -48,6 +48,7 @@ function buildSiteMap(context) {
     const group = frontmatter.group || '';
     const lang = frontmatter.lang || '';
     const experimental = frontmatter.experimental || '';
+    const topLevel = frontmatter.topLevel || '';
```
boxadesign, line 51: Could possibly remove the `topLevel` code completely and check if a `parent` exists and what it is. Should be able to implement something like `{% if pageInfo.parent[0].title === 'home' %}` in the template.

jrbarnes9: Cheers.  I was hoping there would be a better way in the templating.  Will take a look

boxadesign: Should save the overhead of remembering to add frontmatter

jrbarnes9: Yep deffo.  Worked a treat 👍

jrbarnes9:
[Remove sticky header/nav from patternlib](https://trello.com/c/t7VxTJRC/130-remove-sticky-header-nav-from-patternlib)

jrbarnes9:
[Make pattern library layout responsive](https://trello.com/c/1umoceuv/124-make-pattern-library-layout-responsive)

boxadesign (approved):

src/components/hero/_macro.njk
```diff
@@ -14,8 +14,8 @@
     {% set containerClasses = " hero__container--has-collapsible" %}
 {% endif %}
 
-{# if hero is census specific, change column usage #}
-{% if params.censusTheme is defined and params.censusTheme %}
+{# if hero is census-specific, change column usage #}
```
rmccar, line 17: Probably should be if it is census specific or wide page needed, or something like that

src/views/partials/example/_macro.njk
```diff
@@ -1,7 +1,14 @@
 {% macro patternlibExample(params) %}
     {% set params = helpers.generateExampleParams(params, helpers.addDependency) %}
     <div class="patternlib-example js-example">
-        <a href="{{ params.link }}" target="_blank" class="patternlib-example__link u-fs-s">Open this example in a new window</a>
+        <div class="patternlib-example__header">
+            <a href="{{ params.link }}" target="_blank" class="patternlib-example__header-link external-link u-fs-s">Open this example in a new tab
+            <svg id="external-link" class="svg-icon" viewBox="0 0 12 12" xmlns="http://www.w3.org/2000/svg">
```
rmccar, line 6: Should us the external link macro here instead but will also then need to change that macro to allow classes to be set

src/views/layouts/_home.njk
```diff
@@ -1,54 +1,16 @@
-{% extends "./_master.njk" %}
+{% extends "./_page.njk" %}
 
-{% block master_body %}
-    {% from "components/header/_macro.njk" import onsHeader %}
-    {% from "components/footer/_macro.njk" import onsFooter %}
-    {% from "components/skip-to-content/_macro.njk" import onsSkipToContent %}
+{% from "components/header/_macro.njk" import onsHeader %}
+{% from "components/footer/_macro.njk" import onsFooter %}
+{% from "components/skip-to-content/_macro.njk" import onsSkipToContent %}
```
rmccar, line 5:
```suggestion
```
These are no longer needed

src/views/partials/sub-navigation/_macro.njk
```diff
@@ -1,20 +1,20 @@
...
-                      <ul class="patternlib-subnav__sub-items">
-                          {% for anchor in item.anchors %}
-                              <li class="patternlib-subnav__item">
-                                  <a href="{{ anchor.url }}" class="patternlib-subnav__link">{{ anchor.title }}</a>
-                              </li>
-                          {% endfor %}
-                      </ul>
-                  {% endif %}
-              </li>
+                <li class="patternlib-subnav__item{{ ' patternlib-subnav__item--active' if (item.url == params.currentPath) }} list__item">
```
rmccar, line 5: `patternlib-subnav__item` and `patternlib-subnav__list` are no longer adding any css to the elements, so do we need them anymore?

rmccar (changes requested):

src/views/partials/sub-navigation/_macro.njk
jrbarnes9: Should we keep that because we're still using `patternlib-subnav__item--active`?

rmccar: I see what you mean, but is there any point in having them if we never use them because were just using the css from lists anyway. They do make what the elements are more descriptive I suppose. Another thing we could do is something like this in the css:
```
patternlib-subnav__item {
  @extend .list__item;
}
```

Then we would only need the one class and it would still be descriptive

boxadesign (approved):

rmccar (approved):

PR #1465, opened 2021-03-31
Updated session timeout pattern to match what was delivered for Census2021

trevorsaint:
Session timeout pattern has been updated.

**As shown below**

example


jrbarnes9 (commented):
Actually, I think this version might be a better example to document: https://census-website-respond.netlify.app/end-to-end/session-timeout-rh-authentication-v2/ 

Just needs the documentation updating to explain:
- the circumstances when this page should be displayed (session timed out, followed a link)
- what information needs to be included in the bullets (timeout limit in minutes, the transaction the user is trying to access)
- that the heading needs to tell user what they have to do next (start again, or sign in to continue)
- that the cta needs to tell user what they will have to do next (start again, or sign in to continue)

boxadesign (dismissed):

rmccar (commented):
I agree with Jon, would be better if the documentation was updated

src/patterns/error-status-pages/examples/timeout-error/index.njk
```diff
@@ -17,13 +17,24 @@ layout: none
...
 
 {% block main %}
-    <h1>Your session has timed out due to inactivity</h1>
 
-    {% call onsPanel({
-        "classes": "u-mb-m"
-    }) %}
-        <p>To protect your information we have timed you out.</p>
-    {% endcall %}
+    <h1>Sorry, you need to start again</h1>
```
jrbarnes9, line 21: I've just realised this is one of two variations @LauraC119 .  This is the "transaction" timeout page, but there's also the "signed into a survey" timeout page: https://census-website-respond.netlify.app/end-to-end/session-timeout-rh-authentication-v2/ 
So we'll need to either show both examples, or document what you should put in the other case.  Considering the survey scenario is the most common, we should probably use that example.
We might then want to document what a service should use if the user isn't signed into the service via an access code e.g. "Sorry, you'll need to start again" (Transaction e.g. RH) or "Sorry, you'll need to sign in again" (RAS service)

LauraC119:
@trevorsaint there is a word missing from the H1 on the preview. needs to be: 
Sorry, you need to sign in again.


Sorry, didn't spot earlier

LauraC119:
Also need to remove full stop from heading

trevorsaint:
> Also need to remove full stop from heading

Not sure how I missed that. This has been removed.

LauraC119 (approved):

src/patterns/error-status-pages/index.njk
```diff
@@ -164,27 +164,36 @@ The page should have:
     patternlibExample({ "path": "patterns/error-status-pages/examples/503-error/index.njk" })
 }}
 
-## Timeout
+## 408 timeout error
```
jrbarnes9, line 167:
```suggestion
## Session timed out error
```

...after a discussion with the devs, 408 is not the correct code, and in the next round of changes on this pattern, we'll remove all other references to http codes

LauraC119 (approved):
All good from my point of view

src/patterns/error-status-pages/index.njk
```diff
@@ -164,27 +164,36 @@ The page should have:
     patternlibExample({ "path": "patterns/error-status-pages/examples/503-error/index.njk" })
 }}
 
-## Timeout
+## Session timed out error
 
-Use a Timeout page when the user has been timed out of the service due to inactivity.  Explain why the timeout has occured and reassure the user that their progress has been saved. Provide a link to allow them to sign in again.
+Use a timeout error if you need to store the user's information in a browser session with an expiry time to protect their information. When the session expires, use the timeout error message.
```
jrbarnes9, line 169:
```suggestion
Use a session timed out error if you need to store the user's information in a browser session with an expiry time to protect their information. When the session expires, use the timed out error message.
```

src/patterns/error-status-pages/index.njk
```diff
@@ -164,27 +164,36 @@ The page should have:
...
 
-The page should have:
+Timeout could be needed because a user has been inactive, or because they have followed a link to the middle of a process where the required session data is missing.
```
jrbarnes9, line 171:
```suggestion
This error could be needed because a user has been inactive, or because they have followed a link to the middle of a process where the required session data is missing.
```

src/patterns/error-status-pages/index.njk
```diff
@@ -164,27 +164,36 @@ The page should have:
...
             },
             {
-                "text": "a link allowing the user to sign in again"
+                "text": "have a link for the user to sign in again"
             }
         ]
     })
 }}
 
+### Timeout error when user is not signed in
```
jrbarnes9, line 193:
```suggestion
### Session timed out error when user is not signed in
```

LauraC119 (approved):

boxadesign (approved):

Issue #1475, opened 2021-04-09
Text input focus border styling is inconsistent across devices/browsers...


jrbarnes9:
Text input focus border styling is inconsistent across devices/browsers...
On Chroms MacOS it renders as it should...
![Screenshot 2020-11-23 at 10 29 38](https://user-images.githubusercontent.com/43346934/99951957-d05bcc80-2d76-11eb-8dfe-ac8dd0d70fb5.png)

But on Chrome Windows and Chrome Android it is rendering like...
![image](https://user-images.githubusercontent.com/43346934/99952168-24ff4780-2d77-11eb-94df-8d79bbe7e613.png)

<img width="333" alt="Screenshot 2021-04-09 at 15 38 46" src="https://user-images.githubusercontent.com/43346934/114196880-b3045500-9949-11eb-89d8-b6ae8af5bf4f.png">

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/1383#discussioncomment-384266_

PR #1487, opened 2021-04-13
Text input focus border styling fixed

trevorsaint:
Text input focus style, numeric and error styling was buggy in different browsers and also at various zoom levels.

Normally this would not be an issue, but because we use round corners we can’t approach focus styles like this because outline does not support border radius which is causing the visual bugs. Box shadow however is fine on an element that uses border-radius.

Fixes: https://github.com/ONSdigital/design-system/issues/1475

## Before / After

input

numeric

error


jrbarnes9 (approved):

trevorsaint:
[Issue #1475: Text input focus border styling is inconsistent across devices/browsers...](https://trello.com/c/UTZkXZjW/1120-issue-1475-text-input-focus-border-styling-is-inconsistent-across-devices-browsers)

jrbarnes9 (approved):

rmccar (approved):

PR #1493, opened 2021-04-16
Send a web form

trevorsaint:
Content changes from Laura.


rmccar (approved):

src/patterns/send-a-web-form/index.njk
```diff
@@ -19,13 +19,11 @@ Do not use this pattern if the user needs to speak to someone at the contact cen
...
-To help contact staff deal with web form enquiries, you should provide a set of categories based on the most common user needs from research.
+Give users a set of categories to choose from on the web form. These should be based on the most common user needs identified through user research. This helps contact centre staff deal with enquiries.
 
-You should only ask the user to provide the contact details the service needs to handle the request.
+Only ask the user for the contact details the service needs to handle the request.
 
-For Census 2021, a respondent could send a web form to the contact centre to receive the appropriate support to complete the Census.
-
-The census web form included the following, in this order:
+The web form pattern includes the following, in this order:
```
jrbarnes9, line 26: I thought we were making it clear that these are just an example of what you could use in the web form pattern.  Otherwise, I think a user would read this and think they have to ask the user for e.g. what country they live in every time they want to create a web form.

trevorsaint: @jrbarnes9 can you discuss with Laura. I think we are on the mindset of removing a census relationship.

LauraC119: Agree, that we should move away from census-based examples. Suggest changing this list to the following:

The following could be included on a web form:
What do you need help with? (radio selection)
Please provide more detail (field)
Name (field)
Email address (field)
Send message (button)

trevorsaint: Let me get your suggestion added. Thanks.

trevorsaint: Content changes have been added as per the documentation. Census theme removed in favour of our default ONS look and feel.

boxadesign:
@trevorsaint Not a major issue but perhaps replace the "done" button with a link.

trevorsaint:
> @trevorsaint Not a major issue but perhaps replace the "done" button with a link.

Perfectly valid point. I will update this now. Thanks.

boxadesign (approved):

jrbarnes9 (commented):
Laura just wants to replace "Follow this pattern..." with "Use this pattern..." across the board

trevorsaint:
> Laura just wants to replace "Follow this pattern..." with "Use this pattern..." across the board

Ok, will update that now.

jrbarnes9:
> > @trevorsaint Not a major issue but perhaps replace the "done" button with a link.
> 
> Perfectly valid point. I will update this now. Thanks.
> 
> Looking into it, the button macro does not support a button as a link without the arrow. The button component will need updating to allow this.

We just used the CTA button with the chevron in RH for census.  Looking back though, I do wonder if a standard link is better placed here, as it's the end of a transaction e.g. 'Back to homepage'

trevorsaint:
l

> > > @trevorsaint Not a major issue but perhaps replace the "done" button with a link.
> > 
> > 
> > Perfectly valid point. I will update this now. Thanks.
> > Looking into it, the button macro does not support a button as a link without the arrow. The button component will need updating to allow this.
> 
> We just used the CTA button with the chevron in RH for census. Looking back though, I do wonder if a standard link is better placed here, as it's the end of a transaction e.g. 'Back to homepage'

For me it looks odd with the arrow, that in my opinion starts a user journey and not a redirection back to home.

boxadesign (dismissed):

LauraC119:
one tiny thing - there needs to be a space between "text" and "area"

- Please provide more detail (textarea)

trevorsaint:
> one tiny thing - there needs to be a space between "text" and "area"
> 
> * Please provide more detail (textarea)

Hi Laura. I hear you, but textarea is an element in html so makes more sense this way.

LauraC119:
Fair enough

LauraC119 (approved):

jrbarnes9 (approved):

PR #1497, opened 2021-04-21
Phone numbers pattern

jrbarnes9:
Added a pattern for asking users for phone numbers

### How to review
Check the documentation and examples


boxadesign (dismissed):

JNATION85:
@LauraC119 I've updated the content based on your comments

LauraC119 (dismissed):
Looks good to me

rmccar:
Do we want to have any restrictions on what the user can actually enter into the input? As in not allow them to enter letters for example or put a limit on the amount of characters that the user can input?

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,65 @@
...
+To do this, set the autocomplete attribute on the input field to `tel`.
+
+You will need to use the autocomplete attribute to meet [WCAG 2.1 AA standards for accessibility](https://www.w3.org/TR/WCAG21/#identify-input-purpose).
+
+### Error messages
+{{
+    patternlibExample({"path": "patterns/phone-numbers/examples/phone-number-error/index.njk"})
+}}
+
+Use the [correct errors pattern](https://ons-design-system.netlify.app/patterns/error-validation) and show the error details above the input phone number area.
```
jrbarnes9, line 40: Move this sentence above the example

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,65 @@
...
+
+#### If nothing is entered and the answer is required
+Use “Enter [the required type of phone number]”, for example, “Enter a UK mobile number”.
+
+#### If a phone number is incorrect
+Use “Enter a UK mobile number in a valid format, for example, 07700 912345 or +44 7700 912345”.
```
jrbarnes9, line 46: What should the user say if they need to ask for a different type of phone number? Would the examples need to change?

jrbarnes9 (commented):
Just a few inconsistencies between this and the email addresses pattern...
- Do we want to mention the width of the input to allow for variations of a phone number? (i.e. with/without country code etc)
- Should we repeat the bullet list we used in 'how to use'?
- We don't mention autocomplete in email addresses, but maybe this is best suited somewhere else as a general rule? 
- "If the email address is missing" versus "If nothing is entered and the answer is required"
- "...to explain the error to the user and give them a clear example of a valid..." not in this pattern

src/patterns/phone-numbers/index.njk
LauraC119: Yes they would, but not sure we need to have an example for every type of number - could get very busy

LauraC119:
OK:
- Do we want to mention the width of the input to allow for variations of a phone number? (i.e. with/without country code etc)
yes, probably a good idea
- Should we repeat the bullet list we used in 'how to use'?
ah yes, good point on consistency. The email address list will work here
- We don't mention autocomplete in email addresses, but maybe this is best suited somewhere else as a general rule?
I guess it's something that comes up in quite a few of the components. So would it go in Getting started?
- "If the email address is missing" versus "If nothing is entered and the answer is required"
yep
- "...to explain the error to the user and give them a clear example of a valid..." not in this pattern
when I was looking at the error validation messages, repeating "to explain the error to the user " all the time was getting very repetitive and not sure it's needed

src/patterns/phone-numbers/index.njk
LauraC119: yes they would but do we need examples of them all. Could end up being a really long list

jrbarnes9: So maybe we add some guidance like: 

> "Make sure any examples phone numbers in the error message are suitable for the type of number you are asking for.  [Ofcom maintains a list of numbers](https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbers-for-drama) reserved for use in media that you can use.

JNATION85: Can we put conditions on the use of 'OTHER' numbers? So if we are only asking for a UK number that stands, but if we are open to any number can this be turned on with a different message?

JNATION85: > Do we want to have any restrictions on what the user can actually enter into the input? As in not allow them to enter letters for example or put a limit on the amount of characters that the user can input?

The validation is already built into it but we may need explain this more in the documentation. With regards to the character count, research has shown that we need to allow the user to enter phone numbers in a format that suits them, e.g: spaces so we shouldn't restrict the length.

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,119 @@
...
+}}
+
+### Explain why you need the phone number
+Use the optional label__description with the [input component](/components/input) for hint text to reassure users why you need a phone number and what it will be used for.
+
+## Make it clear what type of phone number you need
+Use the `label__description` for hint text to let users know what type of phone number you need, for example, a UK, international or mobile phone number.
+
+## Allow all phone number formats
+The phone number field must accommodate all phone number formats. The length of the field must allow for any spaces or common characters, for example, brackets, dashes or country codes.
```
jrbarnes9, line 52: @LauraC119 Think this could do with mentioning the user.  E.g. 
```suggestion
The phone number field must accommodate all phone number formats so the user can enter the number in a format they’re familiar with. The length of the field must allow for any spaces or common characters, for example, brackets, dashes or country codes.
```

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,119 @@
...
+
+### Validate phone numbers
+
+To help users enter a valid email address, you should:
```
jrbarnes9, line 56:
```suggestion
To help users enter a valid phone number, you should:
```

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,119 @@
...
+{{
+    patternlibExample({"path": "patterns/phone-numbers/examples/phone-number-error/index.njk"})
+}}
+
+#### If the phone number is missing
+Use “Enter [the required type of phone number]”, for example, “Enter a UK mobile number”.
+
+#### If a phone number is incorrect
+Use “Enter a UK mobile number in a valid format, for example, 07700 912345 or +44 7700 912345”, to give them a clear example of a valid phone number.
+Make sure any examples of phone numbers in the error message are suitable for the type of number you are asking for. {{
```
jrbarnes9, line 88: Needs a new line...
```suggestion

Make sure any examples of phone numbers in the error message are suitable for the type of number you are asking for. {{
```

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,119 @@
...
+    onsExternalLink({
+        "url": "https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbers-for-drama",
+        "linkText": "Ofcom maintains a list of numbers"
+    })
+}} reserved for use in media that you can use.
+
+### Use autocomplete
```
jrbarnes9, line 95: Think this should be a level 2 heading as it's not related to 'validate phone numbers'
```suggestion
## Use autocomplete 
```

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,119 @@
...
+            "linkText": "Design System forum"
+        })
+    }}
+{% endcall %}
+
+## Design System forum
+{{
+    onsExternalLink({
+        "url": "https://github.com/ONSdigital/design-system/discussions/1496",
+        "linkText": "Discuss 'Numeric values' on GitHub"
```
jrbarnes9, line 117:
```suggestion
        "linkText": "Discuss ‘Phone numbers’ on GitHub"
```

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,119 @@
...
+    patternlibExample({"path": "components/input/examples/telephone/index.njk"})
+}}
+
+## When to use this pattern
+Only ask users for a phone number when your service has a genuine need for this information. For example, to send an access code by text message. 
+
+## How to use this pattern
+You should always explain why you need the user's phone number, letting them know if you will contact them and when.
+
+Use an [input component](/components/input) and set the input type to `tel` to bring up the correct keypad on touch devices.
```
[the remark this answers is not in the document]
LauraC119: Yes for line 21 but it's already there in the first bullet point so I'd just add a second one about contact:

- tell users why you need a phone number and what it will be used for
- let users know if you will contact them and when
- make it clear what type of telephone number you need
- make sure you allow the user to enter a phone number in a familiar format
- help users enter a phone number in the correct format

But I think for line 23, the bit about the component to use doesn't normally go in the bullet list as it's a specific instruction rather than the rest of the list which is general guidance

src/patterns/phone-numbers/index.njk
LauraC119: I see your point but why did we decide not to include the "familiar" bit that's in the bullet list above? I'm sure we had it there but took it out! But I agree that it helps. Change slightly, though, so we avoid that hanging "with" and don't use "format" twice:
The phone number field must accommodate all phone number variations so the user can enter a number in the format they are used to.

boxadesign (dismissed):

rmccar:
[Telephone numbers](https://trello.com/c/SiLC22d0/137-telephone-numbers)

src/patterns/phone-numbers/index.njk
```diff
@@ -0,0 +1,122 @@
...
+        "linkText": "Ofcom maintains a list of numbers"
+    })
+}} reserved for use in media that you can use.
+
+## Use autocomplete 
+Use the `autocomplete` attribute when you are asking for a phone number as it helps the user fill out a form more quickly.
+
+To do this, set the autocomplete attribute on the input field to `tel`.
+
+You will need to use the autocomplete attribute to meet [WCAG 2.1 AA standards for accessibility](https://www.w3.org/TR/WCAG21/#identify-input-purpose).
```
rmccar, line 103:
```suggestion
You will need to use the autocomplete attribute to meet {{
    onsExternalLink({
        "url": "https://www.w3.org/TR/WCAG21/#identify-input-purpose",
        "linkText": "WCAG 2.1 AA standards for accessibility"
    })
}}
```
This should be styled as an external link

rmccar (approved):

boxadesign (approved):

PR #1503, opened 2021-05-04
Share page folder path update

trevorsaint:
Updated folder path to match the heading of 'share-page' instead of 'share'.

Updated example macro to use single quotes instead of double for consistency.


jrbarnes9 (approved):

trevorsaint:
[Updated share page path to match heading](https://trello.com/c/vcyZnq4q/1146-updated-share-page-path-to-match-heading)

boxadesign (approved):

PR #1508, opened 2021-05-07
Can't add classes to feedback div and border doesn't match the default header theme colour

jrbarnes9:
The feedback component was only built for census.  So this PR allows support for the ONS default brand and expanded the colours variables to enable this.
Also added ability to add classes to the feedback container which will allow more control of spacing with utility classes.

### How to review
Check the feedback component and colour variables


package.json
```diff
@@ -163,5 +162,7 @@
     "minimist": "^1.2.5",
     "**/@percy/agent": "^0.28.6"
   },
-  "dependencies": {}
+  "dependencies": {
+    "node-sass": "4.14"
```
boxadesign, line 166: Need to move this back to `devDependencies`

boxadesign (changes requested):

boxadesign:
image

Think the default bg change or hero needs a tweak.

boxadesign:
Still need to move `node-sass` into `devDependencies`

rmccar (approved):

boxadesign (approved):

PR #1526, opened 2021-05-21
Allow hidden inputs on forms inside onsTables

mfcrocker:
I needed to put an onsButton in an onsTable, but also needed it to have hidden form inputs.

### How to review
Use the form attribute of an onsTable `tds` property, and add a property of `hidden` to it - this should be an array of objects with `name` and `value` properties.

EG:

```
{% set formData = {
                        "action": url_for('messages_bp.create_message'),
                        "button": {
                            "text": "Create message",
                            "id": 'create-message-' + respondent.id,
                            "classes": 'btn--secondary btn-small',
                            "value": 'create-message-view',
                            "name": 'create-message'
                        },
                        "hidden": [
                            {
                                "name": "ru_ref",
                                "value": ru.sampleUnitRef
                            },
                            {
                                "name": "business_id",
                                "value": ru.id
                            },
                            {
                                "name": "business",
                                "value": ru.name
                            },
                            {
                                "name": "survey",
                                "value": survey[1][4:]
                            },
                            {
                                "name": "survey_id",
                                "value": survey[0]
                            },
                            {
                                "name": "msg_to_name",
                                "value": respondent.firstName + ' ' + respondent.lastName
                            },
                            {
                                "name": "msg_to",
                                "value": respondent.id
                            },
                            {
                                "name": "csrf_token",
                                "value": csrf_token()
                            }
                        ]
                    } %}

and then later
{
      "form": formData
}

### DIFF
--- a/src/components/table/_macro-options.md
+++ b/src/components/table/_macro-options.md
@@ -38,11 +38,19 @@
 
 ## form
 
-| Name   | Type                                   | Required | Description                               |
-| ------ | -------------------------------------- | -------- | ----------------------------------------- |
-| method | string                                 | false    | Default is `post` if no value is provided |
-| action | string                                 | true     | The `action` for the form                 |
-| button | `Button` [_(ref)_](/components/button) | false    | Configuration object for the form button  |
+| Name            | Type                                   | Required | Description                                 |
+| --------------- | -------------------------------------- | -------- | ------------------------------------------- |
+| method          | string                                 | false    | Default is `post` if no value is provided   |
+| action          | string                                 | true     | The `action` for the form                   |
+| button          | `Button` [_(ref)_](/components/button) | false    | Configuration object for the form button    |
+| hiddenFormField | object                                 | false    | Configuration object for hidden form fields |
+
+## hiddenFormField
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by rmccar — CHANGES_REQUESTED** [review]
The macro options are going to need to be updated for this. Something like this below with `hiddenFormField` added to the `form` table and then a `hiddenFormField` table added to describe that object.
```
## form

| Name            | Type                                   | Required | Description                                 |
| --------------- | -------------------------------------- | -------- | ------------------------------------------- |
| method          | string                                 | false    | Default is `post` if no value is provided   |
| action          | string                                 | true     | The `action` for the form                   |
| button          | `Button` [_(ref)_](/components/button) | false    | Configuration object for the form button    |
| hiddenFormField | object                                 | false    | Configuration object for hidden form fields |

## hiddenFormField

| Name  | Type   | Required | Description        |
| ----- | ------ | -------- | ------------------ |
| name  | string | false    | Hidden field name  |
| value | string | false    | Hidden field value |
```

boxadesign (approved):

rmccar (approved):

Issue #1461, opened 2021-03-31
Don't highlight validation error on write-in input unless input is mandatory and radio/checkbox is selected


jrbarnes9:
Don't highlight validation error on write-in input unless input is mandatory and radio/checkbox is selected

The below example write-in input field shouldn't have a red border because it's radio hasn't been selected.  Only if the radio is selected, and the write-in input is mandatory, should the input border be highlighted.  

Screenshot of [pattern example prototype](https://ons-design-system.netlify.app/patterns/error-validation/examples/errors-proto/errors/?other-answer=&day=&month=&year=&btn-submit=)
![Screenshot 2021-03-31 at 08 47 11](https://user-images.githubusercontent.com/43346934/113109611-0c5bde00-91fe-11eb-9726-5d996fad1162.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/30#discussioncomment-551559_

PR #1532, opened 2021-05-26
Fix `required` attribute on inputs

rmccar:
### What is the context of this PR?
Fixes: #1469 

### How to review
Check correct attribute is added on a required input

**BREAKING CHANGE** due to now setting the required attribute using its own param

PR #1536, opened 2021-05-27
Use correct styling on radio other errors

rmccar:
Fixes: #1461 

**Is dependent on #1532 being merged**

### How to review
Check errors prototype is behaving correctly as on trello card


rmccar:
[Issue #1461: Don't highlight validation error on write-in input unless input is mandatory and radio/checkbox is selected](https://trello.com/c/fACJ0DE5/1104-issue-1461-dont-highlight-validation-error-on-write-in-input-unless-input-is-mandatory-and-radio-checkbox-is-selected)

boxadesign (approved):

src/patterns/error-validation/examples/errors-proto/errors/index.njk
```diff
@@ -170,6 +170,7 @@ layout: none
                         "value": "other",
                         "other": {
                             "open": true,
+                            "required": true,
```
jrbarnes9, line 173: The errors proto doesn't current do any kind of form post, but it would be good if we could do that.  So this would work.

jrbarnes9 (approved):

Issue #1462, opened 2021-03-31
Don't use a list for error summary if only one error on page


jrbarnes9:
Don't use a list for error summary if only one error on page

Show just a link...
![Screenshot 2021-03-31 at 09 21 20](https://user-images.githubusercontent.com/43346934/113114036-db31dc80-9202-11eb-89b6-318c61ac51c4.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/30#discussioncomment-551659_

PR #1538, opened 2021-05-27
Don't render list if only one item

rmccar:
Fixes: #1462 

### How to review
- Check lists with more than one item still work in the same way
- Check lists with only one item now just render in `<p>` tags and not a list with no list styling added


rmccar:
[Issue #1462: Don't use a list for error summary if only one error on page](https://trello.com/c/ZO8DcmkX/1105-issue-1462-dont-use-a-list-for-error-summary-if-only-one-error-on-page)

boxadesign (approved):

jrbarnes9 (approved):

Issue #739, opened 2020-03-13
Button focus state shouldn't change when the element is hovered.


jrbarnes9:
Focus state shouldn't change when the element is hovered.

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/issues/45#issuecomment-598634777_

PR #1541, opened 2021-05-28
Fix focus hover state on buttons

rmccar:
Fixes: #739

### How to review
Check buttons now are the correct colour (dark yellow) when focused and hovered


rmccar:
[Issue #739: Button Focus state shouldn't change when the element is hovered.](https://trello.com/c/07WQ8tRL/489-issue-739-button-focus-state-shouldnt-change-when-the-element-is-hovered)

boxadesign (approved):

jrbarnes9 (approved):

PR #1542, opened 2021-06-07
Improve printout when spinning up local DS

rmccar:
Fixes an issue with the print out when spinning up the DS locally
image

Will move the first line of this output onto new line
image

Also removes an unused import (postcss-mq-optimize) and removes the package itself because its not used anywhere else in the DS

### How to review
See that on master the top sceenshot is output and on this branch the bottom one is output when spinning up the DS locally by running `yarn start`


rmccar:
[Improve local DS output](https://trello.com/c/W0kLM4vq/1192-improve-local-ds-output)

boxadesign:
We could use this branch to clean up other logging that the build spits out. For instance, I don't think we need all the svg's etc output, could be simplied to statements like "Generating images", "Building pages" etc

rmccar:
> We could use this branch to clean up other logging that the build spits out. For instance, I don't think we need all the svg's etc output, could be simplied to statements like "Generating images", "Building pages" etc

Yeah good idea 👍 . Ill have a look at that

jrbarnes9 (approved):

boxadesign (approved):

Issue #764, opened 2020-06-03
Search component: Few problems with this component whilst working on the 'News and events' design.


trevorsaint:
Few problems with this component whilst working on the 'News and events' design.

The default usage of the macro results in:

1.  **Empty label attribute** (should be populated with a default label title even if it is hidden using a utility class)

2. **Classes modifier** (should allow flexibility in the width of the input if required. A class is added to the input, however, there is no space between class additions, which causes a bug

We also need to look at the ordering of the classes. The class additions seem to come before the already defined width attribute and therefore would have no effect on the width anyway

3. **Responsive width** (should scale nicely and be full width on mobile. However due to the `flex-flow: row wrap` when the screen is reduced the component breaks. Removing this attribute fixes the issue

4. **Unused classes** (classes have been added that have no effect, and should really be removed)

5. **Enhancements** (adding the option for attributes would be a good improvement)

_Originally posted by @trevorsaint in https://github.com/ONSdigital/design-system/issues/656#issuecomment-638140981_

PR #1543, opened 2021-06-07
Search component fixes

rmccar:
Fixes: #764

- Fixes issue with component not having the correct padding at screen width 500px when using the width constrained classes, removes some whitespace from the code and fixes some blank space output into the html around the classes on the input tag

- Fixes and issue with a red double error border when an error object is passed in, the input is focused and the character limit is reached

- Makes char check examples fit the amount of chars allowed to be entered

### How to review
- No blank space around classes on input and button
- Check search component looks correct at 500px screen width
- Check search component looks correct at all widths while adding `input--w-<number>` class. Testing numbers above and below 20
- Check other issues on card are fixed
- Pass an error object into the char check example and enter some test to go passed the char limit, then focus the input to make sure it behaves as other inputs when they have an error and has a black and yellow border


rmccar:
[Issue #764: Search component: Few problems with this component whilst working on the 'News and events' design.](https://trello.com/c/4H1dfQNM/511-issue-764-search-component-few-problems-with-this-component-whilst-working-on-the-news-and-events-design)

trevorsaint (changes requested):
Hi @rmccar I’ve spotted a few issues with this:

**First example:**

We have no label for the first example. If we don’t want a label visually then we should still have a label regardless, with a visually hidden class. This would ensure accessibility and W3C validation passes.

**Second example:**

We seem to have a duplicate placeholder

duplicate_placeholder

Actually after closer inspection, we should not be using the placeholder attribute anyway so this needs to be removed! We are placing the label over the input which is good. The same as how GDS are doing it.

**Other things:**

Can we remove the gaps between classes?

class_gaps

Can we apply a '**search**' type? The same as how GDS have done it.

Other than this all seems to be fine.

rmccar:
> Hi @rmccar I’ve spotted a few issues with this:
> 
> **First example:**
> 
> We have no label for the first example. If we don’t want a label visually then we should still have a label regardless, with a visually hidden class. This would ensure accessibility and W3C validation passes.
> 
> **Second example:**
> 
> We seem to have a duplicate placeholder
> 
> duplicate_placeholder
> 
> Actually after closer inspection, we should not be using the placeholder attribute anyway so this needs to be removed! We are placing the label over the input which is good. The same as how GDS are doing it.
> 
> **Other things:**
> 
> Can we remove the gaps between classes?
> 
> class_gaps
> 
> Can we apply a '**search**' type? The same as how GDS have done it.
> 
> Other than this all seems to be fine.

- Label added for first example
- Duplicate placeholder fixed but we have decided to continue using placeholder because we only use it in search and wont use it where we ask for details
- The gaps between classes will be covered in another PR because it is something that needs to be looked at across the DS
- I have also added the search type to inputs with a search button

trevorsaint (changes requested):
Hi @rmccar just a small thing

Can we change the label for the first example. Filter results is not really suitable for the search component. Something along the lines of 'Search' would be best.

I know we have a specific example for filter results but the button relates to that operation.

**An example of what we did at HMCTS**

search-example

trevorsaint (approved):

boxadesign (approved):

PR #1552, opened 2021-06-14
Document form component validation error messages

jrbarnes9:
Updated the following components and patterns to include error validation documentation and examples:

#### Components
- Access codes
    - Also renamed 'uac' pattern folders to 'access-code' *** **BREAKING CHANGE** ***
- Checkboxes
    - Also updated all examples to be more realistic use cases and use `legendIsPageTitle`
- Dates
- Durations
- Input
    - Also changed prefix/suffix background colour to match secondary buttons
    - Also updated width-constrained examples to use correct number classes
- Mutually exclusive
- Radios
- Select
- Summary
- Textarea
- Upload

#### Patterns
- Dates
- Download resources
- Durations
    - Also removed unnecessary time example
- Email addresses
- Feedback
- Find a support centre
- Numeric values
- Send a web form

Also fixed any typos in the content, and adjusted heading levels for consistency across similar templates.
 
### How to review
- Check all examples and error documentation for all the patterns and components listed above. 
- Check there are no issues with access codes after the folder name change


jrbarnes9:
[Issue #1569: Upload component doesn't call `onsField` or `onsFieldset` so in turn, doesn't call `onsError`, so the validation pattern can't be used with it.](https://trello.com/c/dK0TQeYt/1220-issue-1569-upload-component-doesnt-call-onsfield-or-onsfieldset-so-in-turn-doesnt-call-onserror-so-the-validation-pattern-cant-b)

LauraC119 (approved):
all good from my point of view

src/components/search/index.njk
```diff
@@ -23,9 +23,9 @@ Provide `accessiblePlaceholder: true` and a `label` to create an accessible plac
 
 A [width constraint](/components/input#width-constrained) class can be used to make the width of input smaller. If the search context is known, for instance a sequence of characters of a known size, the width constraint can help users.
 
-### Variants
+###
```
rmccar, line 26: Missing title

src/components/summary/examples/summary-grouped-validation/index.njk
```diff
@@ -29,7 +29,7 @@
                             },
                             {
                                 "rowTitle": "What was the value of the business's total sales of food?",
-                                "errorMessage": "Change one or more of the figures so they total the sum of £600",
+                                "errorMessage": "Change one or more of the figures so they sum to £600",
```
rmccar, line 32: Text needs changing, not sure what its meant to say

src/patterns/feedback/examples/feedback-form-errors/index.njk
```diff
@@ -81,6 +81,9 @@ layout: none
                             "description": "For example, questions not clear, answer options not relevant"
                         },
                         "value": "questions",
+                        "attributes": {
+                            "checked": true
+                        },
```
rmccar, line 86:
```suggestion
                        "checked": true,
```
Attributes doesn't need to be used here

rmccar (commented):
Ive noticed that all the error examples have the pattern error example *then* sample error messages, except for access codes. Maybe we should reverse this on access codes so it fits in with the rest. Also there is no error examples for the experimental patterns ("Find a support centre" and "Download resources")

src/components/summary/examples/summary-grouped-validation/index.njk
jrbarnes9: @LauraC119?

jrbarnes9:
> Ive noticed that all the error examples have the pattern error example _then_ sample error messages, except for access codes. Maybe we should reverse this on access codes so it fits in with the rest. Also there is no error examples for the experimental patterns ("Find a support centre" and "Download resources")

I think it's better to show the example first, but some examples where it's a full page, I think are too long, so you'd miss the content.  This is why I flipped the order of these examples.

rmccar:
> > Ive noticed that all the error examples have the pattern error example _then_ sample error messages, except for access codes. Maybe we should reverse this on access codes so it fits in with the rest. Also there is no error examples for the experimental patterns ("Find a support centre" and "Download resources")
> 
> I think it's better to show the example first, but some examples where it's a full page, I think are too long, so you'd miss the content. This is why I flipped the order of these examples.

Do we still not want examples for "Find a support centre" and "Download resources" though?

jrbarnes9:
> > > Ive noticed that all the error examples have the pattern error example _then_ sample error messages, except for access codes. Maybe we should reverse this on access codes so it fits in with the rest. Also there is no error examples for the experimental patterns ("Find a support centre" and "Download resources")
> > 
> > 
> > I think it's better to show the example first, but some examples where it's a full page, I think are too long, so you'd miss the content. This is why I flipped the order of these examples.
> 
> Do we still not want examples for "Find a support centre" and "Download resources" though?

Yeah I'll add those now.

src/components/summary/examples/summary-grouped-validation/index.njk
LauraC119: should be "sum to £600". This has made me notice, though, that while the visual example is a sum of pounds the content on the page is a percentage. @jrbarnes9 do we normally match the example to the page content?

jrbarnes9: Yeah usually.  I'll update this to match.

LauraC119 (approved):

boxadesign (approved):

Issue #1561, opened 2021-06-18
Add error validation border to select when no option selected


jrbarnes9:
Add error validation border to select when no option selected

Currently:
![Screenshot 2021-06-18 at 09 19 08](https://user-images.githubusercontent.com/43346934/122530652-9efa6500-d016-11eb-9cc8-69cb6a1ba1f8.png)

Should be:
![Screenshot 2021-06-18 at 09 19 58](https://user-images.githubusercontent.com/43346934/122530675-a588dc80-d016-11eb-9890-4220714cf348.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/56#discussioncomment-886812_

PR #1571, opened 2021-06-22
Fix error style on selects in radios

rmccar:
Fixes: #1561
Also fixes whitespace in select class list when classes added with `params.classes`

### How to review
Check errors prototype and select the pepperoni radio to see the select has a red border

### DIFF
--- a/src/components/select/_macro.njk
+++ b/src/components/select/_macro.njk
@@ -21,7 +21,7 @@
         <select
             id="{{ params.id }}"
             name="{{ params.name }}"
-            class="input input--select {{ params.classes }}"
+            class="input input--select{% if params.classes %} {{ params.classes }}{% endif %}{% if params.select.error %} input--error{% endif %}"
             {% if params.value is defined and params.value %}value="{{ params.value}}" {% endif %}
             {% if params.attributes is defined and params.attributes %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value is defined and value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
         >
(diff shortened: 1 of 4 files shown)


rmccar:
[Issue #1561: Add error validation border to select when no option selected](https://trello.com/c/NqKH1Rde/1214-issue-1561-add-error-validation-border-to-select-when-no-option-selected)

src/components/select/_macro.njk
```diff
@@ -21,7 +21,7 @@
         <select
             id="{{ params.id }}"
             name="{{ params.name }}"
-            class="input input--select {{ params.classes }}"
+            class="input input--select{% if params.classes %} {{ params.classes }}{% endif %}{% if params.select.error %} input--error{% endif %}"
```
jrbarnes9, line 24: Does it need the `select`?
```suggestion
            class="input input--select{% if params.classes %} {{ params.classes }}{% endif %}{% if params.error %} input--error{% endif %}"
```

src/patterns/error-validation/examples/errors-proto/errors/index.njk
```diff
@@ -299,6 +304,72 @@ layout: none
...
+                                {
+                                    "value": "extra hot",
+                                    "text": "Extra Hot"
+                                }
+                            ]
+                        }
+                    }
+                ]
+            })
+        }}
```
jrbarnes9, line 372: Need to add this to the initial page of the prototype too, with the 'pepperoni' option selected

jrbarnes9 (changes requested):

src/components/select/_macro.njk
rmccar: Yeah I needed a different param because setting `params.error` would add all the panel styling etc and we would end up with the double error panel situation again. It was the best solution I could think of because it will still allow the correct error styling within radios but also if the select is used on its own.

jrbarnes9 (approved):

boxadesign (approved):

PR #1583, opened 2021-07-06
Fix error when headMeta missing from pageConfig

rmccar:
Error in Runner when headMeta object not supplied. Now checking for it when looking for headMeta.description in pageConfig

### How to review
Tests pass


rmccar:
[Error in runner when headMeta missing from pageConfig](https://trello.com/c/gPHYeVnN/1240-error-in-runner-when-headmeta-missing-from-pageconfig)

PR #1584, opened 2021-07-07
Convert footer language items to list

rmccar:
There is an error in Runner when using footers that has come from the changes to the list component in #1538. Because we now calculate the length of the list and Runner passes in a generator object into the languages list in footers, this is to convert it back to a list before the list component is called. This is because generator objects aren't compatible with `| length`

### How to review
- Check that tests pass
- Check that lists and footers still work correctly


rmccar:
[Convert footer language lists so solve error in runner](https://trello.com/c/hXcFy5N8/1241-convert-footer-language-lists-so-solve-error-in-runner)

jrbarnes9 (approved):

boxadesign (approved):

PR #1587, opened 2021-07-12
Fix jinja bug with selects

rmccar:
Bug when using jinja where it is erroring when looking for `params.select.error` if `params.select` doesn't exist. So Ive added an if to stop this.

### How to review
- Selects still work as expected
- Tests pass


rmccar:
[Fix select jinja bug](https://trello.com/c/ikIAbulO/1243-fix-select-jinja-bug)

boxadesign (approved):

jrbarnes9 (approved):

PR #1588, opened 2021-07-12
Documentation templates

jrbarnes9:
To ensure content consistency in the documentation of components and patterns, I've added template folders at the top level of both components and patterns.  The templates include guidance and required/optional headings to make it easier to document new components and patterns that are all aligned.

### How to review
Check the templates.


jrbarnes9:
[Define common content features for patterns/components/styles](https://trello.com/c/TR8srk0D/57-define-common-content-features-for-patterns-components-styles)

LauraC119 (changes requested):
**Replace with these paragraphs (on component and pattern template):**

Add a short summary here to explain what this component is and what it does, for example, “Accordions let users show or hide sections of related content on the same page."

Add guidance here on when this component should be used and the common problems it can solve

Add warnings here about how the component could be misused, and guidance on what to use instead, for example, use radios not checkboxes

Add guidance here on how to apply error validation if it is needed

LauraC119 (approved):

boxadesign (approved):

PR #1591, opened 2021-07-14
Add ids to question and use to select in relationships

rmccar:
Add an ID to questions and fieldsets and use the IDs to select question in relationships. This is to fix an issue in runner with we have when selecting this with a class as after the legend as a title changes the correct class is no longer being set at the relationship level.

### How to review
Relationships still work as expected

### DIFF
--- a/src/components/date-input/_macro.njk
+++ b/src/components/date-input/_macro.njk
@@ -4,7 +4,6 @@
     {% from "components/input/_macro.njk" import onsInput %}
     {% from "components/error/_macro.njk" import onsError %}
 
-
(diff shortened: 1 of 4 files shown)


rmccar:
[Change how question is selected in relationships](https://trello.com/c/90PoWOio/1247-change-how-question-is-selected-in-relationships)

jrbarnes9 (commented):
Wouldn't it be better to make these params with a default?

rmccar:
> Wouldn't it be better to make these params with a default?

Yeah that would be a good idea normally, but in this case if those ids are set to something else when using relationships it would break the js

boxadesign (approved):

jrbarnes9 (approved):

PR #1610, opened 2021-07-29
Autosuggest - change boolean type for allowMultiple property

boxadesign:

The `allowMultiple` property on the autosuggest component required the boolean value to be a string. This is inconsistent and created problems in other services. 

This PR changes the macro to allow the value to be `true` or `false`.

### How to review
Test that the multiple selection example works as expected.


jrbarnes9 (approved):

rmccar (approved):

ajmaddaford (approved):
Works as expected in eq-questionnaire-runner 👍

PR #1615, opened 2021-07-30
Fix page template header example

rmccar:
Example on the page template page was using the wrong param name and was missing current page styling.

### How to review
Example now looks and works as it should


rmccar:
[Page template example using wrong param](https://trello.com/c/zvSB6meq/1266-page-template-example-using-wrong-param)

jrbarnes9 (dismissed):

boxadesign (approved):

jrbarnes9 (approved):

PR #1627, opened 2021-08-10
Design System - Updated docs for using scripts via npm

boxadesign:
Updated documentation for using JS.

Found here -> /get-started/css-js-assets/#javascript

### How to review
Check the docs.


LauraC119 (dismissed):
1. Header – add the acronym - "JavaScript (JS)" - and then use it after that as we use the acronym in following section on CDN

2. After the header, change every "Javascript" to "JS"

3. "- add JavaScript file to your HTML - import JavaScript using a bundler" – I assume these will be bullet points when formatted correctly?

4. Webpack link not found

5. Change
"Import the node_modules/design-system/js/main.js which will import JavaScript for all components."
to
"Import node_modules/design-system/js/main.js to import the JS for all components."

6. Change 
"To import the JavaScript for a specific component, for instance the button component. You can import node_modules/design-system/components/button/button.dom.js." 
to
"You can import the JS for a specific component. For example, node_modules/design-system/components/button/button.dom.js imports the JS for the button component."

7. Change
"to import which will handle initialising a component."
to
"to import. This file will handle the initialising of a component."

src/pages/get-started/css-js-assets/index.njk
```diff
@@ -50,18 +50,35 @@ Some components use other components, for example, the accordion component uses
...
+
+### Import JS using a bundler
+
+If your service uses a bundler like [Webpack](https://webpack.js.org/) you can select and initialise all component JS using `import` or `require()`. 
+
+Import the `node_modules/design-system/js/main.js` to import the JS for all components. 
+
+#### Import JS for a specific component
+
+You can import the JS for a specific component. For example, `node_modules/design-system/components/button/button.dom.js` imports the JS for the button component. Each component that uses JS will have a `COMPONENT.dom.js` file available to import. This file will handle the initialising of a component.
```
jrbarnes9, line 81:
```suggestion
You can import the JS for a specific component. For example, `node_modules/design-system/components/button/button.dom.js` imports the JS for the button component. Each component that uses JS will have a `<COMPONENT>.dom.js` file available to import. This file will handle the initialising of a component.
```

jrbarnes9 (changes requested):
Code blocks should be inline to better deal with wrapping 
Screenshot_20210909-130744

rmccar (changes requested):
Javascript navigation link is broken. ID of the Javascript h2 element is "javascript-js" and the link is looking for "#javascript"

jrbarnes9 (approved):

rmccar (approved):

PR #1630, opened 2021-08-11
Add search to new build system

boxadesign:
Add basic DS search to new build system.


jrbarnes9:
[Search the design system](https://trello.com/c/gsFFJQCJ/66-search-the-design-system)

jrbarnes9:
I don't think we need the 1px inset box-shadow on `.input:focus` for the search input.  It doesn't need it because it's already changing background colour on focus.

jrbarnes9:
I think the search icon svg might need tweaking so the icon and label appear centred to the button boundary.  At the moment there's more space to the left of the icon than there is the right of the label.
Screenshot 2021-08-25 at 16 32 22

jrbarnes9:
Is there any way we can add some sort of weighting to the title, so it ranks above the synonyms? 

Here for example, you'd expect search to be returned above autosuggest...
Screenshot 2021-08-31 at 12 02 17

rmccar:
image
When the screen width goes down to 309px we get this issue

boxadesign:
> image
> When the screen width goes down to 309px we get this issue

We don't suport anything below 320px

src/components/autosuggest/code.list.searcher.js
```diff
@@ -6,7 +6,16 @@ export default function queryJson(query, data, searchFields) {
     threshold: 0.2,
     location: 1,
     distance: 1000,
-    keys: [searchFields, 'tags'],
+    keys: [
+      {
+        name: searchFields,
+        weight: 0.2,
```
rmccar, line 12: image
Found another place where weighting needs tweaking. Not tested fully but increasing the weight to 0.3 here solves this issue

boxadesign: I'll take a look. I thought that just having a greater weighting on one string query would always weight it higher 🤷🏼‍♂️

jrbarnes9 (changes requested):
The footer overlaps the search results on short pages
Screenshot 2021-09-06 at 14 27 51

src/components/autosuggest/code.list.searcher.js
rmccar: Yeah you would think that! It did fix the issue Jon found. I just made some quick changes and that fixed it. I can't quite remember exactly how the weighting works in Fuse.

rmccar (dismissed):
Other than Jons footer issue, looks good to me

jrbarnes9 (approved):

rmccar (dismissed):

LauraC119 (commented):
Couple of synonyms it might be worth adding for the whole getting started area. I tried "Start" and "Set up" but neither had results. 

Also, not sure why the pages in Contribute and Getting started are categorised as "Documentation" in the search results. Shouldn't they be the same as components or patterns, which are categorised as "Components" or "Patterns"? Could confuse as there is no "Documentation" area.

kruncher (approved):

jrbarnes9:
The yellow focus state border remains on after focus is moved from search input after something entered
Screenshot 2021-09-07 at 09 34 31

rmccar (approved):

jrbarnes9 (approved):

Issue #1628, opened 2021-08-10
Add option to radios to remove borders


jrbarnes9:
New variant for radios without a border, as per [DP's design system](https://dp-design-system.netlify.app/components/radios/). This variant is to be used on census web pages and other new pages that have a radio component.

![Screenshot 2021-08-05 at 10 09 55](https://user-images.githubusercontent.com/19624419/128830359-7d23a47b-7d1e-476b-9656-6a8060c2a567.png)

_Originally posted by @mr-nick17 in https://github.com/ONSdigital/design-system/discussions/56#discussioncomment-1152913_

PR #1652, opened 2021-08-20
Borderless radio/checkboxes variant

mr-nick17:
Adding borderless radio and checkbox variants to support census dataset landing pages

### How to review
- Sense check
- Spelling/grammar 
- Check design matches `dp-design-system` [borderless radio variant](https://dp-design-system.netlify.app/components/radios/)

### DIFF
--- a/src/components/radios/_macro.njk
+++ b/src/components/radios/_macro.njk
@@ -18,8 +18,8 @@
                 {% if params.or is defined and params.or and loop.revindex0 == 0 %}
                     <span class="radios__label u-mt-s u-fs-r--b" aria-hidden="true">{{ params.or }} </span>
                 {% endif %}
-                <p class="radios__item{{ " " + params.containerClasses if params.containerClasses else "" }}">
-                    <span class="radio{{ " " + params.blockClasses if params.blockClasses else "" }}">
+                <p class="radios__item{{ " u-mb-s" if params.borderless else "" }}">
+                    <span class="radio{{ " radio--no-border" if params.borderless else "" }}">
                         <input
                             type="radio"
                             id="{{ radio.id }}"
--- a/src/components/radios/_macro.njk
+++ b/src/components/radios/_macro.njk
@@ -13,14 +13,15 @@
         "legendIsQuestionTitle": params.legendIsQuestionTitle,
         "error": params.error
     }) %}
-        <div class="radios__items">
-            {% for radio in params.radios %}
-                {% if params.or is defined and params.or and loop.revindex0 == 0 %}
-                    <span class="radios__label u-mt-s u-fs-r--b" aria-hidden="true">{{ params.or }} </span>
-                {% endif %}
-                <p class="radios__item{{ " u-mb-s" if params.borderless else "" }}">
-                    <span class="radio{{ " radio--no-border" if params.borderless else "" }}">
-                        <input
+    <div class="radios__items">
+        {% for radio in params.radios %}
+            {% if params.or is defined and params.or and loop.revindex0 == 0 %}
+                <span class="radios__label u-mt-s u-fs-r--b" aria-hidden="true">{{ params.or }}
+                </span>
+            {% endif %}
+            <p class="radios__item{{ " radios__item--no-border" if params.borderless }}">
+                <span class="radio{{ " radio--no-border" if params.borderless }}">
+                    <input
                             type="radio"
                             id="{{ radio.id }}"
                             class="radio__input js-radio{{ " " + radio.classes if radio.classes else "" }}"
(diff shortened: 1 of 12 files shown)


armstrongb:
Check if this resolved issue: #1628 ?

src/components/radios/examples/radios-borderless/index.njk
```diff
@@ -0,0 +1,40 @@
...
+    "title": "What type of accommodation do you live in?",
+    "classes": "u-mt-no",
+    "legendIsQuestionTitle": true
+}) %}
+{{
+        onsRadios({
+            "name": "property-type",
+            "dontWrap": true,
+            "containerClasses": "u-mb-s",
+            "blockClasses": "radio--no-border",
```
boxadesign, line 14: I would change this to a single parameter `borderless: true`. The macro would contain the logic to handle which  classes to add i.e. `{% if borderless === true %} u-mb-s ...` etc

jrbarnes9 (changes requested):

src/components/radios/examples/radios-borderless/index.njk
mr-nick17: Interestingly, this is how I initially developed the feature but wasn't too sure on how much control the user would ordinarily have. I will refactor to use a bool.

src/components/radios/index.njk
```diff
@@ -69,6 +69,14 @@ The submit button should use the `hidden` attribute and also the `u-d-no` utilit
     patternlibExample({"path": "components/radios/examples/radios-clear/index.njk"})
 }}
 
+### Radios without a border
+This variant removes the border surround the radio input and [label](/components/label).
+
+Add the block css class `radio--no-border` and the utility css class `u-mb-s` to the container to ensure adequate spacing between the radio items.
```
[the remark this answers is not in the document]
mr-nick17: As above, will refactor away from the user adding the utility class

boxadesign (dismissed):

jrbarnes9:
I think we should apply this variant to checkboxes too.  We have a good example of using a borderless variant already in the downloadable resources pattern, for the filters: https://ons-design-system.netlify.app/patterns/download-resources/

The purpose is the same.  To avoid the visual clutter created by the borders when they're not the primary purpose of the page.

src/components/radios/index.njk
```diff
@@ -69,6 +69,12 @@ The submit button should use the `hidden` attribute and also the `u-d-no` utilit
     patternlibExample({"path": "components/radios/examples/radios-clear/index.njk"})
 }}
 
+### Radios without a border
+This variant removes the border surrounding the radio input and [label](/components/label).
```
jrbarnes9, line 73: We should add usage guidance (@LauraC119  will provide). And the example should reflect this so it's used for the correct purpose.  We wouldn't want this used for single page question answers, so I would change the example by removing the use of `onsQuestion` and match the options up to the 'get the data' design: 

Screenshot 2021-08-24 at 14 59 27

jrbarnes9 (changes requested):

jrbarnes9 (changes requested):
The variant is missing a focus state.

src/components/radios/_macro.njk
```diff
@@ -18,8 +18,8 @@
                 {% if params.or is defined and params.or and loop.revindex0 == 0 %}
                     <span class="radios__label u-mt-s u-fs-r--b" aria-hidden="true">{{ params.or }} </span>
                 {% endif %}
-                <p class="radios__item">
-                    <span class="radio">
+                <p class="radios__item{{ " u-mb-s" if params.borderless else "" }}">
```
jrbarnes9, line 21: Might be better to handle the margins in the css as the bordered variant has the margin on the `:last-child` removed, which should be the same here for this variant.

LauraC119 (changes requested):
Suggest we replace content in the radios without border section with this:
This variant removes the border around the radio input and label. Do not use it for question pages with radios. It is better to have a border. It makes the radios more visible for users and gives them a larger target area to select (the label and the boundary).

Only remove the radio border for pages that are not following the practice of “one thing per page”, where the radios are not the primary call to action. For example, the radios could be one part of a multipurpose page, like the checkboxes used for the filters on the [download resources pattern](https://ons-design-system.netlify.app/patterns/download-resources/).

src/components/radios/index.njk
mr-nick17: I agree that the example should be aligned closer to its intended purpose; the only thing I would say against using the 'get the data' design is that it is a pattern and includes additional components like the primary call to action. Happy to include the download formats and maybe using the 'select download format' as the legend.

jrbarnes9: Yeah I think that will do

jrbarnes9 (changes requested):
This just needs to also be applied to checkboxes, then this is good to go.

jrbarnes9 (dismissed):

rmccar (dismissed):
Only one issue, because the border has been taken away there is a large area to the right of the label on the radio/checkbox that the user is able to click to select the radio/checkbox. Im thinking we should have a variable width for borderless versions and have the width end at the end of the label text. Other than that it looks good

rmccar (approved):

jrbarnes9 (approved):

PR #1702, opened 2021-09-28
Remove public path override

rmccar:
This PR is to remove the public path override stuff that was needed in the old build system because the js files were chunked and the path needed to be updated with the cdn_url so that they could be resolved. But after moving to Gulp we don't need this anymore because they aren't chunked anymore and keeping it in is causing a `main.js:7 Uncaught ReferenceError: __webpack_public_path__ is not defined` error in the console in the Python systems

### How to review
- Tests pass
- DS builds correctly

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -38,12 +38,12 @@ const sassOptions = {
 
 const scripts = [
   {
-    entryPoint: ['./src/js/public-path-override.js', './src/js/main.js'],
+    entryPoint: './src/js/main.js',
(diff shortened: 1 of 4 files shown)


rmccar:
[Remove public path override as it's not needed anymore.](https://trello.com/c/A24MiyGI/1356-remove-public-path-override-as-its-not-needed-anymore)

boxadesign (approved):

jrbarnes9 (approved):

PR #1703, opened 2021-09-28
Adjust sub navigation group ordering and folder structure.

kruncher:

Add ability to sub-navigation helper to order sub navigation groups (We want to be able to order the groups manually.)

AND 

Allow documentation pages to placed in group folders. For example, this would allow all pages in the "Content guides" group (at /guidance/) to be stored in a single folder called `/content-guides/`.  This will help navigate to the files as the 'guidance' section grows.

...it may be more appropriate to remove the `group` frontmatter, and instead allow the folders to be dictate the displayed group subheadings, by converting the folder name into displayable text e.g. `/browser-testing/` > Browser testing... or perhaps the group folder could contain some sort of config file which allows you to set it's order and display title

Ticket: https://trello.com/c/2vW8Ekho/89-sub-navigation-group-ordering-and-folder-structure

### Details
- The `group` frontmatter has been removed in favour of using the directory structure with 'group.yml' definition files instead. The `group` frontmatter is no longer supported.
- If a page template directory contains child directories with 'group.yml' files then those will be treated as groups.
- If a page template directory does not contain child directories with 'group.yml' files; then any child pages will be a part of the default group.
- This functionality was integrated into the page meta data which can be accessed via the site map.

### How to review
1. Test with `yarn start` locally.
2. Test with `yarn build && npx serve build` locally.


boxadesign (approved):
Tested locally, works as expected.

jrbarnes9 (approved):

PR #1735, opened 2021-10-20
Fixed pagination icons

jrbarnes9:
- Content pager next/prev icons were pointing in the wrong direction after the macro was updated to use `onsIcon()`. The `transform` style was removed to correct this.
- Also moved the right-to-left example which uses the content pager, from base page template to the new 'guide' pattern

### How to review
Check pagination component icons and guide examples.

### DIFF
--- a/src/components/content-pagination/_content-pagination.scss
+++ b/src/components/content-pagination/_content-pagination.scss
@@ -34,10 +34,6 @@
 
   &__link-title {
     display: block;
-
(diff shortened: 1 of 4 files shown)


jrbarnes9:
[Add documentation for "right-to-left" guide pattern](https://trello.com/c/WQIbmkuV/1004-add-documentation-for-right-to-left-guide-pattern)

PR #1758, opened 2021-11-09
Correct hub and spoke examples

jrbarnes9:
The hub example in the hub and spoke pattern incorrectly shows an information panel under the heading.  This removes that panel, as well as adding a 'complete' example which includes a warning panel. This also adds two spoke examples which are linked to within the partially complete hub example.

### How to review
Check hub examples and links to spoke examples.

### DIFF
--- a/src/patterns/pages/hub-and-spoke/examples/hub-complete/index.njk
+++ b/src/patterns/pages/hub-and-spoke/examples/hub-complete/index.njk
@@ -55,7 +55,7 @@ layout: none
                                 ]
                             },
                             {
-                                "rowTitle": "Accommodation",
+                                "rowTitle": "Household accommodation",
                                 "rowItems": [
                                     {
                                         "iconType": "check",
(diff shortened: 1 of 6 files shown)


jrbarnes9:
[Remove info panel from hub and spoke example and add submit example](https://trello.com/c/vwA78XT3/1404-remove-info-panel-from-hub-and-spoke-example-and-add-submit-example)

rmccar (commented):
Wouldn't we want to either have the other pages also as examples or create a prototype like the error one? I don't know if people would know to click the submit button to see the other pages because they're only examples

jrbarnes9:
> Wouldn't we want to either have the other pages also as examples or create a prototype like the error one? I don't know if people would know to click the submit button to see the other pages because they're only examples

Yeah ok i'll show the spoke examples too then

rmccar (approved):

boxadesign (approved):

PR #1769, opened 2021-11-17
Fix Jinja compatability issue with checkboxes

boxadesign:
Jinja compatibility Issue with `checkbox.other.selectAllchildren`  inside `checkboxes/_macro.njk`.

The 'other' property wasn't being checked before checking for `selectAllChildren` which jinja doesn't like.

### How to review
Test it works.


jrbarnes9 (approved):

Paul-Joel (approved):
This clears the problem I was having

PR #1791, opened 2021-12-02
Fixed sticky Table of Contents invalid html

jrbarnes9:
Using the sticky variant of `onsTableOfContents()` renders invalid html because the opening `<aside>` tag is wrapped in a condition to set `skipLink`. This is not required for sticky variant because user will only have to pass through the nav only once. 

### How to review
Build the TOC component with sticky variant and check html is valid.


mr-nick17 (approved):
Verified the html is now valid

boxadesign (approved):

PR #1810, opened 2022-01-05
Checkboxes - select parent checkbox if child is selected on page load

boxadesign:
When using nested checkboxes if a child checkbox is checked on page load the parent should be selected by default.

### How to review
Set `checked: true` on a child checkbox (other checkboxes) and confirm the parent is selected and all child checkboxes are displayed.

image

### DIFF
--- a/src/components/checkboxes/_macro.njk
+++ b/src/components/checkboxes/_macro.njk
@@ -31,6 +31,7 @@
         {% endif %}
         <div class="ons-checkboxes__items">
             {% for checkbox in params.checkboxes %}
+                {% set childIsChecked = false %}
                 {% set labelHTML = checkbox.label.text %}
                 {% if params.mutuallyExclusive is defined and params.mutuallyExclusive %}
                     {% set exclusiveClass = ' ons-js-exclusive-group-item' %}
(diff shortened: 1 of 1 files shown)


jrbarnes9 (approved):

rmccar (approved):

rmccar:
Looks good but should we add an example where it is open and showing the child checkboxes without the user having to click one? Just because I think that would make it clearer

boxadesign:
> Looks good but should we add an example where it is open and showing the child checkboxes without the user having to click one? Just because I think that would make it clearer

I did think of doing that but this is quite edge case and not sure it's really a variant, more just functionality that's available and kinda expected.

rmccar:
> > Looks good but should we add an example where it is open and showing the child checkboxes without the user having to click one? Just because I think that would make it clearer
> 
> I did think of doing that but this is quite edge case and not sure it's really a variant, more just functionality that's available and kinda expected.

Yeah you're right its not a variant, only reason why I thought of it was I was looking round for an example with child checkboxes and couldn't find one for a bit so was just thinking of the users. Don't need a new example just add `checked: true` to the one we have. Its just a suggestion

PR #1813, opened 2022-01-11
Fix for th and td classes conditions

jrbarnes9:
This ensures a space is added between classes when using table `thClasses` and `tdClasses` params.
Also moved `thClasses` to it's correct location so it's now set per item in the `th` array.

### How to review
Check classes can correctly be applied to `th` and `td` table elements.


jrbarnes9:
[No space before `thClasses`  and `tdClasses`](https://trello.com/c/u0Cbko4q/1449-no-space-before-thclasses-and-tdclasses)

rmccar (approved):

Issue #1803, opened 2021-12-17
Allow Accordion to be displayed as open by default


IncredibleMrTim:
Currently we are unable to display the Accordion component as open by default on setting the `aria-expanded` attribute and the `ons-collapsable--open` class, the accordion opens and then closes again as the page loads.

```
   <div id="accordion" class="ons-accordion">
  <div id="accordion-1" class="ons-collapsible ons-js-collapsible ons-collapsible--accordion ons-collapsible--open" data-btn-close="Hide" data-group="accordion">
    <div class="ons-collapsible__heading ons-js-collapsible-heading" aria-expanded="true">
      <div class="ons-collapsible__controls">
        <h2 class="ons-collapsible__title">Total retail turnover</h2>
```

boxadesign:
https://github.com/ONSdigital/design-system/pull/1815

PR #1815, opened 2022-01-11
Accordion - create simple variant

boxadesign:
Related to #1803 

Created a simple version of the accordion that doesn't use any button controls and adjusts the layout to work in smaller areas e.g. a side column for filtering.

Also allow the accordion items to be open on page load.

### How to review
Review the "Simple" variant shown on the accordion component page.

### DIFF
--- a/src/components/accordion/index.njk
+++ b/src/components/accordion/index.njk
@@ -67,7 +67,7 @@ Accordions hide content so make sure your section headings are clear so the user
 
 Under certain conditions a simplified accordion may be appropriate. The example below shows the accordion being used in a narrow column for filtering results. The button controls and borders have been removed to allows the accordion to fit into a smaller area. 
 
-To use this variant include the parameter `'variant': 'simple'` in your implementation.
+To use this variant include the parameter `'variants': 'simple'` in your implementation.
 
 The example shown uses the parameter `'open': true` which makes each accordion item open on page load.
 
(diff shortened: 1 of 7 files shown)


src/components/accordion/index.njk
```diff
@@ -59,6 +60,26 @@ The accordion uses JavaScript so when it's turned off, the sections of content a
 ### Use clear labels
 Accordions hide content so make sure your section headings are clear so the user does not need to work hard to understand which heading to choose.
 
+
+## Variants
+
+### Simple
+
+Under certain conditions a simplified accordion may be appropriate. The example below shows the accordion being used in a narrow column for filtering results. The button controls have been removed to allows the accordion to fit into a smaller area.
```
jrbarnes9, line 68:
```suggestion
Under certain conditions a simplified accordion may be appropriate. The example below shows the accordion being used in a narrow column for filtering results. The button controls have been removed to allow the accordion to fit into a smaller area. 
```

jrbarnes9 (commented):
Is it possible to set only one of the accordions to be open on page load (rather than all)? I could imagine this being beneficial in some cases where you want to show what's in the first accordion.

boxadesign:
> Is it possible to set only one of the accordions to be open on page load (rather than all)? I could imagine this being beneficial in some cases where you want to show what's in the first accordion.

We don't have any use case for that and I think we should limit this type of behaviour so not to encourage it where it shouldn't really be used.

rmccar (approved):

jrbarnes9 (approved):

jrbarnes9:
[Allow Accordion to be displayed as open by default](https://trello.com/c/xdW5hXum/1441-allow-accordion-to-be-displayed-as-open-by-default)

PR #1816, opened 2022-01-11
Collapsible semantic headings default and focus style issues

jrbarnes9:
- Changed default tag from `div` to `h2` so the collapsible title is always a heading for a semantic structure.
  - This also fixes an alignment issue for when it was wrapped in a div.
  - Also removed  h2 `titleTag` settings from any examples as these are now not required
- Fixed focus styles so chevron is always black and underline removed when collapsible title is in focus.

### How to review
- Check collapsibles and accordions to ensure titles are heading tags and can still be overridden.
- Check focus/hover styles of collapsibles.


jrbarnes9:
[Collapsible issues](https://trello.com/c/5ava75vV/1448-collapsible-issues)

rmccar:
Really small thing but the closed chevron looks like it needs to be moved to the right slightly when focused on
image

rmccar (approved):

boxadesign (approved):

PR #1819, opened 2022-01-13
Add ability to set row ids on summaries

rmccar:
This adds the ability to set ids on summary rows. This is to allow scrolling to specific rows in runner.

I have also fixed a typo

### How to review
Check ids are set correctly

### DIFF
--- a/src/components/summary/_macro.njk
+++ b/src/components/summary/_macro.njk
@@ -22,7 +22,7 @@
                     {% endif %}
                     {% if group.headers is defined and group.headers and group.rows is defined and group.rows %}
                         <table class="ons-summary__items">
-                            <thead {% if group.groupHeaderId is defined and group.groupHeaderId %}id="{{ group.groupHeaderId }}" {% endif %}class="ons-u-vh">
+                            <thead {% if group.id is defined and group.id %}id="{{ group.id }}" {% endif %}class="ons-u-vh">
                                 <tr>
                                     {% for header in group.headers %}
                                         <th>{{ header }}</th>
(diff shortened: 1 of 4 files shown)


rmccar:
[Added ability to set ids on summary rows](https://trello.com/c/dh1JHqBi/1454-added-ability-to-set-ids-on-summary-rows)

rmccar:
[Introduce anchor link to summary page tables (2 SP)](https://trello.com/c/aNrekImJ/5064-introduce-anchor-link-to-summary-page-tables-2-sp)

jrbarnes9 (dismissed):

MebinAbraham (changes requested):
Awaiting confirmation.

src/components/summary/_macro.njk
```diff
@@ -35,7 +35,7 @@
                                 {% if row.error is defined and row.error %} {% set itemClass = " ons-summary__item--error" %}{% endif %}
                                 {% if row.total is defined and row.total %} {% set itemClass = itemClass + " ons-summary__item--total" %}{% endif %}
 
-                                <tbody class="ons-summary__item{{ itemClass }}">
+                                <tbody {% if row.rowId is defined and row.rowId %}id="{{ row.rowId }}" {% endif %}class="ons-summary__item{{ itemClass }}">
```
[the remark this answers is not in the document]
rmccar: As discussed eQ needs the ability to set ids on the `<tr>` and `<tbody>` in each row so they have been added

src/components/summary/_macro.njk
```diff
@@ -35,15 +35,15 @@
...
                                     {% if row.errorMessage is defined and row.errorMessage or (row.rowItems | length > 1 and row.rowTitle) %}
                                         <tr class="ons-summary__row">
                                             <th colspan="3" class="ons-summary__row-title ons-u-fs-r">{{ row.errorMessage or row.rowTitle }}</th>
                                         </tr>
                                     {% endif %}
 
                                     {% for rowItem in row.rowItems %}
-                                        <tr class="ons-summary__row{{ " ons-summary__row--has-values" if rowItem.valueList else "" }}">
+                                        <tr {% if row.rowItemId is defined and row.rowItemId %}id="{{ row.rowItemId }}" {% endif %}class="ons-summary__row{{ " ons-summary__row--has-values" if rowItem.valueList else "" }}">
```
MebinAbraham, line 46: As below, this should come from the row item itself, so like `rowItem.rowItemId`. Also given these are always in the context of what they are used in, do we need to prefix each thing?.

Maybe it should be:
1. `group.groupHeaderId` > `group.id`
2. `row.rowId` > `row.id`
3. `rowItem.rowItemId` > `rowItem.id`

Thoughts?

rmccar: Yeah I can see that, I don't really mind either way. I just thought it was more explicit with all the different ids that are now able to be set and there are already other params that have similar prefixes. @jrbarnes9 @boxadesign what do you think?

boxadesign: I think it is better to just use the name `id`. It's provided inside it's context so shouldn't lead to any confusion. Also, if the parent name is changed it saves needing to pick through any other params that also contain the parent name as a prefix

rmccar: Yeah good point ill change it

MebinAbraham (approved):

boxadesign (approved):

PR #1821, opened 2022-01-17
Timeout modal - fix status code check and multiple fetches

boxadesign:
The timeout modal has been implemented in EQ and has raised a couple of issues:

- Setting of a new expiry time happening multiple times for a single action
- Need to specifically test for a status code of 401 not 200

### How to review
This needs to be tested in EQ as it requires connection to an endpoint.


MebinAbraham (approved):
Described issues no longer present within eQ.

jrbarnes9 (approved):

rmccar (approved):

PR #1871, opened 2022-01-26
Allow removal of the download attribute from download buttons

rmccar:
PR to allow the `download` attribute to be removed from download buttons

### How to review
- Check download buttons work correctly with and without the `download` param
- New docs makes sense


rmccar:
[Save answers as PDF broken when submission has expired](https://trello.com/c/USVNTQ2j/5078-save-answers-as-pdf-broken-when-submission-has-expired)

jrbarnes9 (approved):

boxadesign (approved):

PR #1989, opened 2022-02-15
Timeout panel component

rmccar:
Adds the countdown panel variant.

Also: 
- removes bug where attributes are being added to panels in two places
- removes some whitespace and solves some indentation issues
- refactors timeout.js to allow it to be used by the countdown panel and the countdown modal
- updates docs

### How to review
- Check new tests cover all functionality
- Check tests pass
- Check new panel variant works as expected


src/components/timeout-panel/timeout-panel.js
```diff
@@ -0,0 +1,91 @@
...
+
+    // Settings
+    this.timers = [];
+    this.timerRunOnce = false;
+
+    // Start module
+    this.initialise();
+  }
+
+  async initialise() {
```
kruncher, line 28: Why are `initialise` and `startUiCountdown` `async` functions? the `await` on `runTimer` doesn't seem like it does anything because `runTimer` doesn't await on anything (nor return a custom Promise).

boxadesign: @kruncher @rmccar Alot of this code has been lifted from another module which does similar things and those same functions require waiting for api responses. I think its probably worth a refactor here of the timeout-modal to put any reuseable code into its own module e.g. the ui countdown part so it can be imported into that and this module.

boxadesign: I think it's also worth looking at the above comment I made about creating a separate module for the countdown that both the modal and panel can import.

boxadesign: Sorry I misread that you said you would refactor.

boxadesign:
@rmccar With no connections being made to the back end, if the user refreshes the page the timer restarts. Just checking this is expected behaviour? I would've thought the timeout would need to sync or at least the initial timeout time on the first load of the page would need to be stored (session storage perhaps).

boxadesign:
> > @rmccar With no connections being made to the back end, if the user refreshes the page the timer restarts. Just checking this is expected behaviour? I would've thought the timeout would need to sync or at least the initial timeout time on the first load of the page would need to be stored (session storage perhaps).
> 
> Yeah that's a good point. Maybe I'll need to reuse more of the modal code

There seems like there's alot of cross over here just the implementation differs. It may be worth us pairing up on this.

rmccar:
> > > @rmccar With no connections being made to the back end, if the user refreshes the page the timer restarts. Just checking this is expected behaviour? I would've thought the timeout would need to sync or at least the initial timeout time on the first load of the page would need to be stored (session storage perhaps).
> > 
> > Yeah that's a good point. Maybe I'll need to reuse more of the modal code
> 
> There seems like there's alot of cross over here just the implementation differs. It may be worth us pairing up on this.

Yeah mate, sounds good

rmccar:
[Countdown on thank you page (5 SP)](https://trello.com/c/K8KpQm1f/5037-countdown-on-thank-you-page-5-sp)

yatesn:
PR is in draft - need to clarify assumptions about functionality. Documentation may need updating

src/components/panel/index.njk
```diff
@@ -21,6 +21,42 @@ Panels are used to draw information to the attention of the user.
...
 }}
 
+### Warning with countdown
+{{
+   patternlibExample({"path": "components/timeout-panel/examples/warn/index.njk"})
+}}
+
+Use this warning panel variant when a service sets a session expiry time to protect the user’s information.
+
+When the countdown reaches zero, the browser will navigate to a specified URL and show some alternate text. If the current URL is provided the page will just be refreshed.
```
jrbarnes9, line 31: Is "show some alternate text" referring to the message shown in the panel when the time runs out? If so, that should come first...
```suggestion
When the session expiry timer reaches zero, the `countdownExpiredText` message will be displayed and the browser will redirect to the URL set for a [session timed out error page](/patterns/error-status-pages/#session-timed-out-error-when-user-is-not-signed-in). If the current URL is set, the page will reload.
```

src/components/panel/index.njk
```diff
@@ -21,6 +21,42 @@ Panels are used to draw information to the attention of the user.
...
+
+The content of the countdown panel is set using the parameters shown in the macro options table in the previous example.
+
+You will also need to set the parameters:
+
+- `sessionExpiresAt` with the current expiry time in ISO format on page load
+- `serverSessionExpiryEndpoint` with a relative url to the server’s API endpoint for the component to send requests to
```
jrbarnes9, line 38:
```suggestion
- `serverSessionExpiryEndpoint` with a relative URL to the server’s API endpoint for the component to send requests to
```

src/components/panel/index.njk
```diff
@@ -21,6 +21,42 @@ Panels are used to draw information to the attention of the user.
...
+- `redirectUrl` with a url to the [session timed out error page](/patterns/error-status-pages/#session-timed-out-error-when-user-is-not-signed-in) the user is directed to when the session has timed out
```
jrbarnes9, line 39:
```suggestion
- `redirectUrl` with a URL to the [session timed out error page](/patterns/error-status-pages/#session-timed-out-error-when-user-is-not-signed-in) the user is directed to when the session has timed out
```

src/components/panel/index.njk
```diff
@@ -21,6 +21,42 @@ Panels are used to draw information to the attention of the user.
...
+The request is made to check if the same session is being interacted with in another tab.
+
+A request to get the current expiry time will be sent when the following happens:
+
+- at 20-second intervals after the page loads
+- before the page starts to redirect
+
+#### Failure responses from the server
+
+If the response status does not have a successful `200` code, the component will handle the response as a session expiry and redirect the user to the url set in the `redirectUrl` parameter.
```
jrbarnes9, line 58:
```suggestion
If the response status does not have a successful `200` code, the component will handle the response as a session expiry and redirect the user to the URL set in the `redirectUrl` parameter.
```

kruncher (approved):
Looks goods to me

src/components/panel/index.njk
```diff
@@ -21,6 +21,42 @@ Panels are used to draw information to the attention of the user.
...
+### Warning with session expiry timer
+{{
+   patternlibExample({"path": "components/timeout-panel/examples/warn/index.njk"})
+}}
+
+Use this warning panel variant when a service sets a session expiry time to protect the user’s information.
+
+When the session expiry timer reaches zero, the `countdownExpiredText` message will be displayed and the browser will redirect to the URL set for a [session timed out error page](/patterns/error-status-pages/#session-timed-out-error-when-user-is-not-signed-in). If the current URL is set, the page will reload.
+
+The content of the countdown panel is set using the parameters shown in the macro options table in the previous example.
```
jrbarnes9, line 33:
```suggestion
The content of the warning panel is set using the parameters shown in the macro options table in the previous example.
```

jrbarnes9 (approved):

kruncher (approved):

Issue #1996, opened 2022-02-22
Collapsible 'show' elements failing WCAG 4.1.1 Parsing (Level A)


jrbarnes9:
A recent DAC audit has flagged a level A issue with the collapsible component. 

WCAG Reference: 
4.1.1 Parsing (Level A)

The ‘Show’ elements are not nested accordingly to the HTML5 standards. The ‘Show’ 
elements are buttons that are nested within a link, and this is not valid markup. This may 
cause problems for assistive technology. 

```
<div class="ons-collapsible__heading ons-js-collapsible-heading svelte-1gosn14" 
role="link" aria-controls="topic-0" aria-expanded="false" data-ga-action="Close 
panel">
[...] 
<button type="button" class="ons-btn ons-collapsible__btn ons-js-collapsible-
button ons-u-d-no@xxs@s ons-btn--secondary ons-btn--small" aria-controls="topic-0" 
data-ga-action="Close panel"> 
<span class="ons-btn__inner ons-js-collapsible-button-inner">Show</span> 
</button> 
[...]
</div> 
```

_Originally posted by @James-Nitron in https://github.com/ONSdigital/design-system/discussions/148#discussioncomment-2206300_

boxadesign:
This is on hold pending testing from DAC. I have created a new version that will need to be tested before going forward.

https://deploy-preview-2013--ons-design-system.netlify.app/components/accordion/
https://deploy-preview-2013--ons-design-system.netlify.app/components/collapsible/

PR #2013, opened 2022-02-28
Collapsible - use details element and update functionality

### LINKED ISSUE #2253: Accordion show/hide all button doesn't announce current state of accordions (DAC usability)
DAC suggest adding `aria-expanded` to the 'show all' and 'hide all' buttons.

(Page 86-87 of June 2022 DAC DS report...)

---

## GOV Accordions (Usability)

Accordions do not follow the GOV Design System specifications.

### Reference:
Usability & GOV.UK Design System requirement

### Issue ID: DAC_Usability_06
URL: https://ons-design-system.netlify.app/components/accordion/examples/accordion/
Page title: Example
Journey: Journey 1, Step 1, b

### Screenshot:
Accordion pages provide a ‘Show all’ button. The button can expand all of the accordions on the page. There is no aria-expanded attribute for the show all button; however, the accordion links have an aria-expanded attribute. This may cause some confusion to screen reader users.

### Current code ref(s):
#accordion-1 > div.ons-collapsible__heading.ons-js-collapsible-heading 
```
<button type="button" class="ons-btn ons-js-collapsible-all ons-u-mb-s ons-btn--secondary ons-btn--small" data-close-all="Hide all" data-group="accordion" data-ga-label="Hide all"> <span class="ons-btn__inner ons-js-collapsible-all-inner">Show all</span> </button>
```

### Screen reader comments: 
“The ‘show all’ accordion on the ‘example’ page does not comply with the GOV.UK Design System guidelines as the state of the accordion is not announced. Including the collapse or expand within the button would make it simpler to understand if the form label is expanded or collapsed.”

### Solution:

We recommend that the GOV Design System accordions are adopted to the office for national statistics accordions. The GOV Design System accordions are known to be easy to use.
See GOV Design System Accordions for an example.

boxadesign:
Have refactored the collapsible component (and accordion) to use the `details` element.  This addresses #1996 and simplifies our approach to this component. 

The "Hide/Show" button that is displayed inside the content of a collapsible or displayed next to the title of an accordion has been removed as this has been shown to be unessecary for usability and by moving to using the native html5 `details` element does not fail on our accessibility requirements.

The `details` element also works without javascript enabled so there is no reduction in use.

This PR removes the javascript that handled the `aria` attributes as these are not needed with the `details` element. This will require a second test with DAC to ensure the components perform as required with various accessibility software tools.

Fixes #2253 and #1996 

### Breaking changes

Removal of the `variants` parameter as we do not need the `simple` variant as the output is now the same as what the `simple` variant produced.

### How to review

Test the collapsible and accordion components and any patterns that they are used in.


jrbarnes9:
This approach was approved by DAC, but we need to add `aria-expanded` to the "show all" button. We should also remove the underline from the single collapsible (and only show on hover) as part of this.

kruncher (dismissed):
This looks good to me.

One suggestion; it might be nice to move the group/accordion logic into the accordion component instead of having it inside collapsible (and to rename from `collapsible.group.js` to `accordion.js`).

jrbarnes9 (commented):
Does the HTML markup needs trimming before it's used as value for `data-ga-label`?
Screenshot 2022-07-25 at 11 46 19

kruncher (approved):

jrbarnes9 (approved):

PR #2041, opened 2022-03-14
Fix jinja error for field

rmccar:
This PR fixes a jinja error in RASRM where in onsField it is looking for `params.label.inline` when `label` doesn't exist these changes will fix that.

Also removes `legendIsQuestionTitle` from some onsField calls because onsField doesn't use that param.
And adds quotes where needed in some examples

### How to review
- Check examples all still look and work as expected

### DIFF
--- a/src/components/textarea/_macro.njk
+++ b/src/components/textarea/_macro.njk
@@ -68,7 +68,6 @@
             "id": params.fieldId,
             "classes": params.fieldClasses,
             "dontWrap": params.dontWrap,
-            "legendIsQuestionTitle": params.legendIsQuestionTitle,
             "error": params.error
         }) %}
             {{ field | safe }}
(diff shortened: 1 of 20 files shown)


src/components/textarea/_macro.njk
```diff
@@ -68,7 +68,6 @@
             "id": params.fieldId,
             "classes": params.fieldClasses,
             "dontWrap": params.dontWrap,
-            "legendIsQuestionTitle": params.legendIsQuestionTitle,
```
boxadesign, line 71: I think this is still needed for mutually exclusive when not used in a question macro

rmccar: I think thats for onsFieldset not onsField. Ive left it in for onsFieldset

boxadesign (approved):

jrbarnes9 (approved):

PR #2096, opened 2022-03-28
Add ability to set attributes on list items

rmccar:
Add ability to be able to set HTML attributes on link list items. This will allow data attributes to be set for runner error panel tests to find elements

This PR also removes the `rel` and `index` params. `rel` can now be set using the attributes param and it was felt `index` was no longer needed because there are other ways of prefixing lists with numbers.

Also this PR will fix the issue with extra spaces in the class attribute on list items.

### What are the breaking changes?
The breaking changes included in this PR mean that the `rel` and `index` attributes will no longer be available on lists.
- Use of `index` should be replaced with setting `"element": 'ol'` on the list and if needed add the `data-qa` attribute using the new `attributes` param. 
- `rel` should also now be set using the new `attributes` param.

### How to review
- Add an attribute to a list item that is a link and see that it has been added to the HTML
- See that extra space is removed from list item class lists

Fixes: #2097

### DIFF
--- a/src/components/lists/_macro-options.md
+++ b/src/components/lists/_macro-options.md
@@ -17,12 +17,10 @@
 | url | string | false | Will wrap the text in a link |
 | listClasses | string | false | Classes to be added to the list item |
 | variants | string | false | Used for variations. Available `variants` values: `inPageLink` for use with error lists when a `url` is also provided |
-| rel | string | false | allows setting of rel attribute for list item links |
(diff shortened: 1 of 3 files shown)


jrbarnes9 (commented):
Maybe we should remove `rel` and `index` as params and just allow those attributes to be added via `item.attributes`?

jrbarnes9 (approved):

boxadesign (approved):

PR #2151, opened 2022-04-26
Mutual exclusive multiple options

### LINKED ISSUE #2082: Allow ability to provide more than one mutually exclusive option
Allow having more than one mutually exclusive option. If the number of options is more than one make them radios if its only one then it will still be a checkbox

rmccar:
This PR is to enable multiple mutually exclusive options. When more than one exclusive option is provided the options become radio buttons but if only one is provided then it stays a checkbox.

This includes breaking changes 
- the `checkbox` param which was an object is now renamed `exclusiveOptions` and now an array of objects
- the `deselectCheckboxAdjective` is now called `deselectExclusiveOptionAdjective`

Fixes #2082

Also fixes an issue with extra space being output in the checkbox class list in the HTML
And removes a couple of triggerEvent lines to fix errors with the component working in IE11

### How to review
Check new multiple mutually exclusive options works as expected


rmccar:
[Multiple mutually exclusive options DS change (3 SP)](https://trello.com/c/4GriW2aG/5232-multiple-mutually-exclusive-options-ds-change-3-sp)

kruncher (approved):
@boxadesign do you think this breaks any service side logic? (services that subscribe to change/input events on the input controls directly)

boxadesign (approved):

kruncher (approved):

Issue #2141, opened 2022-04-13
Add disabled option to checkboxes


jrbarnes9:
The use of checkboxes in filters requires the addition of a disabled state for items which return zero result counts.

- `$Grey35` can safely be used for the checkbox and label as [WCAG2.1 states](https://www.w3.org/TR/WCAG21/#contrast-minimum) "disabled interface need not meet contrast minimum" (see incidental)
- `aria-disabled` could be applied to the `label` and `input` to supplement the standard `<input ... disabled />` [MDN reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled)
- cursor should be set to `cursor: not-allowed;` for label and input.

As seen in 

![Publications filter selected (disabled zero results)](https://user-images.githubusercontent.com/22910336/163198948-3455e1cc-6485-4237-8a90-375da1c35a07.png)

and
<img width="197" alt="Screenshot 2022-04-13 at 11 09 40" src="https://user-images.githubusercontent.com/22910336/163197697-571e75be-04c1-4d8c-ba4d-f84c34087c78.png">

@jondewijones following this item for use in search.

_Originally posted by @armstrongb in https://github.com/ONSdigital/design-system/discussions/13#discussioncomment-2559997_

jrbarnes9:
@yatesn New request from DP

jrbarnes9:
Resolved in #2187

PR #2187, opened 2022-05-16
Feature/disabled checkboxes

jondewijones:
Added disabled styling for checkbox and checkbox labels elements - https://github.com/ONSdigital/design-system/issues/2141

### How to review
-Sense check
-Test work/look as they should


src/components/checkboxes/_checkbox-macro.njk
```diff
@@ -7,6 +7,7 @@
             id="{{ params.id }}"
             class="ons-checkbox__input ons-js-checkbox {{ params.inputClasses }}"
             value="{{ params.value }}"
+            {%if params.disabled %}disabled aria-disabled="true"{%endif %}
```
boxadesign, line 10: We probably want to be a bit more verbose here and check for the value of `disabled`. Jinja2 in our python services can sometimes throw a wobbly when we don't define variables in specific ways. So maybe something like...

`{% if params.disabled is defined and params.disabled == true %}`

jondewijones: 👍

src/components/checkboxes/_checkbox.scss
```diff
@@ -30,6 +30,16 @@ $checkbox-padding: 11px;
     &:checked {
       background-size: 14px;
     }
+
+    &:disabled {
+      border: 2px solid $color-grey-35;
```
jrbarnes9, line 35: Could you assign a variable to this colour in `_colors.scss`? e.g. `$color-borders-disabled`

src/components/checkboxes/_checkbox.scss
```diff
@@ -30,6 +30,16 @@ $checkbox-padding: 11px;
...
+      cursor: not-allowed;
+    }
+
+    &:disabled + label {
+      color: $color-grey-35;
```
jrbarnes9, line 40: Could you assign a variable to this colour in `_colors.scss`? e.g. `$color-text-disabled`

src/components/checkboxes/_checkbox.scss
jondewijones: Added variables under `// Panels and status` as disabled felt like status but happy to move to a better location

src/components/checkboxes/_checkbox.scss
jondewijones: Added variables under `// Panels and status` in `_colors.scss` as disabled felt like status but happy to move to a better location

boxadesign (approved):

jrbarnes9 (approved):

PR #2199, opened 2022-05-23
Checkbox macro - parameter to visually hide label

boxadesign:
This variant visually hides the label of a single checkbox with the parameter `hideLabel: true`

This parameter will visually hide a label for a checkbox.

The example created can be viewed here: /components/checkboxes/#checkbox-with-hidden-label

### How to review
Sanity check the code additions and documentation.


kruncher (approved):
Looks good to me

PR #2212, opened 2022-05-26
Remove inert attribute to fix modal

rmccar:
Updating to Chrome 102 has uncovered a bug when using the `inert` attribute with modals. Because it now is supported by Chrome as of 102 it is causing the modal button itself to become inert so that the user is unable to click on it. The functionality we are getting from `inert` (disabling the page content below the modal) we are already getting for free by using the modal. So this PR is to remove the use of `inert` from the DS.

### How to review
Test in eQ to see that the page underneath the modal is not interactable/tabbable. Instructions on how to do this are at the bottom of the readme on this page (https://github.com/ONSdigital/eq-questionnaire-runner) and its best to use the `test_timeout_modal` schema

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -52,7 +52,6 @@ export default class Modal {
         document.querySelector('body').classList.add(ie11Class);
       }
 
-      this.makePageContentInert();
       this.saveLastFocusedEl();
 
       if (event) {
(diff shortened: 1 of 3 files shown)


src/tests/spec/modal/modal.spec.js
```diff
@@ -73,10 +69,6 @@ describe('Component: Modal', function() {
         expect(document.querySelector('.ons-page').getAttribute('aria-hidden')).to.equal('false');
```
boxadesign, line 69: Need to remove these aria-hidden tests also

rmccar: Done

rmccar: Decided to reinstate this as I think we still need to set aria-hidden values on the page and the modal

src/components/modal/modal.js
```diff
@@ -79,17 +79,16 @@ export default class Modal {
     }
   }
 
-  makePageContentInert() {
+  toggleModalAndPageAriaHidden() {
```
boxadesign, line 82: We don't need to do any of this as the `dialog` handles all of the background page state. What you originally changed was correct, to remove it all,

rmccar: Ah ok ill remove the commits

boxadesign (approved):

MebinAbraham (approved):
Fixed the function issue 👍 

Worth noting, I do think there is another bug (non-critical). When I am active on the page, move movement etc, it doesn't seem to care, I still get the timeout modal if I hadn't referesh/posted the page.

PR #2217, opened 2022-05-30
Add missing `s` in fieldClasses

JoelClemence:
Fixing a bug I found whereby the `fieldClasses` are not being applied to the field

### How to review
Create a select in the following way

```
{{
    onsSelect({
        "id": "second_period",
        "name": "second_period",
        "classes": "ons-u-wa--@xxs",
        "fieldClasses": "ons-grid__col--flex ons-col-2@m",
        "label": {
            "text": "Second Period",
        },
        "options": data.period_options
    })
}}
```
The rendered HTML should add the classes: `grid__col--flex ons-col-2@m` to the ons field however these are not added

### DIFF
--- a/src/components/select/_macro.njk
+++ b/src/components/select/_macro.njk
@@ -4,7 +4,7 @@
 
     {% call onsField({
         "id": params.fieldId,
-        "classes": params.fieldClases,
+        "classes": params.fieldClasses,
         "legendClasses": params.legendClasses,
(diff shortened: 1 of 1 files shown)


DanWhite3 (approved):

kruncher (changes requested):
Needs a unit test for regression purposes.

jrbarnes9:
This was fixed in #2256

PR #2267, opened 2022-06-20
Remove space from checkbox label class list

rmccar:
Currently there is a space added at the end of checkbox label class lists when a custom class isn't specified

### How to review
Compare this branch against master and see that space is removed after `ons-checkbox__label` in html

### DIFF
--- a/src/components/checkboxes/_checkbox-macro.njk
+++ b/src/components/checkboxes/_checkbox-macro.njk
@@ -20,7 +20,7 @@
             "for": params.id,
             "inputType": "checkbox",
             "text": params.label.text,
-            "classes": "ons-checkbox__label " + params.label.classes | default(''),
+            "classes": "ons-checkbox__label" + ( ' ' + params.label.classes if params.label.classes else ""),
             "description": params.label.description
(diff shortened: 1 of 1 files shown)


boxadesign:
@rmccar Been discussing this and we're thinking of ditching the need to not have whitespace when classes aren't applied. It doesn't break standards and there is no performance hit

rmccar:
> @rmccar Been discussing this and we're thinking of ditching the need to not have whitespace when classes aren't applied. It doesn't break standards and there is no performance hit

Yeah its a pretty low priority issue, its just something I find a bit annoying. I think it makes the code look untidy, especially in this case where we show it in examples to the user and its something thats pretty easy to fix

boxadesign:
@rmccar Are you happy to close this? We don't want to cater for this anymore due to addition of logic that doesn't really solve any technical problems.

rmccar:
> @rmccar Are you happy to close this? We don't want to cater for this anymore due to addition of logic that doesn't really solve any technical problems.

Ok yeah fair enough if thats the decision thats been made

PR #2271, opened 2022-06-21
Fix/simple accordion tabindex

### LINKED ISSUE #2147: Accordion collapsibles of the simple variant are not in the tab order
Accordion collapsibles of the simple variant are not in the tab order

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/1377#discussioncomment-2629401_

boxadesign:
Fixes #2147 

Also adds tests in the collapsible script tests to cover the scenario when the collapsible is and accordion and when the collapsible is an accordion and the simple variant.

### How to review
Review code and check tests

### DIFF
--- a/src/components/collapsible/collapsible.js
+++ b/src/components/collapsible/collapsible.js
@@ -6,6 +6,7 @@ export default class Collapsible {
     // State
     this.group = collapsibleElement.getAttribute('data-group');
     this.isAccordion = collapsibleElement.classList.contains('ons-collapsible--accordion');
+    this.isSimpleAccordion = collapsibleElement.classList.contains('ons-collapsible--simple');
 
     // Elements
     this.collapsible = collapsibleElement;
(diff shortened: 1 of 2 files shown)


src/components/collapsible/collapsible.js
```diff
@@ -29,7 +30,7 @@ export default class Collapsible {
     this.collapsibleHeader.setAttribute('role', 'link');
     this.collapsibleHeader.setAttribute('aria-controls', collapsibleId);
 
-    if (!this.isAccordion) {
+    if (!this.isAccordion || (this.isAccordion && this.isSimpleAccordion)) {
```
kruncher, line 33: This could be simplified to:
```javascript
if (!this.isAccordion || this.isSimpleAccordion) {
```

kruncher: `this.isAccordion` has to be true if it wasn't false.

kruncher (approved):

jrbarnes9 (approved):

PR #2277, opened 2022-06-23
Add visually hidden screen reader context to onsExternalLink

### LINKED ISSUE #2145: External link component missing visually hidden 'new tab' message for screen readers (WCAG Level A)
External link component missing visually hidden 'new tab' message for screen readers

See `screenreaderMessage` used in [lists](https://ons-design-system.netlify.app/foundations/typography/#lists)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/233#discussioncomment-2614687_

jrbarnes9:
Fixes #2145 

- Added visually hidden span of text to external links, with brackets to allow a pause after button label
- Added attributes to svg to ensure it's not focusable or read by screen readers
- Also updated CTA 'new window' button macro, options and example to bring inline with external link
- Updated/added tests to reflect changes

### How to review
- Check screen readers correctly announce that an external link will open in a new tab
- Check tests pass


boxadesign (approved):
Looks OK but requires master to be merged in to run macro tests in CI.

src/components/button/examples/button-link-new-window/index.njk
```diff
@@ -2,7 +2,11 @@
 {{
     onsButton({
         "text": 'Web chat',
-        "url": '#0',
-        "newWindow": true
+        "url": '/patterns/send-a-web-form/examples/form/',
+        "newWindow": true,
+        "newWindowDescription": 'opens in a new window',
```
rmccar, line 7: If we want it to default to say it opens in a new tab don't we want all references to `newWindow` to change to `newTab`?

jrbarnes9: Was just trying to avoid a breaking change tbh

rmccar: Yeah fair enough makes sense to make the change I think, and it's something that is pretty easy to fix. Not a big thing though

boxadesign (approved):

rmccar (approved):

src/components/button/examples/button-link-new-window/index.njk
jrbarnes9: We have a number of naming inconsistencies we have an epic for, so we'll pick up all those breaking changes at the same time: #1888

PR #2290, opened 2022-06-29
Fix mutually exclusive bug for runner

rmccar:
This change will allow mutually exclusive radio buttons to be no longer hardcoded  to `mutuallyExclusiveRadio` which fixes a bug in runner. The bug in runner is stopping the clicked answer being picked up because it uses the mutually exclusive options name to find it.

### How to review
Check that multiple mutually exclusive options still work and look as expected

### DIFF
--- a/src/components/mutually-exclusive/_macro.njk
+++ b/src/components/mutually-exclusive/_macro.njk
@@ -52,7 +52,7 @@
                             "dontWrap": true,
                             "radios": params.exclusiveOptions,
                             "inputClasses": "ons-js-exclusive-option",
-                            "name": "mutuallyExclusiveRadio",
+                            "name": params.exclusiveOptions[0].name,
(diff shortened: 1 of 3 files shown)


jrbarnes9 (changes requested):
It's currently allowing both mutually exclusive radios to be selected...

Do our tests check for this? If so, they should be failing. If not, we should add a test?
Screenshot 2022-06-29 at 11 03 21

boxadesign:
@rmccar You will need to add a test to check that the correct `name` value is added to both mutually exclusive options

boxadesign:
> 

Ok. If you're testing that the `name` attribute is there then you shouldn't need to test that only one is selected. That's standard radios behaviour.

boxadesign (approved):

jrbarnes9 (approved):

PR #2297, opened 2022-06-30
Fix poweredby Jinja2 error

rmccar:
This fixes a bug in Jinja2 which causes an error if `params.poweredBy` isn't set on a footer

### How to review
Check footers still work as expected

### DIFF
--- a/src/components/footer/_macro.njk
+++ b/src/components/footer/_macro.njk
@@ -9,25 +9,13 @@
         {% set lang = 'en' %}
     {% endif %}
 
-    {% if params.poweredBy is defined and params.poweredBy %}
-        {% if params.poweredBy.alt is defined and params.poweredBy.alt %}
-            {% set poweredByAlt = params.poweredBy.alt %}
-        {% else %}
-            {% if params.lang == 'cy' %}
-                {% set poweredByAlt = 'Swyddfa Ystadegau Gwladol' %}
-            {% else %}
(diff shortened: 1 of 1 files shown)


src/components/footer/_macro.njk
```diff
@@ -10,15 +10,15 @@
...
+
         {% if params.poweredBy is defined and params.poweredBy and params.poweredBy.logo is defined and params.poweredBy.logo %}
             {{
                 onsIcon({
                     "iconType": params.poweredBy.logo,
                     "altText": params.poweredBy.alt | default("Office for National Statistics")
                 })
             }}
-        {% else %}
+        {% elif params.poweredBy is defined and params.poweredBy %}
```
jrbarnes9, line 21: This is causing the footers to lose their default ons logo.

jrbarnes9: I think we just need to remove the params for the alt text if on lines 26 and 33, and just hard code them because we know what the icons are.

jrbarnes9 (changes requested):

src/components/footer/_macro.njk
rmccar: I have managed to fix that but if you're happy hardcoding them then Ill just do that 👍

jrbarnes9: Yeah that's fine

jrbarnes9 (approved):

boxadesign (approved):

PR #2312, opened 2022-07-11
2309 hero bg images

### LINKED ISSUE #2309: Update hero to allow for background images
Hero component needs to be more flexible, and to be controlled by brand css, so we should allow background images to be applied at relevant breakpoints.

IDS requires this for w/c 11/7

Image
Image
Image

jrbarnes9:
Resolves #2309 

- Moved census-specific styling to census theme css
- Added IDS brand theme css
- Replaced census params with common `variants`

### How to review
- Check percy and macro tests pass
- Check layout of all hero variants at all breakpoints


src/components/hero/_macro.njk
```diff
@@ -3,7 +3,7 @@
 {% from "components/icons/_macro.njk" import onsIcon %}
 
 {# style adjustments if dark or light skin #}
-{% if params.censusThemeDark is defined and params.censusThemeDark %}
```
kruncher, line 6: Test for `censusThemeDark` needs to be removed.

src/components/hero/_macro.njk
```diff
@@ -3,7 +3,7 @@
...
+{% if params.variants is defined and 'dark' in params.variants %}
```
kruncher, line 6: Tests for `params.variants` need to be added; there is a good example of this type of test here: https://github.com/ONSdigital/design-system/blob/master/src/components/accordion/_macro.spec.js#L57

kruncher (changes requested):

kruncher (dismissed):

kruncher (approved):

PR #2316, opened 2022-07-12
Fix footer Jinja2 bug

rmccar:
Fixes a bug with the new footer changes. Causes an error in a Jinja2 environment where its testing for `params.poweredBy.partnership` but not checking if `params.poweredBy` exists first

### How to review
Footer still works as expected

### DIFF
--- a/src/components/footer/_macro.njk
+++ b/src/components/footer/_macro.njk
@@ -38,8 +38,6 @@
                     "altTextId": 'poweredby-alt'
                 })
             }}
-        {% else %}
-            {{ onsLogo | safe }}
         {% endif %}
     {% endset %}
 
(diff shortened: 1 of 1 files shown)


src/components/footer/_macro.njk
```diff
@@ -38,8 +38,6 @@
                     "altTextId": 'poweredby-alt'
                 })
             }}
-        {% else %}
```
kruncher, line 41: Why is this part being removed? it doesn't seem to be related to the `params.poweredBy.partnership` thing which is being raised.

rmccar: If you follow the logic whatever logo is being set here is being used in the `params.poweredBy.partnership` bit below. To keep the same functionality I had to in effect move this else down there

jrbarnes9 (approved):
LGTM. Everything seems to output the same as expected.

kruncher (approved):

PR #2323, opened 2022-07-13
Fix: Cookies settings - check for missing cookie type

boxadesign:
Fixes https://github.com/ONSdigital/ons-ids-website/issues/37

The cookie settings page JS required each `cookieType` to have a corresponding set of `radios`.

This PR fixes this by checking for a matching radio before setting it to `checked`

### How to review
Remove a set of radios from the settings page and check there are no console errors and that all other radios are checked.


kruncher (approved):
This looks good to me.

jrbarnes9 (approved):

PR #2324, opened 2022-07-13
Navigation: add aria-label to each item

### LINKED ISSUE #2320: Add `ariaLabel` option to `Navigation.Item` for use in navigation components (i.e. header, list of links, etc)
It would be useful to have `Navigation.Item` support the `ariaLabel` option in a similar way to how other elements do.

This is useful for situations where a navigation link is shortened and does not match the title of the target page. For example, the link could be output as:

```html
<a href="./the-page" aria-label="Example of a page">Example</a>
```

boxadesign:
Closes #2320 

### How to review
Check the code and output is correct

### DIFF
--- a/src/components/header/_macro-options.md
+++ b/src/components/header/_macro-options.md
@@ -56,12 +56,13 @@
 
 ## Item
 
-| Name    | Type   | Required | Description                                                           |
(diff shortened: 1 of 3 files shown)


kruncher (approved):
Looks good to me.

PR #2329, opened 2022-07-15
Fix font weight on fieldset legends

rmccar:
There is a difference in font weight between fieldsets and labels. The fieldset css is inherited from label but is then overridden by a `font-weight: normal;` line. This PR removes that line.

Screenshot 2022-07-15 at 10 47 35

### How to review
Test the duration example on master and on this branch with these params set to see the difference
```   
"dontWrap": false,
"legendOrLabel": "label",
```

### DIFF
--- a/src/components/checkboxes/_macro.njk
+++ b/src/components/checkboxes/_macro.njk
@@ -85,7 +85,7 @@
                                     "text": checkbox.other.label.text
                                 },
                                 "legend": checkbox.other.legend,
-                                "legendClasses": checkbox.other.legendClasses | default('') + ' ons-u-mb-xs',
+                                "legendClasses": checkbox.other.legendClasses,
                                 "value": checkbox.other.value,
                                 "autoSelect": checkbox.other.autoSelect,
                                 "selectAllChildren": checkbox.other.selectAllChildren,
--- a/src/components/checkboxes/examples/checkboxes-other-radios/index.njk
+++ b/src/components/checkboxes/examples/checkboxes-other-radios/index.njk
@@ -31,7 +31,7 @@
                     "otherType": "radios",
                     "id": "phone-time",
                     "name": "phone-time",
-                    "legend":  "Choose preferred time of day",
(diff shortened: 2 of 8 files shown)


rmccar:
[Font weight wrong on duration fieldsets](https://trello.com/c/vfaQP4Ry/5371-font-weight-wrong-on-duration-fieldsets)

rmccar:
[DS v52 bug - Duration labels not being displayed correctly](https://trello.com/c/FKIKamSt/5364-ds-v52-bug-duration-labels-not-being-displayed-correctly)

boxadesign (dismissed):

jrbarnes9 (commented):
Nested legends always had regular font-weight. So this has caused a change to that...
Screenshot 2022-07-18 at 15 03 11
Screenshot 2022-07-18 at 15 03 14
Screenshot 2022-07-18 at 15 04 33

jrbarnes9:
From Percy it looks like there's a difference in bottom margins between labels, legends and nested label/legends. Should just look to align all these, perhaps by extending `ons-u-mb-xs` rather than specific .6 or .4 rem spacing.

rmccar:
> `checkboxes__label` is rendered when not set... 
Screenshot 2022-07-21 at 15 47 16

Screenshot 2022-07-21 at 16 04 21
Its required to be set

jrbarnes9:
> > `checkboxes__label` is rendered when not set...
> > Screenshot 2022-07-21 at 15 47 16
> 
> Screenshot 2022-07-21 at 16 04 21 Its required to be set

Oh yeah I can see now that issue is on the main branch. Probably explains why the margin was removed from the nested legend then.  I'll raise a separate ticket.  I don't think we'd need both legend and label for nested checkboxes

jrbarnes9 (approved):

jrbarnes9 (approved):

kruncher (approved):
Looks good to me

PR #2349, opened 2022-07-21
Fix: Make input prefix/suffix params mandatory to render component

### LINKED ISSUE #2344: Suffix/prefix aria-labelledby is required but can be left empty
It's currently possible to build an input with a suffix or prefix, but without setting the `id` to be used by the input's `aria-labelledby` attribute.  Conditions should be set in the macro so the component can't be built without it.

jrbarnes9:
Fixes #2344 

- Added conditions to input macro to ensure suffix/prefix objects have `id` set because it's required for `aria-labelledby` on the input
- Added conditions to input macro to ensure suffix/prefix objects have `title` set because it's required for `<abbr>` element to be valid
- Added new macro tests for these conditions
- Updated macro tests and examples to add these required params

### What are the breaking changes?

- All inputs using `params.prefix` must set both `params.prefix.id` and `params.prefix.title` to render.
- All inputs using `params.suffix` must set both `params.suffix.id` and `params.suffix.title` to render.

### How to review
- Check tests pass and that all prefix/suffix inputs cannot render unless title and id are set.


src/components/input/_macro.njk
```diff
@@ -71,21 +71,26 @@
         {% endif %}
 
         {% if params.prefix is defined and params.prefix or params.suffix is defined and params.suffix %}
-            {% if params.prefix is defined and params.prefix %}
-                {% set prefixClass = " ons-input-type--prefix" %}
-            {% endif %}
+            {% if (params.prefix.id is defined and params.prefix.id) or (params.suffix.id is defined and params.suffix.id) %}
```
rmccar, line 74:
```suggestion
            {% if (params.prefix is defined and params.prefix and params.prefix.id is defined and params.prefix.id) or (params.suffix is defined and params.suffix and params.suffix.id is defined and params.suffix.id) %}
```
This line is causing an error in jinja2 unless you make this change. 
It also means you can get rid of the if above it but if you do that, not setting the id will still render inputs without no suffix/preffix. If you leave it in there will be nothing rendered if the id isn't set instead. Im not sure which functionality you're going for

kruncher (dismissed):
Looks good to me

src/components/input/_macro.njk
jrbarnes9: Yeah I ideally we should prevent both the input and prefix rendering if they don't set the required attributes.  Should make it easier to see when something is missing I think

kruncher (approved):

rmccar (approved):

Issue #2357, opened 2022-07-28
Simplify checks for defined properties since Jinja2 can be configured to avoid that issue


kruncher:
Currently there are a lot of excessive conditions when testing for defined properties in component macros. These checks were added so that the component macros would work in Jinja2 (as well as Nunjucks):

```nunjucks
{% if foo is defined and foo.bar is defined and foo.bar.baz %}
```

It seems that Jinja2 can be configured to operate in a similar way to how Nunjucks operates (see https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.Undefined).

These checks can now be simplified:

```nunjucks
{% if foo.bar.baz %}
```

Before merging these improvements into `main` it is important to test against a Jinja2 project.

kruncher:
Relates to https://github.com/ONSdigital/eq-questionnaire-runner/pull/909

PR #2359, opened 2022-07-29
Remove undefined checks

kruncher:
Resolves issue #2357 

### How to review
1. Verify that tests pass with `yarn test`.
2. Verify that documentation pages render properly.
3. Verify that components still work with Jinja2.
4. 
### Notes
The majority of the cases were automatically replaced with the following script which finds all occurrences that match the pattern `foo.bar is defined and foo.bar` where foo.bar matches on both sides:
```js
const fs = require("fs-extra");
const glob = require("glob").sync;

async function processFile(filePath) {
  let fileContents = await fs.readFile(filePath, { encoding: "utf8" });
  fileContents = fileContents.replace(/(([a-zA-Z_]+[a-zA-Z0-9_]*)(\.([a-zA-Z_]+[a-zA-Z0-9_]*))*) is defined and (([a-zA-Z_]+[a-zA-Z0-9_]*)(\.([a-zA-Z_]+[a-zA-Z0-9_]*))*)/g, (a, b, c, d, e, f, g, h) => {
    return b === f ? b : a;
  });
  await fs.writeFile(filePath, fileContents, { encoding: "utf8" });
}

(async () => {
  await Promise.all(glob("**/*.njk").map(processFile));
})();
```
And then all remaining cases where `is defined` was used were updated by hand.

There were also several odd occurrences of the form `foo.bar and foo.bar is defined`.


boxadesign (dismissed):

rmccar (approved):

rmccar (dismissed):
Just testing this again now with eQ and I have a few failing tests since you merged the main branch in. Ill work out what the issues are tomorrow morning and get back to you

src/components/panel/_macro.njk
```diff
@@ -50,9 +50,9 @@
                 <span class="ons-panel__assistive-text ons-u-vh">{{ params.assistiveTextPrefix | default("Announcement: ") }}</span>
             {% endif %}
 
-            {% if params is defined and params %}
-                {% if params.title is defined and params.title %}
-                    {% if params is defined and params and params.type == 'error' %}
+            {% if params %}
```
boxadesign, line 53: I don't think we need to test `if params`

boxadesign:
@kruncher I would suggest your next commit message includes `[test-visual]` to run the visual regression tests to check any problems with output.

rmccar (approved):
Tested again this morning all seems good to me as long as Matts comment is sorted

jrbarnes9 (approved):

boxadesign (approved):

Issue #2243, opened 2022-06-16
Download resources results not announced to screen reader after filtering (WCAG Level AA)


jrbarnes9:
Add `role="status"` to the number of results div which changes after filtering, to announce to screen readers.

Page 58-59 of June 2022 DAC DS report...

----
## Status Message
Screen reader users are not informed when more content has been added to the page.
WCAG Reference:
4.1.3 Status Messages (Level AA)
Understanding Status Messages |How to Meet Status Messages

### Issue ID: DAC_Status_Message_01

URL: https://ons-design-system.netlify.app/patterns/download-resources/examples/download-resources/
Page title: Service name
Journey: Journey 2, Step 2

Changing the state of checkboxes within the filters section of the page results in the text ‘9 results of 150’ updating to reflect the updating results. The change of results text is not audibly read to screen reader users, meaning that the user may be unaware that results have updated.

### Current code ref(s):
```
#main-content > div.ons-js-adv-filter > div > div.ons-grid__col.ons-col-8\@m > div.ons-adv-filter__results-options > div.ons-adv-filter__results-count <div class="ons-adv-filter__results-count"> <span class="ons-js-adv-filter__results-count">9</span> results of 150 </div>
```

### Solution:
Results information that changes once the user has selected a filters option must be audibly announced for screen reader users to be aware that content has updated. To audibly announce the results text, an attribute role=status must be used. The role=status attribute must be added to the main container of the text that is to be announced when the page loads; this means that any change to that div will announce.

PR #2366, opened 2022-08-03
Accessibility: download resources fixes

### LINKED ISSUE #2241: Screen reader users able to navigate to content behind download resources filter dialog (WCAG Level A)
On viewports under 740px, when the filter panel is shown, the content behind is still visible to screen readers.

(Page 52-53 of the June 2022 DAC DS report...)

---
## Dialog
Dialogs were found that were not marked-up accessibly.
### WCAG Reference:
2.4.3 Focus Order (Level A)
Understanding Focus Order | How to Meet Focus Order
Issue ID: DAC_Dialog_01
URL: https://ons-design-system.netlify.app/patterns/download-resources/examples/download-resources/
Page title: Service name
Journey: Journey 2, Step 2
### Screenshot:
Screenshot 2022-06-16 at 13 45 21

Screen reader mobile users that activate the ‘Show filters’ button are not trapped to the filter dialog that appears. This means that users are able to navigate to the content that is behind the dialog. Visual screen reader users may become disorientated without trapping the users focus to the dialog.

### Current code ref(s):
```
#filter-panel <div class="ons-adv-filter__panel ons-js-adv-filter__panel ons-adv-filter__panel--is-visible" id="filter-panel"> <h2 class="ons-u-fs-l">Filters</h2> […] </div>
```
### Solution:
Ensure that the contents of the ‘Search’ page is not available to screen reader users when the ‘Filter’ page is displayed. Consider hiding the ‘Search’ page content using the CSS property ‘display: none’ and the aria-hidden="true" attribute to prevent this from being encountered by VoiceOver and TalkBack until users select ‘Show search results’.
We recommend that the filters expanding button is an accordion rather than displaying dialog type content. This can make it easier for screen reader users to navigate the filter content.

See GOV Design System Accordion for an example on accordions.

boxadesign:
Fixes #2241 #2243 

`aria-hidden` and `ons-u-d-no` are toggled on the results sorting (`.ons-adv-filter__results-options`) and results list (`..ons-js-adv-filter__gallery`) when the filter button and filter actions are pressed.

Added `role="status"` to the result count. I tested this on voiceover and it successully announced the updated result count but would be good if it could have another test.

Also this PR removes the download resources pattern examples from the DS. The component has one specific use case (census.gov.uk) and the code does not meet DS production standard.

### How to review
View the download resources page on a viewport <740 and toggle the filter button and confirm the attributes are toggled.


kruncher (approved):
Looks good to me.

jrbarnes9 (commented):
The filters toggle button is still accessible when the filter panel is shown...

Screenshot 2022-08-04 at 12 37 10

jrbarnes9 (commented):
Can now also access the header focusable elements...

Screenshot 2022-08-04 at 14 13 56

jrbarnes9 (approved):

PR #2377, opened 2022-08-05
Accessbility: Sortable table icon colour contrast

### LINKED ISSUE #2247: Sortable table column icons below min contrast ratio (WCAG Level AA)
Increase contrast of sortable table icons to min 3:1

Screenshot 2022-06-16 at 15 14 53

(Page 68-69 of June 2022 DAC DS report...)

----

### Issue ID: DAC_Colour_Contrast_Non_Text_02

URL: https://ons-design-system-prototypes.netlify.app/prototypes/dac-may2022/table-1.html#tabId2
Page title: Tabs and sortable and scrollable tables
Journey: Journey 5, Step 13

### Screenshot:
Each interactive button for sortable tables contains an up and down pointing triangle. The triangles provide visual indication that there is sort functionality present. Low vision users may have difficulties seeing the up and down triangles due to the colour contrast against adjacent colours being below the minimum.

Foreground: #BCBCBD
Background: #FFFFFF
The contrast ratio is: 1.9:1
1.4.11 Non-text Contrast (AA)
Fail for UI components and graphical objects
### Current code ref(s):
#tabId1 > table > thead > tr > th:nth-child(1) > button 
```
<button aria-label="Sort by ID" type="button" data-index="0" class="ons-table__sort-button">ID <svg id="sort-sprite-id" class="ons-svg-icon " viewBox="0 0 12 19" xmlns="http://www.w3.org/2000/svg" focusable="false" fill="currentColor"> <path class="ons-topTriangle" d="M6 0l6 7.2H0L6 0zm0 18.6l6-7.2H0l6 7.2zm0 3.6l6 7.2H0l6-7.2z"></path> <path class="ons-bottomTriangle" d="M6 18.6l6-7.2H0l6 7.2zm0 3.6l6 7.2H0l6-7.2z"></path> </svg> </button>
```

Low vision comments: “Next to id, title, abbr, legal basis, frequency, date and status are two arrows pointing up and down, they are light grey on a white background which makes them difficult to see due to the contrast, this means users may miss out on these functions.”

### Solution:
Interactive elements that are not disabled must meet a minimum contrast ratio of 3:1
The minimum contrast ratio of 3:1 helps low vision users see non-text elements with minimal difficulty.

boxadesign:
Fixes #2247 

Colour contrast on inactive sort icons was too low.

### DIFF
--- a/src/components/table/_table.scss
+++ b/src/components/table/_table.scss
@@ -13,7 +13,7 @@
   &__cell {
     @include nth-element(1, 0);
 
-    border-bottom: 2px solid $color-grey-100;
+    border-bottom: 2px solid $color-grey-75;
     overflow: hidden;
     padding: 0.5rem 0 0.5rem 1rem;
     text-align: left;
@@ -65,7 +65,7 @@
       }
 
       .ons-table__body .ons-table__row {
-        border-bottom: 2px solid $color-grey-100;
+        border-bottom: 2px solid $color-grey-75;
         display: block;
         margin-bottom: 1rem;
       }
@@ -148,7 +148,7 @@
     [aria-sort='descending'].ons-table__header {
       .ons-svg-icon {
         .ons-topTriangle {
-          fill: $color-grey-35;
+          fill: $color-grey-75;
         }
         .ons-bottomTriangle {
           fill: $color-text;
@@ -169,7 +169,7 @@
           fill: $color-text;
         }
         .ons-bottomTriangle {
-          fill: $color-grey-35;
+          fill: $color-grey-75;
         }
       }
       .ons-table__sort-button:focus {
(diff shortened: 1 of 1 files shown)


jrbarnes9 (commented):
The problem we have now is the active sort icon doesn't have enough contrast against the inactive icon, so you can't tell which way the column is being sorted. So I think we could probably lighten the inactive grey as it's technically 'disabled' at this point so possibly doesn't need to meet the contrast ratio against the white bg. 

So if we used `$color-grey-75` only for the inactive icon when the table has been sorted, that passes as 3:1 against the `$color-black` icon according to [non-text contrast of adjacent colours](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)

Screenshot 2022-08-11 at 11 45 13

boxadesign:
@jrbarnes9 I've adjusted so that when nothing has been sorted (on page load) the icons use `75` and upon sorting the "inactive" icon uses `35`

jrbarnes9 (approved):

Issue #2254, opened 2022-06-16
DS Examples should have an `h1` page heading when shown in their own tab (DAC usability)


jrbarnes9:
When you select 'Open this example in a new tab', the contents of the new tab should have a h1 level page heading to help user orientation.

(Page 89-90 of June 2022 DAC DS report...)

---

## Section Missing Heading (Usability)

Sections of the page do not have unique headings to separate them from other content.

### Reference:
Usability & GOV.UK Design System requirement

### Issue ID: DAC_Usability_08
URL: https://ons-design-system.netlify.app/components/accordion/examples/accordion/
Page title: Example
Journey: Journey 1, Step 1, b

Some example pages do not have a main heading that introduce the page content to users. Main headings help screen reader users understand the purpose of the page, and quickly orientate themselves where they are within the service.

### Screen reader comments: 
“When testing the headings on the ‘example’ page I found that there are no headings at level 1. As well as contravening the GOV.UK Design System rules regarding page title and headings at level 1 matching, pages without H1’s are problematic as they indicate main content on the page.”

### Solution:
We advise adding a main heading to the service page. Main headings help understand the content within the page and are also used for navigating the page.

PR #2380, opened 2022-08-08
Accessibility: Add DS example page title and heading

boxadesign:
Fixes: #2254 

This change alters the current hard coded "Example" used for the `title` `head` tag for the example template (viewed when a component example is opened in its own tab). It now creates a pattern of "Example: _Component name_"

A new `h1` is also included in the example template which uses the same pattern and is visually hidden so screen readers can still read and understand the context of the page.

This change will require some DAC testing as some components render their own `h1` i.e. when the component is used with the question macro. This will render two `h1` elements. It needs to be tested to understand from an accessibility perspective that the occasional output of two `h1` elements creates less confusion than no output on components that don't render an `h1` as part of the example.

We could work around this by having a frontmatter property that allows us to exclude the visually hidden `h1` manually but this could prove prone to error (not using the property).

### How to review
- View a component page
- Click to open the component example in a new tab
- Observe the `title` and hidden `h1`

### DIFF
--- a/src/views/layouts/_example.njk
+++ b/src/views/layouts/_example.njk
@@ -1,10 +1,12 @@
 {% extends "./_master.njk" %}
 {% set example = true %}
+{% set pageTitle = "Example: " ~ pageInfo.title | replace("-", " ") | capitalize %}
 
-{% set pageConfig = {"lang": pageInfo.lang, "title": "Example" } %}
+{% set pageConfig = {"lang": pageInfo.lang, "title": pageTitle } %}
 
 {% block pageContent %}
     <main class="ons-patternlib-page__example">
+        <h1 class="ons-u-vh">{{ pageTitle }}</h1>
         {% block main %}
         {% endblock %}
     </main>


jrbarnes9:
Would be worth raising a ticket to improve some of the example folder names so they better reflect the example, now we're exposing these.
These, for example, don't have enough context/aren't readable:
- /components/hero/examples/default/ (`Example: Default`)
- /components/hero/examples/branded/ (`Example: Branded`)

jrbarnes9 (commented):
Not working for all examples for some reason...
Screenshot 2022-08-11 at 12 09 52

boxadesign:
> Not working for all examples for some reason... Screenshot 2022-08-11 at 12 09 52

@jrbarnes9 This is due to examples that use `pageConfig` and don't set a `title` or do set a `title` in some other format. There's not alot we can do about that.

jrbarnes9 (approved):

PR #2387, opened 2022-08-10
Accessibility: Skip to content - move after preHeader

### LINKED ISSUE #2337: Move skip link after cookie banner in example (DAC usability issue)
DAC: Move skip link after cookie banner in cookies example, and document

Would also be useful to:
- Add Cookie banner, phase banner etc in the blocks exploded view to users can see what they should be used for
- Add a page example to the skip link component docs to see where it's positioned, how it links to main etc.
- Add cookie banner to base page template?

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/715#discussioncomment-3181057_

Page 53 of EQ RAS DAC report July 2022...

----

Skip to content link (usability)
The cookie section may be missed by blind users due to the positioning of the skip to content link.
WCAG Reference:
Usability & GOV.UK Design System requirement.
Issue ID: DAC_Skip_link_usability
URL: https://eq.surveys.onsdigital.uk/questionnaire/introduction-block/
Page title: Page is not available - ONS Business Surveys
Journey: Task 1 step 1
Screenshot:
Screenshot 2022-07-19 at 15 07 49

User focus is taken to the skip to content link before the cookies section. The reason this may be problematic is because some blind users may use this feature unaware that the cookies section is present.
If you’re following the GDS pattern then the skip[ to content link is situated after this section allowing users to encounter it.
Current code ref(s): body > div > div > a

```
 <a class="ons-skip-link" href="#main-content">Skip to main content</a>
```

Screen reader comments: “I explored the start of the page in context and noted that the ‘Skip to main content’ link appeared before the cookie banner. This could result in the cookie banner being bypassed by some screen reader users who decide to jump straight to the page’s main content. In addition, this behaviour does not comply with the GOV.UK Design system best practice. To address this concern, the ‘Skip to main content’ link should preferably appear following the cookie banner. The issue is present with JAWS, NVDA, VoiceOver, TalkBack.”

Solution:
We recommend repositioning the skip to content link so that it is encountered after the cookies section.
Screenshot 2022-07-19 at 15 07 53

Example: 
```
<header class="ons-header " role="banner"> <a class="ons-skip-link" href="#main-content">Skip to main content</a>
```

Please refer to the GOV.UK Design System homepage for further guidance.

boxadesign:
Fixes #2337 

The issue linked above also requested the following changes:

- Add Cookie banner, phase banner etc in the blocks exploded view to users can see what they should be used for
- Add a page example to the skip link component docs to see where it's positioned, how it links to main etc.
- Add cookie banner to base page template?

Changes:

- Skip link component placement in page template has been moved to after `preHeader` block (where the cookie banner should be placed)
- The documentation for cookie banner and skip to content have been updated to provide guidance on placement. I feel this is better than creating yet more examples.
- I don't think adding the cookies banner to the page template is the right thing to do. It will just cause bloat of an already complicated macro.

### DIFF
--- a/src/components/skip-to-content/index.njk
+++ b/src/components/skip-to-content/index.njk
@@ -27,7 +27,9 @@ All ONS pages must use the skip to content component above the header.
 ## How to use this component
 The skip to content link component is visually hidden until it is brought into focus with a keyboard press.
 
-It should be the first element of the page content, and link to the page's `#main` div.
+It should be the first element of the page content if not using the [cookies banner](/patterns/cookies-settings/). If your service does use the cookies banner, then it should immediately follow.  It should link to the page's main content `div` using `#id` (`id` matching the `id` attribute if the `div`).
+
+If using the [page template](/foundations/page-template) the skip to content component is automatically included in the correct place and will link to a `div` with an `id` of `main-content`. To override this you can manually add it using the `skipLink` block.
 
 ## Help improve this component
 Let us know how we could improve this component or share your user research findings.
--- a/src/components/skip-to-content/index.njk
+++ b/src/components/skip-to-content/index.njk
@@ -29,7 +29,7 @@ The skip to content link component is visually hidden until it is brought into f
 
 It should be the first element of the page content if not using the [cookies banner](/patterns/cookies-settings/). If your service does use the cookies banner, then it should immediately follow.  It should link to the page's main content `div` using `#id` (`id` matching the `id` attribute if the `div`).
 
-If using the [page template](/foundations/page-template) the skip to content component is automatically included in the correct place and will link to a `div` with an `id` of `main-content`. To override this you can manually add it using the `skipLink` block.
+If using the [base page template](/foundations/page-template) the skip to content component is automatically included in the correct place and will link to a `div` with an `id` of `main-content`. To override this you can manually add it using the `skipLink` block.
 
 ## Help improve this component
 Let us know how we could improve this component or share your user research findings.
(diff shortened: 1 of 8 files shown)


src/components/skip-to-content/index.njk
```diff
@@ -27,7 +27,9 @@ All ONS pages must use the skip to content component above the header.
 ## How to use this component
 The skip to content link component is visually hidden until it is brought into focus with a keyboard press.
 
-It should be the first element of the page content, and link to the page's `#main` div.
+It should be the first element of the page content if not using the [cookies banner](/patterns/cookies-settings/). If your service does use the cookies banner, then it should immediately follow.  It should link to the page's main content `div` using `#id` (`id` matching the `id` attribute if the `div`).
+
+If using the [page template](/foundations/page-template) the skip to content component is automatically included in the correct place and will link to a `div` with an `id` of `main-content`. To override this you can manually add it using the `skipLink` block.
```
jrbarnes9, line 32:
```suggestion
If using the [base page template](/foundations/page-template) the skip to content component is automatically included in the correct place and will link to a `div` with an `id` of `main-content`. To override this you can manually add it using the `skipLink` block.
```

src/patterns/help-users-to/cookies-settings/index.njk
```diff
@@ -21,7 +21,7 @@ There are two parts to implementing cookies control within a service; cookies ba
     patternlibExample({ "path": "components/cookies-banner/examples/cookies-banner/index.njk" })
 }}
 ## How to use
-The cookies banner must be placed directly above the ONS [header](/components/header) at the very top of the page.
+The cookies banner must be placed directly above the ONS [header](/components/header) at the very top of the page. If using the [page template](/foundations/page-template) it should be placed inside the `preHeader` block. This will ensure that it is displayed before the [skip to content](/components/skip-to-content) link.
```
jrbarnes9, line 24:
```suggestion
The cookies banner must be placed directly above the ONS [header](/components/header) at the very top of the page. If using the [base page template](/foundations/page-template) it should be placed inside the `preHeader` block. This will ensure that it is displayed before the [skip to content](/components/skip-to-content) link.
```

jrbarnes9 (approved):

Issue #2259, opened 2022-06-17
Move fieldset descriptions outside of `<legend>` (DAC suggestion)


jrbarnes9:
...and use `aria-describedby` on the fieldset with ID of description.

Otherwise, the labels (title + description) are too long for out-of-context menu

Fixes #2348

PR #2392, opened 2022-08-11
Accessibility - Label/Legend description change

### LINKED ISSUE #2348: The `<div>` for `ons-fieldset__description` is invalid HTML when nested within the `<legend>` when using `params.legendIsQuestionTitle`
The `<div>` for `ons-fieldset__description` is invalid HTML when nested within the `<legend>` when using `params.legendIsQuestionTitle`

The description should sit outside the `<legend>` and use it's `id` to reference it the fieldset with `aria-describedby`

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/1380#discussioncomment-3196692_

boxadesign:
Fixes #2348  and #2259 

In addition to moving the description out of the `legend` tag it made sense to do this for checkboxes and radios. Currently, if a description is provided for a checkbox label it is rendered inside the `label` which would in theory give the same issue; label + description is often too long for screen readers.

The reason the description lived inside the `label` was so that the entire text was clickable. For this to remain I have added `aria-hidden` to the visible description and added a secondary visually hidden description outside of the `label`.

The `label` and description are related in both `label` and `legend` uses by `aria-describedby`.

Tests have been updated to include these new changes.

### How to review
Ensure the different scenarios are adhered to, in particular:

- When a description is provided with a checkbox/radio it renders the visually hidden description and has the aria-hidden attribute on the visible description BUT confirm that this doesn't happen when the implementation is not checkboxes or radios
- All components render as previously - some css has changed to accomodate the `legend` description being moved.

### DIFF
--- a/src/components/fieldset/_macro.njk
+++ b/src/components/fieldset/_macro.njk
@@ -15,7 +15,7 @@
             >
                 <legend{% if params.description %} aria-describedBy="{{ descriptionID }}"{% endif %} class="ons-fieldset__legend{% if params.legendClasses is defined and params.legendClasses %} {{ params.legendClasses }}{% endif %}{% if params.description %} ons-fieldset__legend--with-description{% endif %}">
                     {%- if params.legendIsQuestionTitle -%}
-                        <h1 id="fieldset-legend-title" class="ons-fieldset__legend-title ons-u-mb-no ons-u-pb-m{% if params.legendTitleClasses is defined and params.legendTitleClasses %} {{ params.legendTitleClasses }}{% endif %}">
+                        <h1 id="fieldset-legend-title" class="ons-fieldset__legend-title{% if params.legendTitleClasses is defined and params.legendTitleClasses %} {{ params.legendTitleClasses }}{% endif %}">
                             {{- params.legend | safe -}}
                         </h1>
                     {%- else -%}
--- a/src/components/fieldset/_macro.njk
+++ b/src/components/fieldset/_macro.njk
@@ -13,13 +13,13 @@
                 class="ons-fieldset{% if params.classes is defined and params.classes %} {{ params.classes }}{% endif %}"
                 {% if params.attributes is defined and params.attributes %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value is defined and value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
             >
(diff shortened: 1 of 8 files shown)


jrbarnes9 (commented):
Caused some spacing issues flagged in percy

src/components/fieldset/_macro.njk
```diff
@@ -12,20 +13,20 @@
...
             >
-                <legend class="ons-fieldset__legend{% if params.legendIsQuestionTitle is defined and params.legendIsQuestionTitle %} ons-u-mb-no{% endif %}{% if params.legendClasses is defined and params.legendClasses %} {{ params.legendClasses }}{% endif %}">
+                <legend{% if params.description %} aria-describedBy="{{ descriptionID }}"{% endif %} class="ons-fieldset__legend{% if params.legendClasses is defined and params.legendClasses %} {{ params.legendClasses }}{% endif %}{% if params.description %} ons-fieldset__legend--with-description{% endif %}">
                     {%- if params.legendIsQuestionTitle -%}
-                        <h1 id="fieldset-legend-title" class="ons-fieldset__legend-title ons-u-mb-m{% if params.legendTitleClasses is defined and params.legendTitleClasses %} {{ params.legendTitleClasses }}{% endif %}">
+                        <h1 id="fieldset-legend-title" class="ons-fieldset__legend-title ons-u-mb-no ons-u-pb-m{% if params.legendTitleClasses is defined and params.legendTitleClasses %} {{ params.legendTitleClasses }}{% endif %}">
```
jrbarnes9, line 18: For some reason the values for the padding util classes (`xxs`, `m` and `l`)  don't match up with the equivilent margin classes - so switching this from margin to padding has increased this spacing by .5rem. I think the spacing should remain at 1.5rem.

jrbarnes9 (commented):
Need to remove the `padding-bottom` from `ons-radio__label--with-description`
Screenshot 2022-08-15 at 14 12 39

jrbarnes9 (commented):
Still too much space between heading and component
Screenshot 2022-08-16 at 15 59 24

src/components/fieldset/_macro.njk
jrbarnes9: @boxadesign Did you see this? ☝️

jrbarnes9:
> Still too much space between heading and component Screenshot 2022-08-16 at 15 59 24

There's still more than a 1.5rem gap because there's margin on the legend as well as the title padding

boxadesign:
@jrbarnes9 This should be good now

jrbarnes9 (approved):

kruncher (approved):

PR #2408, opened 2022-08-18
Fix: Use semantically correct element for inputs with prefix/suffix

### LINKED ISSUE #2137: Durations should not use `abbr` element if `title` is the same as the visible suffix
Durations should only use the `<abbr>` attribute (with `title`) if the visible text is an abbreviation. The `title` for the `<abbr>` should never be the same as the visible suffix

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/19#discussioncomment-2553032_

jrbarnes9:
Fixes #2137 

- Sets a `span` for the prefix/suffix of an input if no long form `title` for the abbreviated visible text is set.
- Added new macro input tests to check for these
- Added missing Axe tests for input macro
- Updated mutually exclusive JS to not rely on `title` attribute to select prefix/suffix label

### How to review
- Check inputs with prefix/suffix only use the `abbr` element if a `title` is set. Otherwise a `span` should be used.
- Check new tests


boxadesign (approved):

src/components/input/_macro.njk
```diff
@@ -71,7 +71,7 @@
         {% endif %}
 
         {% if params.prefix is defined and params.prefix or params.suffix is defined and params.suffix %}
-            {% if (params.prefix is defined and params.prefix and params.prefix.id is defined and params.prefix.id and params.prefix.title is defined and params.prefix.title) or (params.suffix is defined and params.suffix and params.suffix.id is defined and params.suffix.id and params.suffix.title is defined and params.suffix.title) %}
+            {% if (params.prefix is defined and params.prefix and params.prefix.id is defined and params.prefix.id) or (params.suffix is defined and params.suffix and params.suffix.id is defined and params.suffix.id) %}
```
kruncher, line 74: The `is defined` part isn't needed anymore

kruncher (approved):

boxadesign (approved):

PR #2409, opened 2022-08-18
Fix: Checkboxes - open other tab order

### LINKED ISSUE #2375: Tabbing onwards from visible text input put focus on first input rather than the next focusable element.
Tabbing onwards from visible text input put focus on first input rather than the next focusable element.

Screenshot 2022-08-05 at 12 14 08

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/13#discussioncomment-3332782_

boxadesign:
Fixes #2375 

An event to handle tabindex occurs when the checkbox is blurred which caused this issue. The event should only execute when using radios.

### DIFF
--- a/src/components/radios/check-radios.js
+++ b/src/components/radios/check-radios.js
@@ -3,12 +3,13 @@ export default class CheckRadios {
     this.radio = radio;
     this.openOther = openOther;
     this.input = this.openOther.querySelector('.ons-input');
+    this.input.tabIndex = -1;
 
     this.setInputBlurAttributes();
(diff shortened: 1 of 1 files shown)


jrbarnes9 (commented):
Also causes tabbing issues for radios with visible input...
- tabbing off from the visible input puts focus back on the radio, rather than on to the next focusable element on the page
- after visible input is filled, you can't tab back into it without changing the selected radio

jrbarnes9 (approved):

kruncher (approved):
This seems good to me.

kruncher:
Seems good although there is no test.

Issue #2424, opened 2022-08-30
There is a paste error in the navigation component `params.navigation.subNavigation.sitsiteBasePatheBasePath`


kruncher:
See: https://github.com/ONSdigital/design-system/blame/main/src/components/navigation/_macro.njk#L89

kruncher:
I observe the following properties being used:
- `params.navigation.siteBasePath`
- `params.navigation.subNavigation.siteBasePath`
- `params.navigation.subNavigation.sitsiteBasePatheBasePath` (the typo)

Wouldn't the `siteBasePath` come from `params.navigation.siteBasePath`?

kruncher:
Is it still "by design" to not mark a menu item "active" if it's for the root/home page?

PR #2427, opened 2022-08-31
Add tests and document that `currentPath` of navigation can be an array of strings.

### LINKED ISSUE #2422: Navigation: Current page is not marked for nested sub-navigation
In the following example it is not possible to have both "Guidance" (the parent section) and "ONS sub-branding" (the page) marked as the current page at the same time:

Image

A solution might be to allow an array of strings to be provided for `currentPath` such that multiple navigation items can be marked as current. If a string is provided (as with the current implementation) then the macro would wrap that inside an array so that the macro implementation only needs to deal with arrays.

kruncher:
Resolves #2422, #2424 and #2428

### How to review
- Review documentation.


src/components/header/_macro-options.md
```diff
@@ -44,7 +44,7 @@
 | ariaLabel    | string                | false    | The `aria-label` attribute added to the `<nav>` element. Defaults to “Main menu”.                                                          |
 | itemsList    | array`<Item>`         | true     | Settings for an array of [list items](#item) for each navigation link                                                                      |
 | siteBasePath | string                | false    | If the root path for your site is not `/` (i.e. if using the prototype kit), you can set this path to the home page to set the active page |
-| currentPath  | string                | true     | The path of the current active page                                                                                                        |
+| currentPath  | string or string[]    | true     | The path of the current active page. Multiple paths can be provided using an array to highlight nested navigation.                         |
```
jrbarnes9, line 47: For `params.variants` we document the type as: "array or string"

jrbarnes9 (approved):

jrbarnes9 (approved):

jrbarnes9 (approved):

rmccar:
I think this ability to set multiple levels as the current path in navigation should also be added to the section navigation component and the vertical version example for that should show this working

src/components/navigation/_macro.njk
```diff
@@ -79,14 +79,14 @@
...
                                 {% if item.sections %}
                                     {% for section in item.sections %}
                                         {% if section.sectionTitle %}<h3 class="ons-navigation__list-header">{{ section.sectionTitle }}</h3>{% endif %}
                                         <ul class="ons-navigation__list ons-navigation__list--child ons-list--dashed ons-u-ml-s ons-u-mt-xs">
                                             {% for child in section.children %}
-                                                <li class="ons-navigation__item ons-list__item {{ ' ons-navigation__item--active' if (child.url == params.navigation.subNavigation.currentPath) or (child.url != (params.navigation.subNavigation.sitsiteBasePatheBasePath | default('/')) and child.url in params.navigation.subNavigation.currentPath) }}">
+                                                <li class="ons-navigation__item ons-list__item {{ ' ons-navigation__item--active' if (child.url == params.navigation.currentPath) or (params.navigation.currentPath is not string and child.url in params.navigation.currentPath)}}">
```
rmccar, line 89:
```suggestion
                                                <li class="ons-navigation__item ons-list__item {{ ' ons-navigation__item--active' if (child.url == params.navigation.currentPath) or (params.navigation.currentPath is not string and child.url in params.navigation.currentPath) }}">
```
Really small nit but this space shouldn't be taken out

boxadesign:
> I think this ability to set multiple levels as the current path in navigation should also be added to the section navigation component and the vertical version example for that should show this working

We considered this but currently there is no requirement to do this on section nav

jrbarnes9 (approved):

boxadesign (approved):

PR #2435, opened 2022-09-02
Fix DAC issue convert summaries to use definition lists

rmccar:
This PR resolves a DAC issue with summaries that have multiple items per row. Screen readers have had trouble with row titles being read out correctly. So this change is to convert summaries to use definition lists instead of tables.
To get this to work and align correctly in all instances with flexbox, because theres no `colspan` alternative I have had to change the summary to go to the vertical view on the `m` breakpoint rather than the `s` one.

Also:
- Updates the grouped example to show how rows with multiple items should be configured and fixes the formatting on one example.
- Fixes bug where if there is a `rowtitle` and an `errorMessage` then only the `errorMessage` is shown
- Fixes issue where values are slightly out of alignment with titles
- Fixes a spacing issue where there is a `rowtitle` and a `summaryTitle` but no group title there was too much space between the two titles

### How to review
- Check grouped summary is read out correctly with a screenreader
- Check that summaries still work correctly
- Pass in  a `rowtitle` and an `errorMessage` and see that both are now displayed


rmccar:
[Should be using `groups` for summary of multiple answers to single question](https://trello.com/c/SBGYAYRE/5380-should-be-using-groups-for-summary-of-multiple-answers-to-single-question)

boxadesign:
So the recommendation is to use a definition list in the same way gov.uk do -> https://design-system.service.gov.uk/patterns/check-answers/

I don't think it would take much to convert this. I think if we were not going to prioritise this now the alternative would be to place text inside the problematic tr (What are your monthly household expenses?) and put that in a h4directly after the h3 (Spending). This way it sits outside of the table containing the items in the group.

boxadesign (approved):

jrbarnes9:
Could you run a Percy check?

boxadesign:
I don't think this part of the docs is relevant any longer 
image

boxadesign (approved):

jrbarnes9 (approved):

PR #2441, opened 2022-09-12
Modal - allow GA tracking on events

### LINKED ISSUE #2403: Adding support for GTM to modals
On eQ we need to have GTM events fired for:
- When the modal is opened 
- When the modal is closed
- The type of modal (timeout / regular)
- If it is a timeout one, when the user clicks the button to extend the session

So we would like support added for GTM events for these so this information can be passed to GTM/GA.

boxadesign:
Resolves #2403 

Adds the following GA tracking attributes when the parameter `enableGA: true` is provided:

- `data-ga-action:  Modal closed/opened by _event.type_ event`
- `data-ga-label: _modalType_ modal closed/open`
- `data-ga-category:  _modalType_ modal`

These attributes are set under the following scenarios:

- When the modal opens via timed or click event
- When the modal is closed via timed or click event
- When the modal is closed via escape keypress

### How to review
Add the parameter `enableGA: true` to the timeout modal example and test the attributes are set as per the spec.
The tests cover the scenarios.

### DIFF
--- a/src/components/modal/_macro-options.md
+++ b/src/components/modal/_macro-options.md
@@ -1,8 +1,9 @@
-| Name       | Type   | Required | Description                                                                               |
-| ---------- | ------ | -------- | ----------------------------------------------------------------------------------------- |
-| id         | string | true     | The HTML `id` for the modal’s `<dialog>` element. Defaults to “dialog”.                   |
-| classes    | string | false    | Classes to add to the modal’s `<dialog>` element                                          |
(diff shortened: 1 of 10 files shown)


rmccar:
Looks good, just wondering do we want to add something to the modal docs about GA, like whats on the collapsible page?

rmccar (approved):

jrbarnes9 (approved):

PR #2444, opened 2022-09-15
Search component: allow an accesible icon only version of the button

### LINKED ISSUE #1927: Attached button variant for search
The search icon is ubiquitous, so we probably don't need a label on the search component button, and follow convention and attach the icon-only button to the input

image

Consider adding the DS search example as a compact "buttonless" variant too for uses such as: https://github.com/ONSdigital/dp-design-system/pull/72

boxadesign:
Resolves #1927 

### How to review
Test that the change addresses the requirements.

### DIFF
--- a/src/components/search/_macro-options.md
+++ b/src/components/search/_macro-options.md
@@ -5,7 +5,12 @@
 
 ## searchButton
 
-| Name                   | Type                                   | Required | Description                                                   |
-| ---------------------- | -------------------------------------- | -------- | ------------------------------------------------------------- |
-| Button configuration   | `Button` [_(ref)_](/components/button) | true     | Settings for the button.                                      |
-| visuallyHideButtonText | boolean                                | false    | Will wrap the provided `text` value in a visually hidden span |
+| Name                   | Type    | Required | Description                                                                                                         |
+| ---------------------- | ------- | -------- | ------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


kruncher (approved):

src/components/search/_macro-options.md
```diff
@@ -1,4 +1,11 @@
...
+| Name         | Type                                 | Required | Description                                    |
+| ------------ | ------------------------------------ | -------- | ---------------------------------------------- |
+| input        | `Input` [_(ref)_](/components/input) | true     | Settings for the input.                        |
+| searchButton | object`<searchButton>`               | true     | Settings for the [searchButton](#searchButton) |
+
+## searchButton
+
+| Name                   | Type                                   | Required | Description                                                   |
+| ---------------------- | -------------------------------------- | -------- | ------------------------------------------------------------- |
+| Button configuration   | `Button` [_(ref)_](/components/button) | true     | Settings for the button.                                      |
```
jrbarnes9, line 10: Unless we're importing the whole component with all params, we usually list out here which of the options you can set here, so it's clear what can be changed. Otherwise a user might try to set a param that hasn't been used here, and report the component is broken. So in this case: type, id, classes, attributes, and iconType.

jrbarnes9 (commented):
I would hide the buttonText on the accessible placeholder example too.

kruncher (approved):
Looks good to me

PR #2450, opened 2022-09-21
Fix: Checkboxes and radios with other inner fieldset

boxadesign:
Issue observed with the checkboxes/radios script for implementations that used the nested "other" fieldset containing radios/checkboxes.

The `change` event for radios does not provide `false` (only `true`). This caused problems with deselecting child (inner fieldset) radios and checkboxes.

The component has the following implementation options:

- Checkboxes with nested checkboxes or radios
- Checkboxes with nested checkboxes that can pre select child checkboxes
- Radios with nested checkboxes or radios
- Radios with nested checkboxes that can pre select child checkboxes

These were handled by a single script for both checkboxes and radios. This PR separates the concerns by creating a seperate script for checkboxes and radios. 

### How to review
- Test the scenarios above and importantly that child radios/checkboxes are unchecked when the parent radio/checkbox is unchecked

### DIFF
--- a/src/components/accordion/accordion.spec.js
+++ b/src/components/accordion/accordion.spec.js
@@ -67,7 +67,7 @@ describe('script: accordion', () => {
       }),
     );
 
-    const ariaExpanded = await page.$eval('button[data-test-trigger]', element => element.getAttribute('aria-expanded'));
+    const ariaExpanded = await page.$eval('button[data-test-trigger]', element => element.ribute('aria-expanded'));
     expect(ariaExpanded).toBe('true');
   });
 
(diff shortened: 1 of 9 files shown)


kruncher (commented):
The following behaves differently: http://localhost:3002/components/radios/#radio-with-revealed-checkboxes

1. Checkboxes are all selected by default (originally they were all unchecked by default).

    **Original:**
    image

    **New:**
    image

2. When the checkbox selection is changed; switching to another radio and then back again now re-selects all the checkboxes. Originally it maintained the user's customisation:

    **Original:**
    image
    **New:**
    image

jrbarnes9 (approved):

PR #2451, opened 2022-09-22
Remove `dsExample` parameter from components and introduce `isDesignSystemExample` global.

kruncher:
Remove `dsExample` parameter from components since it is only required to assist with the rendering of component examples and should not be used by services.

### How to review
- Verify documentation still renders properly.
- Sanity check.


boxadesign (approved):

src/components/button/_macro.njk
```diff
@@ -64,7 +64,7 @@
         {% endif %}
     {% endif %}
 
-    {% set tag = "a" if params.url or params.dsExample else "button" %}
```
jrbarnes9, line 67: Do we even need this if it wasn't being used in any examples? It means now that all the button examples are producing invalid HTML by having attributes like `type` on an hyperlinks e.g.
```
<a type="submit" class="ons-btn">
        <span class="ons-btn__inner"><span class="ons-btn__text">Save and continue</span>
            </span>
</a>
```

jrbarnes9: I think we just add `onsubmit="return false;"` or `action="#0"` on all the forms in examples instead

boxadesign (approved):

jrbarnes9 (approved):

boxadesign (approved):

PR #2454, opened 2022-09-22
Remove website generation

kruncher:

This pr removes the website build process from the DS as all documentation now lives in the service manual. 

### How to review
- Check to see if any other logic can be removed from the design system.
- Verify that build/preview commands now shows a listing of examples instead of documentation pages.


src/layout/_template.njk
```diff
@@ -16,9 +16,6 @@
     {% set cdn_url = (pageConfig.cdn.url if pageConfig.cdn and pageConfig.cdn.url) or "https://cdn.ons.gov.uk/sdc/design-system" %}
     {% set slash = "" if cdn_url | last == "/" else "/" %}
     {% set assetsURL = cdn_url + slash + ((pageConfig.cdn.version if pageConfig.cdn and pageConfig.cdn.version) or release_version or designSystemVersion) %}
```
kruncher, line 18: `or release_version` probably isn't needed anymore (unless other services make use of this).

boxadesign: release version is used by EQ, RASRM for setting the DS version in their app.

boxadesign (approved):

jrbarnes9 (approved):

PR #2460, opened 2022-09-27
DS - convert sass color variables to css custom properties

### LINKED ISSUE #2455: Convert colour sass variables to css custom properties
To allow greater flexiblity for services to override colour themes and not have a node/pre-processing requirement we need to convert the colour sass variables to use modern and browser compatible custom css property.

The other benefit is that we won't need to host the theme files inside the DS.

boxadesign:
Resolves #2455


kruncher (dismissed):
Appears to work as I would expect when including a theme overrides stylesheet after the main design system stylesheet. The syntax change also seems good to me.

kruncher (approved):

jrbarnes9 (approved):

PR #2473, opened 2022-10-06
Add initial GA attributes to modal

rmccar:
GA tracking isn't currently working for modals and it looks like its because the attributes aren't immediately available on page load so this will add them initially to the page on load

### How to review
Add `enableGA = true` to the timeout modal example and see that the attributes are now set on page load before being changed when the modal opens and closes


boxadesign:
@rmccar Looks good but will need tests for the changes

jrbarnes9 (approved):

PR #2479, opened 2022-10-06
Fix: Downloadable resources - fix accessibility issues from DAC retest

### LINKED ISSUE #2474: DAC retest - fix issues with downloable resources
The DAC retest failed the downloadable resources pattern on the following:

- Page 17 - Dialog - Screen reader mobile users that activate the ‘Show filters’ button are not trapped to the 
filter dialog that appears. This means that users are able to navigate to the content that is 
behind the dialog. Visual screen reader users may become disorientated without trapping 
the users focus to the dialog.

- Page 19 - Label in name - Screen reader users navigating the page through a mobile device, will hear the button ‘Show 
filters’ announce as ‘Toggle filters’. A visual screen reader user could be confused by the 
different announcement, as it is expected to hear ‘Show filters’. Additionally, ‘Toggle filters’ 
is not descriptive for the purpose of the button as a dialog is displayed rather than content 
toggling. 

- Page 21 - Status message - Changing the state of checkboxes within the filters section of the page results in the text ‘9 
results of 150’ updating to reflect the updating results. The change of results text is not 
audibly read to screen reader users, meaning that the user may be unaware that results 
have updated.

boxadesign:
Fixes #2474 

As a quick fix for the accessibility issues and as we don't view the pattern as production ready the changes are as follows:

- The wrapping filter element is moved outside of `.ons-page` when the show filters button is clicked (on mobile view). The block is appended to `body`. This is toggled.
- `ons-page` is hidden and receieves `aria-hidden= true` when the show filters button is clicked. This is toggled.
- `role=status` has been added to the results count
- Removed `aria-label` from the button as its not required.

### How to review
Check the fixes work in the correct way.


kruncher (approved):

jrbarnes9 (commented):
Not sure what's causing it, but the `select` seems to be listed twice in Voiceover's form controls menu...
PXL_20221010_085747478

jrbarnes9 (approved):

PR #2522, opened 2022-11-04
Fix: Address autosuggest aria status message

### LINKED ISSUE #2367: Focus state is lost when traversing between autosuggest results panel and 'I can't find my address' link
Focus state is lost when traversing between autosuggest results panel and 'I can't find my address' link

autosuggest-focus

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/1387#discussioncomment-3317659_

boxadesign:
Fixes #2367 

The address autosuggest aria status message when the api connection had failed contained a link.

Now it contains just a string for the announced message.

This bug was spotted as focus state attached to the hidden link when tabbing out of the message.

### How to review
Check the message. Check that focus works as it should.


kruncher (approved):
I'm not sure how this fixed the issue; but it seems to be working as expected :)

PR #2523, opened 2022-11-09
Fix: nested input error state

### LINKED ISSUE #2413: Nested inputs are styled as errors when parent wrapped in error panel regardless
When checkbox or radio is wrapped in an error panel, any nested input is styled with a red border even if it not the reason for the error.  This should only be styled as an error if its parent is `checked` or it has the `required` attribute.

Screenshot 2022-08-19 at 10 49 37

Screenshot 2022-08-19 at 10 57 42

boxadesign:
Fixes #2413 

The change is to add an additional check to only style the input as an error if the parent checkbox/radio is checked.

### How to review
Check via the errors prototype

https://deploy-preview-2523--ons-design-system.netlify.app/patterns/correct-errors/examples/errors-proto/

### DIFF
--- a/src/components/checkboxes/_macro.njk
+++ b/src/components/checkboxes/_macro.njk
@@ -79,7 +79,7 @@
                                 "otherType": checkbox.other.otherType,
                                 "options": checkbox.other.options,
                                 "width": checkbox.other.width,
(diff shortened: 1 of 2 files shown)


kruncher (approved):
Looks good to me

PR #2525, opened 2022-11-09
Remove required attribute from from form inputs

### LINKED ISSUE #2524: `required` attribute isn't needed on any inputs
We don't recommend using html5 validation but allow a way for services to use it. So this logic should be removed.

boxadesign:
Resolves #2524 

We don't recommend using the `required` attribute for browser based validation so we shouldn't include the option to use it or reference it in the Design System.

### How to review
Check that all references are removed.


kruncher (dismissed):

src/patterns/help-users-to/correct-errors/examples/errors-proto/errors/index.njk
```diff
@@ -425,7 +423,7 @@ layout: none
                         "checked": true,
                         "value": "pepperoni",
                         "other": {
-                            "otherType": "select",
+                            "otherType": "input",
```
jrbarnes9, line 426: This doesn't match the state of the previous page now. I don't think we even need this example though, to be honest. We're already showing an example of a nested select error above.

jrbarnes9 (dismissed):

jrbarnes9 (approved):

PR #2530, opened 2022-11-15
Fix: Tabs component - check for hash in url update

boxadesign:
Fixes: #2472 

Update logic around initial tab activation when page loads.

### How to review
Check it works as expected and tests pass


kruncher (approved):
Looks good to me. Tested locally with service manual.

jrbarnes9 (approved):

Issue #2510, opened 2022-10-26
Household style questions with other questions in summary


rmccar:
There is a need in eQ to allow household style summary questions to be displayed in the same summary group as other questions. Currently there are a few issues doing this in the DS. 

- Summary link needs to go under the household question not at the bottom of the summary
- If no rows in the question the placeholder text isn't displayed because of the other rows in the summary existing

<img width="913" alt="Screenshot 2022-10-26 at 11 20 49" src="https://user-images.githubusercontent.com/42928680/198002026-4508c6a7-f02e-47f2-a611-61940eb3eef9.png">

PR #2537, opened 2022-11-21
Fix summary style for grouped household style answers

rmccar:
- Fixes answer alignment for grouped "household" style answers
- Adds these type of answers into one of the examples
- Adds some spacing to answers that come after "household" style answers
- Removes headers param from summary test because the param is no longer used

Fixes: #2510 

### How to review
- Check that all types of summary still work and look as expected
- Check that summary answers for "household" style answers are now aligned at all screen sizes

### DIFF
--- a/src/components/summary/_macro.njk
+++ b/src/components/summary/_macro.njk
@@ -112,7 +112,7 @@
                     {% endif %}
 
                     {% if group.summaryLink %}
-                        <div class="ons-summary__link{% if group.placeholderText or group.rows %} ons-u-pt-s{% endif %}{% if group.placeholderText is not defined and group.rows | length > 1 %} ons-u-bt{% endif %}">
+                        <div class="ons-summary__link{% if group.placeholderText or group.rows %} ons-u-pt-s{% endif %}{% if group.placeholderText is not defined and group.rows | length > 1 %} ons-u-bt{% endif %}{% if not group.last %} ons-u-mb-xl{% endif %}">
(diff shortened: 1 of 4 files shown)


rmccar:
[Fix alignment in DS for grouped list collector answers](https://trello.com/c/PlhYTpdF/5481-fix-alignment-in-ds-for-grouped-list-collector-answers)

src/components/summary/_macro.njk
```diff
@@ -112,7 +112,7 @@
                     {% endif %}
 
                     {% if group.summaryLink %}
-                        <div class="ons-summary__link{% if group.placeholderText or group.rows %} ons-u-pt-s{% endif %}{% if group.placeholderText is not defined and group.rows | length > 1 %} ons-u-bt{% endif %}">
+                        <div class="ons-summary__link{% if group.placeholderText or group.rows %} ons-u-pt-s{% endif %}{% if group.placeholderText is not defined and group.rows | length > 1 %} ons-u-bt{% endif %}{% if not group.last %} ons-u-mb-xl{% endif %}">
```
jrbarnes9, line 115: Should probably add a test to `_macro.spec.js` for the addition of this class

jrbarnes9 (commented):
Actions not right-aligned in Safari...
Screenshot 2022-11-30 at 10 58 37

rmccar:
> Actions not right-aligned in Safari... Screenshot 2022-11-30 at 10 58 37

I noticed this in Percy myself but I cant recreate it locally in Safari, is it possible that Percy is using an older version?
It also seems to be an issue on the `main` branch and not just on my PR branch so is probably its own issue out of scope of this PR

PR #2538, opened 2022-11-22
Fix header issues on small screens

rmccar:
Fixes an issue with the DS header on small screens.
Also removes the `ons-patternlib-page__header` class being set because it isn't used.

There is this wrapping at about 405px:
Screenshot 2022-11-22 at 11 48 49

### How to review
See that these issues are resolved at the respective screen widths

### DIFF
--- a/src/components/label/_label.scss
+++ b/src/components/label/_label.scss
@@ -17,7 +17,7 @@
   }
 
   &--placeholder {
-    font-size: 1rem;
+    font-size: 0.95rem;
     font-weight: $font-weight-regular;
     left: 10px;
     position: absolute;
--- a/src/components/label/_label.scss
+++ b/src/components/label/_label.scss
@@ -17,7 +17,7 @@
   }
 
   &--placeholder {
-    font-size: 0.95rem;
+    font-size: 1rem;
     font-weight: $font-weight-regular;
     left: 10px;
     position: absolute;
--- a/src/components/navigation/_macro.spec.js
+++ b/src/components/navigation/_macro.spec.js
@@ -71,7 +71,7 @@ const PARAMS = {
 };
 
 const SITE_SEARCH_AUTOSUGGEST = {
-  label: 'Search design system',
+  label: 'Search the design system',
   instructions:
     "Use up and down keys to navigate results once you've typed more than two characters. Use the enter key to select a result. Touch device users, explore by touch or with swipe gestures.",
   ariaYouHaveSelected: 'You have selected',
@@ -133,7 +133,7 @@ describe('macro: navigation', () => {
         containerClasses: 'ons-autosuggest-input--header',
         classes: 'ons-input-search ons-input-search--icon',
         label: {
-          text: 'Search design system',
+          text: 'Search the design system',
           id: 'ons-site-search-label',
           classes: 'ons-u-pl-m ons-label--white',
         },
--- a/src/components/navigation/_macro.spec.js
+++ b/src/components/navigation/_macro.spec.js
@@ -71,7 +71,7 @@ const PARAMS = {
 };
 
 const SITE_SEARCH_AUTOSUGGEST = {
-  label: 'Search the design system',
+  label: 'Search design system',
   instructions:
     "Use up and down keys to navigate results once you've typed more than two characters. Use the enter key to select a result. Touch device users, explore by touch or with swipe gestures.",
   ariaYouHaveSelected: 'You have selected',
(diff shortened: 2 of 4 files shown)


src/components/navigation/_macro.spec.js
```diff
@@ -71,7 +71,7 @@ const PARAMS = {
 };
 
 const SITE_SEARCH_AUTOSUGGEST = {
-  label: 'Search the design system',
+  label: 'Search design system',
```
boxadesign, line 74: I think we should keep it with the "the" as thats the task

src/components/label/_label.scss
```diff
@@ -17,7 +17,7 @@
   }
 
   &--placeholder {
-    font-size: 1rem;
+    font-size: 0.95rem;
```
boxadesign, line 20: I don't think we need to do this as we don't support below 325px

rmccar: I spoke to Jon about this because I wasn't sure what the minimum we support was, he said its usually 375px because that was the smallest smartphone but the Galaxy Fold is now only 280px

src/components/navigation/_macro.spec.js
rmccar: I could make the placeholder slightly smaller and add the "the" back in if we do decide we want to support the Galaxy Fold?

src/components/label/_label.scss
boxadesign: If we support 280px then loads of things will break as our `xxs` breakpoint is 300px currently. I don't think we should support this size as yet as its not a mainstream thing.

rmccar: Ok fair enough, probably best have a chat with Jon, if you guys decide to not support anything below that then let me know and Ill remove this change from the PR

jrbarnes9: OK, let's consider/deal with smaller viewport support in a separate issue, because like Matt says, we'd need to adjust the smallest breakpoints and change much more to fully support it.
We'd probably need to make the assessment based services' analytics.

src/components/navigation/_macro.spec.js
jrbarnes9: Yeah keep it in for now, i'll raise a ticket to address all the work required to support small viewports.

jrbarnes9:
I'm not seeing this wrapping in Chrome. What browser are you seeing it in? 
Screenshot 2022-11-23 at 15 11 55

rmccar:
> I'm not seeing this wrapping in Chrome. What browser are you seeing it in? Screenshot 2022-11-23 at 15 11 55

In chrome too. Just had a look and it seems to only be a problem if you resize the screen without going into mobile view. I wasn't using the mobile view I was just resizing the screen so was getting the wrapping

jrbarnes9:
> > I'm not seeing this wrapping in Chrome. What browser are you seeing it in? 
> 
> In chrome too. Just had a look and it seems to only be a problem if you resize the screen without going into mobile view. I wasn't using the mobile view I was just resizing the screen so was getting the wrapping

Strange, I'm still not seeing it.
Screenshot 2022-11-24 at 17 35 40

rmccar:
> > > I'm not seeing this wrapping in Chrome. What browser are you seeing it in?
> > 
> > 
> > In chrome too. Just had a look and it seems to only be a problem if you resize the screen without going into mobile view. I wasn't using the mobile view I was just resizing the screen so was getting the wrapping
> 
> Strange, I'm still not seeing it. Screenshot 2022-11-24 at 17 35 40

Just noticed that I only seem to see it on a 4k monitor and not on a 1080p one

jrbarnes9 (dismissed):

jrbarnes9 (commented):
I'm not convinced we should be altering the component for this very specific scenario only found in the design system website header. 

Maybe it should instead be adding an inline style adjustment in the website's header.

rmccar:
> I'm not convinced we should be altering the component for this very specific scenario only found in the design system website header.
> 
> Maybe it should instead be adding an inline style adjustment in the website's header.

Yeah fair enough although it could be an issue for any service that has a name the same length or longer then "Design System" but that could be hard to account for without knowing the length

rmccar:
> I'm not convinced we should be altering the component for this very specific scenario only found in the design system website header.
> 
> Maybe it should instead be adding an inline style adjustment in the website's header.

Ive changed it so it only affect the DS home header

boxadesign:
A service could be called anything so we need it to be able to wrap. "Design system" will soon be "Service manual" so I don't think we should spend time on this and just close the issue and this pr

rmccar:
> A service could be called anything so we need it to be able to wrap. "Design system" will soon be "Service manual" so I don't think we should spend time on this and just close the issue and this pr

Ok fair enough, its a pretty minimal issue anyway

PR #2541, opened 2022-11-24
Video only load when cookies enabled

### LINKED ISSUE #2536: Change the way video embed works for youtube/vimeo
Currently we don't prevent cookies being set by youtube when the user does not enable them.

The solution to this is to what gov.uk do:

1 - By default youtube cookies are not set so we should just display a link to the video on youtube.com
2 - If the user allows youtube cookies a script should run when a video is available on a page, on page load to embed the video (as we use static sites we can't do this server side so will need to be done by the client).
3 - We should always use the domain youtube-nocookies.com so user preferences are not interferred with when viewing videos on ons services.
4- do the above for vimeo embeds also

boxadesign:

Resolves #2536 

Video embeds should not load if the "campaigns" cookie is not enabled.

The component now takes an image and link text as additional properties. The image and text is shown by default if cookies are not enabled.

If the cookies are enabled, the page will hide the placeholder content and load the iframe (shows and sets the src).

### How to review
Load the page, observe that the placeholder content is shown.
Go to the cookies settings pattern page and accept the cookies, navigate to the video component and confirm that the video loads.


src/components/video/examples/video/index.njk
```diff
@@ -1,7 +1,14 @@
...
-        "title": "Census 2021 promotional TV advert"
+        "videoEmbedUrl": 'https://www.youtube.com/embed/_EGJlvkgbPo',
+        "videoLinkURL": 'https://www.youtube.com/watch?v=_EGJlvkgbPo',
+        "image": {
+            "largeSrc": '/patternlib-img/youtube-example-cover.png',
+            "smallSrc": '/patternlib-img/youtube-example-cover.png',
+            "alt": "Youtube video"
+        },
+        "title": 'Census 2021 promotional TV advert',
+        "linkText": 'Watch "Census 2021 promotional TV" on Youtube'
```
jrbarnes9, line 12:
```suggestion
        "linkText": 'Watch “Census 2021 promotional TV advert” on Youtube'
```

jrbarnes9 (commented):
Placeholder image is not as wide as the iframe, or the link and focus state.
Screenshot 2022-11-28 at 14 03 26
Screenshot 2022-11-28 at 14 11 48

jrbarnes9 (approved):

kruncher (approved):

jrbarnes9 (approved):

PR #2544, opened 2022-11-25
Fix page config welsh footer logo

rmccar:
The welsh footer logo isn't being displayed when using the footer component through page config. This PR fixes that and updates the example to show that.

Also removes the language param which isn't being used

### How to review
Check help users to change language example and see that the welsh ONS logo is now displayed


rmccar:
[Footer not displaying Welsh ONS logo](https://trello.com/c/CYYd2bv6/5485-footer-not-displaying-welsh-ons-logo)

jrbarnes9 (approved):

PR #2548, opened 2022-11-30
Change onsCollapsible to onsDetails, rename css classes

### LINKED ISSUE #2531: "Collapsible" does not accurately describe the component
Considering its default state is "collapsed", it's really an "expandable". However, after discussing with content designer, we think the component should be renamed to "Details", in line with html element it uses, and its use elsewhere in government design systems. This will also help differentiate it from the accordion component.

boxadesign:
Resolves #2531 

This is a NON breaking nunjucks change. Mark up has changed.

- onsCollapsible is now onsDetails
- Component page is now found under /components/details
- `collapsible` class is renamed `details`
- Existing onsCollapsible is still available (calls onsDetails under the hood)
- Tests retained for onsCollpasible usage
- onsCollapsible will be removed after an agreed timeframe

### How to review
Check all details and accordion examples work and tests pass


kruncher (dismissed):
This seems good to me

rmccar:
Looks good but thinking we should show the details component in the list when you search for "collapsible" in the DS site search box

kruncher (approved):

PR #2566, opened 2022-12-23
Fix: Pagination - update and simplify logic

### LINKED ISSUE #2565: Pagination logic isn't consistent or simple to document
The logic for the long pagination component seems to ensure – where applicable – that the user can see and select two pages before and two pages after the [current] selected page, as well as the first page and the last page.

For example:
> - [1] 2 3 … 10*
> - 1 [2] 3 4 … 10
> - 1 2 [3] 4 5 … 10*
> - 1 2 3 [4] 5 6 … 10
> - 1 ... 3 4 [5] 6 7 … 10
> - 1 … 4 5 [6] 7 8 ... 10
> - 1 … 5 6 [7] 8 9 10
> - 1 … 6 7 [8] 9 10
> - 1 … 7 8 [9] 10
> - 1 … 8 9 [10]*

But there are a few scenarios(*) in the macro that currently don't seem to fit this rule:
Screenshot 2022-12-21 at 16 14 27
Screenshot 2022-12-21 at 16 15 02
Screenshot 2022-12-21 at 17 51 58

boxadesign:
Fixes #2565 

The logic has been changed to address the issues in the linked ticket and also to display the pagination items more inline with GDS (only show one prev/next page)

### How to review
Check that the logic works for all selected page items

### DIFF
--- a/src/components/pagination/_macro.njk
+++ b/src/components/pagination/_macro.njk
@@ -20,22 +20,21 @@
                   <a href="{{ params.pages[prevPageIndex].url }}" class="ons-pagination__link" rel="prev" aria-label="Go to the previous page (Page {{ currentPageIndex - 1 }})">{{ params.previous }}</a>
               </li>
           {% endif %}
+          {% if currentPageIndex > 2 %}
             <li class="ons-pagination__item{% if firstPage.current == true %} ons-pagination__item--current{% endif %}">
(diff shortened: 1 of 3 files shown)


jrbarnes9:
Added #2567 for adding first/last examples to documentation

jrbarnes9 (approved):

kruncher (approved):

PR #2571, opened 2023-01-06
Move breadcrumbs in blocks exploded view example

### LINKED ISSUE #2542: Page template blocks exploded view example shows breadcrumbs within `preMain`, but they are placed before that when using page config
Page template blocks exploded view example shows breadcrumbs within `preMain`, but they are placed before that when using page config

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/195#discussioncomment-4229086_

boxadesign:
Fixes #2542

### How to review
Check the exploded view example in foundations / page template

### DIFF
--- a/src/foundations/layout/page-template/examples/base-page-template-block-areas/index.njk
+++ b/src/foundations/layout/page-template/examples/base-page-template-block-areas/index.njk
@@ -111,9 +111,6 @@ layout: example
 
     <div class="ons-pattern-lib-block">
         <div class="ons-pattern-lib-block__label">block: pageContent</div>
-
-        <div class="ons-pattern-lib-block">
-            <div class="ons-pattern-lib-block__label">block: preMain</div>
             {{
                 onsBreadcrumbs({
                     "ariaLabel": 'Breadcrumbs',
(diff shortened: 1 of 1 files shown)


jrbarnes9:
I'm not clear now what the intention of the exploded block view is for. As well as showing which blocks are available, is it also to show which blocks you should use to place certain components if you're not using `pageConfig`? 

If that's the case, would expect users to use `preMain` block for the breadcrumbs?

jrbarnes9:
> > I'm not clear now what the intention of the exploded block view is.
> > As well as showing which blocks are available, is it also to show which blocks you should use to place certain components if you're not using `pageConfig`?
> > If that's the case, would we expect users to use `preMain` block for the breadcrumbs?
> 
> OK, so are we saying that we want breadcrumbs to be the default content set inside `preMain`? That makes sense if that block isn't known to be used for anything else.
> 
> You would still set it obviously via pageConfig

Yep. Makes sense to me.

kruncher (dismissed):

jrbarnes9 (approved):

kruncher (approved):

PR #2581, opened 2023-01-16
Decouple button link from button component to replace existing call to action component

boxadesign:
WIP

### How to review


src/components/call-to-action/index.njk
```diff
@@ -6,20 +6,43 @@ synonyms: cta
 +++
 {% from "views/partials/example/_macro.njk" import patternlibExample %}
 {% from "components/external-link/_macro.njk" import onsExternalLink %}
-{% from "components/panel/_macro.njk" import onsPanel %}
 +++
 
+
```
jrbarnes9, line 11: Need to replicate this in Craft

src/components/call-to-action/index.njk
```diff
@@ -6,20 +6,43 @@ synonyms: cta
...
+Setting `"variants"` with `'ghost'` will style the button, and setting `iconType` with `exit` adds the appropriate icon for a “sign out” or “exit” button.
+{{
+    patternlibExample({
+        "path": "components/call-to-action/examples/button-ghost/index.njk",
+        "backgroundColor": "#206095"
     })
 }}
 
-## Default
+### Download
```
jrbarnes9, line 39: This one is an awkward fit. Because I don't think users will conceptualise downloading as file navigating away from the current page.

boxadesign: So this is an issue of decoupling and naming I think. A download link and even a save and sign out button isn't a CTA. 

Perhaps everything in the docs could live under "button" but the code is split and we just call it 'link-button'

boxadesign: I don't think the code and docs need to be the same. In the /button/ folder we could just have `_macro.njk` and `_link-button-macro.njk`

jrbarnes9: In the short term let's keep the docs in 'Button'. I'm going to look at redesigning CTAs/link-buttons to be more "materially honest". Maybe call this macro `_action-link.njk` as the design will likely deviate away from looking like a button.

PR #2618, opened 2023-02-16
Added `novalidate` to form config

### LINKED ISSUE #2240: Add `novalidate` to form in page template
All forms should really use server-side validation so we can control the errors presented to the users. We don't explicitly document that client-side validation should not be used.

Currently we rely on a service adding `novalidate` to their forms. We could put this attribute as a default on the page template form so it doesn't need to be manually set.  ~Then provide an override parameter to remove it if a service really needs to.~

jrbarnes9:
Fixes #2240 

To help users prevent the browser validating inputs such as `number` and `email`, added `novalidate` as a hardcoded attribute to the base page template `form` config.

Browser validation should be avoided because:
- the visual style, placement and content of HTML5 error messages cannot be made consistent with the ONS Design System
- we know that the ONS Design System error details and error summary components are accessible

We will also add guidance to the 'input' component and 'correct errors' pattern in a separate ticket telling users why they should add the attribute if not using base page template.

---
**Note:**  Developers already setting `novalidate` with `form.attributes` in the base page template will need to remove the attribute to avoid invalid HTML with the duplicate attributes.

---

### How to review
Check attribute is applied by default when using `form` config in base page template.

### DIFF
--- a/__snapshots__/layout/_template.spec.js.snap
+++ b/__snapshots__/layout/_template.spec.js.snap
@@ -636,7 +636,7 @@ exports[`base page template matches the full configuration snapshot 1`] = `
             <div class=\\"ons-page\\">
                 <div class=\\"ons-page__content\\">
                     
-                        <form class=\\"form-class\\" method=\\"POST\\" autocomplete=\\"off\\" novalidate=\\"null\\">
+                        <form class=\\"form-class\\" method=\\"POST\\" novalidate=\\"\\" autocomplete=\\"off\\" action=\\"/example.html\\">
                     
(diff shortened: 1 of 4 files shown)


boxadesign (approved):
I don't think we should state this a breaking change IMO.

rmccar:
I think we also need to remove `novalidate` being set in the errors proto. Other than that looks good

rmccar (approved):

boxadesign (approved):

jrbarnes9:
Outdated by #2634

Issue #2623, opened 2023-02-28
New utility class for hiding elements from print view from the service side


rmccar:
There is a need on eQ to be able to hide elements from the print view from the service side. Adding a utility class for this would allow more control over what is hidden from the print view from the service side.

PR #2624, opened 2023-02-28
Add print hidden utility class

rmccar:
On the eQ service there is a need to set some elements to be hidden from printing so this PR creates a new utility class (`ons-u-ph`) that will allow this control from the service side.

Also removes the setting of the `--ons-color-black` var as this is not needed.

Fixes: #2623 

### How to review
Add new class to element to see that it is now hidden from printing

### DIFF
--- a/src/scss/print.scss
+++ b/src/scss/print.scss
@@ -8,7 +8,8 @@ html {
 .ons-footer,
 .ons-cookies-banner,
 .ons-language-links,
-.ons-breadcrumb {
+.ons-breadcrumb,
+.ons-u-ph {
   display: none !important;
 }
 
--- a/src/scss/print.scss
+++ b/src/scss/print.scss
@@ -1,7 +1,3 @@
-html {
-  --ons-color-black: #222;
-}
-
 .ons-btn,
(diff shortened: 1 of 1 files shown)


rmccar:
[DS changes for printing bugs](https://trello.com/c/Va7VImc1/5607-ds-changes-for-printing-bugs)

ONS-Guilhem-Forey (dismissed):
Tested ✅

jrbarnes9 (commented):
I think this would probably be better located in `/utilities` (probably in `_visibility.scss`) so it can be found along with all the other utility classes and extended in the print scss?

rmccar:
> I think this would probably be better located in `/utilities` (probably in `_visibility.scss`) so it can be found along with all the other utility classes and extended in the print scss?

I get what you're saying but to do that because of the way the css is structured I would need to import the vars, helpers and visibility css files into the print css, which feels like a lot to get that working to me. Really I think we should document the utility classes somewhere instead.

jrbarnes9 (approved):

boxadesign (approved):

Issue #2625, opened 2023-03-01
When printing pages elements are broken by page breaks


rmccar:
Currently when printing a page there is a chance that page breaks could happen within an element when the content is longer than a page

<img width="597" alt="image (1)" src="https://user-images.githubusercontent.com/42928680/222182490-822d18d7-433d-4298-aaa8-9671f72318e3.png">
<img width="615" alt="image (2)" src="https://user-images.githubusercontent.com/42928680/222182501-02deb697-dd5f-4cf2-951b-979c2702c856.png">

PR #2626, opened 2023-03-01
Fix printing page breaking inside elements

rmccar:
Fixes: #2625 

Also updates some css that was being repeated when there was a mixin available to use

### How to review
Create a page with content longer than a page of A4 including long panels and test the comparison between printing a page on the main branch with this one and see that panels or paragraphs aren't broken by page breaks

### DIFF
--- a/src/components/header/_header.scss
+++ b/src/components/header/_header.scss
@@ -1,8 +1,8 @@
 // Block
 .ons-header {
+  @include font-smoothing;
+
   display: block;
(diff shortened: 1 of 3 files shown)


rmccar:
[DS changes for printing bugs](https://trello.com/c/Va7VImc1/5607-ds-changes-for-printing-bugs)

jrbarnes9 (approved):

boxadesign (approved):

PR #2633, opened 2023-03-09
Add extra functionality to onsQuestion

rmccar:
- Adds the ability to set a warning on a question
- Adds the ability to set guidance, definition or instruction if `legendIsQuestionTitle` is set
- Adds the ability to set id and custom attributes on guidance
- Removes unused imports
- Updates browserlist to remove caniuse lite warning
- Adds ability to set heading and content attributes on question definitions

### How to review
- Add a warning to a question and check that it renders correctly


boxadesign (dismissed):
LGTM

src/patterns/question/example-question-anatomy.njk
```diff
@@ -61,6 +61,22 @@ noTabs: true
...
+                {% call onsPanel({
+                    "type": "warn"
+                }) %}
+                    <p>This question is important</p>
+                {% endcall %}
+            </td>
+            <th class="ons-pl-question-anatomy__description" scope="col">
+                <h2 class="ons-u-fs-r ons-u-mb-no">Question warning <span class="ons-u-fs-s">(optional)</span></h2>
+                <p><code>warning</code></p>
+                <p class="ons-u-fs-s ons-u-mb-no">Only to be used to add emphasis to the question</p>
```
jrbarnes9, line 76:
```suggestion
                <p class="ons-u-fs-s ons-u-mb-no">Only to be used to warn users of a negative consequence to answering the question incorrectly</p>
```

src/patterns/question/example-question-anatomy.njk
```diff
@@ -61,6 +61,22 @@ noTabs: true
             </th>
         </tr>
     </tbody>
+    <tbody class="ons-pl-question-anatomy__row">
+        <tr>
+            <td class="ons-pl-question-anatomy__component">
+                {% call onsPanel({
+                    "type": "warn"
+                }) %}
+                    <p>This question is important</p>
```
jrbarnes9, line 70: To avoid users thinking you can use this just to mark a question as 'important' or 'mandatory', I would just change this to an obvious placeholder, or change the whole example used to match a real use case where all elements would be used.
```suggestion
                    <p>This is a warning notice</p>
```

jrbarnes9 (commented):
This new element needs to be documented, so you'll need to add a new issue ticket in the service manual repo and add it as a dependency: https://github.com/ONSdigital/ons-service-manual-website/issues. Otherwise the documentation will get out of sync with the macros and examples.

jrbarnes9 (approved):

boxadesign (approved):

Issue #1714, opened 2021-10-04
Search component input height too tall on xs/s bp


jrbarnes9:
Search component input height too tall on xs/s bp

![Screenshot 2021-10-04 at 09 52 49](https://user-images.githubusercontent.com/43346934/135822176-be1e6ddb-edef-4319-b0c7-30bff2d217b1.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/656#discussioncomment-1423608_

PR #2659, opened 2023-05-23
fix: search button margin top

alec-hamilton-ons:
A `margin-top` was being added to the `.ons-search__btn` class that was making it look incorrect.

#1714 

### How to review

1. Load Netlify preview of the component on the PR.
2. Check component behaviour for example search with character check.

### DIFF
--- a/src/components/search/_search.scss
+++ b/src/components/search/_search.scss
@@ -1,16 +1,9 @@
 .ons-search-component {
   display: flex;
   flex-flow: row wrap;
+  gap: 0.5rem;
 
   @include mq(s) {
     flex-flow: row nowrap !important;
-  }
-  .ons-search__btn {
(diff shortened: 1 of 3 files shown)


src/components/search/_search.scss
```diff
@@ -6,7 +6,7 @@
     flex-flow: row nowrap !important;
   }
   .ons-search__btn {
-    margin-top: -3px;
```
willcAND, line 9: was this css property not taking effect then?

alec-hamilton-ons: Yes - It was being overridden by the default margin styles being applied on line 12

willcAND (approved):

alessioventuriniAND (approved):

willcAND (approved):

Issue #2652, opened 2023-05-12
Radios becoming misaligned when zooming page


rmccar:
When the page is zoomed in or out the radio buttons are becoming misaligned as in the screenshot. Seems to be down to the way the radios are styled and the spacing isn't compatible with zooming in the browser. Tested on Chrome and Edge and seems to be an issue on both.

![Screenshot 2023-05-12 at 10 18 12](https://github.com/ONSdigital/design-system/assets/42928680/6cf9e3be-438d-4cf2-b68e-ae32d11af257)

PR #2666, opened 2023-06-01
fix: changed how radio buttons are displayed

alec-hamilton-ons:

Selected radio buttons were appearing off center at various window zoom sizes. I have implemented a fix that is not 100% perfect but much better than previous. The error is due to the way browsers calculate viewport units compared with DOM units.

#2652 

### How to review
Look at netlify and zoom in and out to see if the radio buttons behave appropriately.


alessioventuriniAND (approved):
looks perfect to me. The black dot is always centred now at every breakpoint.

precious-onyenaucheya-ons (approved):

Issue #2679, opened 2023-06-14
The accessible names of the "Change" buttons on the summary component should be generated from visually hidden text, not `aria-label`


emilymaylawes:
As discussed in [#279](https://github.com/ONSdigital/ons-service-manual-website/issues/279), `aria-label` is being used to generate the accessible name of the 'Change' button. It is preferred to use visually hidden text instead to add a more contextual expansion of the accessible name of these buttons so that screen reader users will know exactly what the button will do.

`aria-label` runs the risk of failing [2.5.3 Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html) and is also [less supported by assistive technology](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/).

Instead of using the attribute `ariaLabel`, replace with an attribute such as `visuallyHidden` which will add a visually hidden span at the end of the button's label. This is likely a breaking change for existing products using these attributes.

PR #2682, opened 2023-06-19
Enhancement/2679/accessible names of change buttons in summary component

alec-hamilton-ons:
###BREAKING CHANGES
This PR removes the `aria-label` attribute from the `<a>` element. If you are using the `aria-label` for anything, this will no longer be possible. 

#2679 

The "Change" button in the Summary component was using an aria-label for screen readers. I have removed this functionality and added a visually hidden span element so screen readers will now read out "change answer for [question]"

### How to review
Go to the summary component examples and check that a screen reader says the above when you tab to the change button.


precious-onyenaucheya-ons (approved):

alessioventuriniAND (approved):
This is based on the creation of a service manual ticket to update the summary component documentation around accessibility.

PR #2683, opened 2023-06-19
remove value attribute in select tag

### LINKED ISSUE #2202: Invalid parameter `value` on rendered `<select>` element in `select` component
When a `value` parameter is provided for the `select` component the `<select>` element is rendered with a `value` attribute:
```nunjucks
{% if params.value is defined and params.value %}value="{{ params.value}}" {% endif %}
```
This is not valid HTML markup:

Image

precious-onyenaucheya-ons:
fixes #2202 
removed invalid value attribute on rendered select element in select component.

### How to review
Check if a value attribute is rendered with the select element within the select component.

This could be a BREAKING CHANGE if users have previously used the value attribute within the select element.


alec-hamilton-ons (approved):
Looks like these will be breaking changes. Can you update the PR description with instructions similarly to previous breaking change tickets?

Other than that looks all good!

Issue #2084, opened 2022-03-24
[Bug]: Outputs invalid classes when optional parameters `termCol` and `descriptionCol` are not provided to metadata component


kruncher:
Either a default value should be assumed for each of these parameters when unspecified; or no column class should be output.

**Invalid Output:** `ons-col-@m`

alec-hamilton-ons:
Closing issue following discussion with @rmccar & @alessioventuriniAND. The params mentioned are not optional so the changes are not necessary.

PR #2692, opened 2023-06-26
fix: made termCol & descriptionCol optional

alec-hamilton-ons:
#2084 

### How to review
Pull branch, remove `termCol` & `descriptionCol` from example file. Inspect in dev tools to see that no invalid classes are output.

### DIFF
--- a/src/components/metadata/_macro.njk
+++ b/src/components/metadata/_macro.njk
@@ -2,11 +2,11 @@
   <dl class="ons-metadata ons-metadata__list ons-grid ons-grid--gutterless ons-u-cf{{ (' ' + params.classes) if params.classes }}"{% if params.id %} id="{{ params.id }}" {% endif %}{% if params.metadataLabel %} title="{{ params.metadataLabel }}" aria-label="{{ params.metadataLabel }}"{% endif %}>
     {% for item in params.itemsList %}
       {% if item.term | length %}
-          <dt class="ons-metadata__term ons-grid__col ons-col-{{ params.termCol }}@m">{{ item.term }}</dt>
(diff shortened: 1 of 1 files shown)


rmccar (commented):
Technically this isn't an issue as those two params are down as required in the macro options. So maybe we should not do this and close the issue? Might need a discussion on whether we want to make this change?

alec-hamilton-ons:
> Technically this isn't an issue as those two params are down as required in the macro options. So maybe we should not do this and close the issue? Might need a discussion on whether we want to make this change?

I did notice this yes. My thoughts were that if for some reason they don't add the params, then it won't output anything invalid, which seems to me like a good thing (even if it is very unlikely). 

Another option would be to make the params optional.

rmccar:
> > Technically this isn't an issue as those two params are down as required in the macro options. So maybe we should not do this and close the issue? Might need a discussion on whether we want to make this change?
> 
> I did notice this yes. My thoughts were that if for some reason they don't add the params, then it won't output anything invalid, which seems to me like a good thing (even if it is very unlikely).
> 
> Another option would be to make the params optional.

Just tested not setting them and 

> > Technically this isn't an issue as those two params are down as required in the macro options. So maybe we should not do this and close the issue? Might need a discussion on whether we want to make this change?
> 
> I did notice this yes. My thoughts were that if for some reason they don't add the params, then it won't output anything invalid, which seems to me like a good thing (even if it is very unlikely).
> 
> Another option would be to make the params optional.

Screenshot 2023-06-26 at 13 18 51

Removing these params breaks the rendering of the component. So if we were to make them optional we would need to agree on defaults. I don't think this is a change we want to make personally.

Need to avoid `&` in branch names too because it makes it more difficult to check them out

alec-hamilton-ons:
Alright, following discussion here and seeing as the params mentioned are not optional, I am going to close this PR and ticket as the changes do not appear necessary.

Issue #2197, opened 2022-05-19
Document list component - too many conditional statements


boxadesign:
```
                                    {%- if document.metadata.date is defined and document.metadata.date %}
                                        <li class="ons-document-list__item-attribute">
                                            {% if document.metadata.date -%}
                                                {% set prefixClasses = "ons-u-fw-b" if document.metadata.date.showPrefix is defined and document.metadata.date.showPrefix == true else "ons-u-vh" %}
                                                <span class="{{ prefixClasses }}">{{ document.metadata.date.prefix | default("Published") }}: </span>
                                            {%- endif -%}
                                            {% if document.metadata.date.iso is defined and document.metadata.date.iso -%}
                                                <time datetime="{{ document.metadata.date.iso }}">{{ document.metadata.date.short }}</time>
                                            {%- endif %}
                                        </li>
                                    {% endif %}
```

Above we're checking ` {%- if document.metadata.date is defined and document.metadata.date %}` at the top level and then inside this statement. We're also checking for other "required" parameters such as `document.metadata.date.iso`

Issue #2671, opened 2023-06-13
Remove Python Specific Syntax - 2


willcAND:
***NEEDS VALIDATING***

Across the codebase we are using the python specific syntax seen below.

One example would be inside the `select` component.

`{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}`

Inside this same component we are using the following shorthand version. Could we remove the above line in favour of the shorthand?

`{% for attribute, value in option.attributes %}{{ attribute }}="{{ value }}"{% endfor %}`

Need to test this to see if it causes any issues inside a python application.

yatesn:
May need @rmccar support?

willcAND:
Will cause multiple breaking changes across 90% of the components - added low prio label.

daniellecorke:
Tasks for this ticket:

1. Investigate if making this change breaks anything in Python. 
2. Discuss with the team to decide whether further works/tickets are required. 
3. If so, create tickets to outline the work required.

PR #2721, opened 2023-07-13
improved logic for document list component 

### LINKED ISSUE #2196: Document list component - Featured item config and logic improvements
For a document list item to be set as "featured" there is the property `featured: true` this adds a class that handles the styling of that item. There is also the optional parameters `fullWidth: bool` and `wide: bool`. 

The documentation and macro options state that these should be used only when `featured: true`. Also, the check for `wide` is only made inside the check for `fullWidth`. You can set `fullWidth` without setting `featured` which would break layout.

This could be improved by creating something like a `featured` object (that also accepts a boolean value) and then nest the two additional properties and use better key naming such as `fullBleedBackground: bool` and `wideContainer1280: bool`  or something to that effect.

precious-onyenaucheya-ons:
fixes #2196 and #2197

### How to review

### DIFF
--- a/src/components/document-list/_macro.njk
+++ b/src/components/document-list/_macro.njk
@@ -2,9 +2,9 @@
     <ul{% if params.id %} id="{{ params.id }}"{% endif %} class="ons-document-list{{ ' ' + params.classes if params.classes else '' }}"{% if params.attributes %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %} {{attribute}}="{{value}}"{% endfor %}{% endif %}>
         {% for document in params.documents %}
 
-        <li class="ons-document-list__item{{ ' ons-document-list__item--featured' if document.featured }}{{ ' ons-document-list__item--full-width' if document.fullWidth == true }}{{ ' ' + document.classes if document.classes else '' }}"{% if document.attributes %}{% for attribute, value in (document.attributes.items() if document.attributes is mapping and document.attributes.items else document.attributes) %} {{attribute}}="{{value}}"{% endfor %}{% endif %}>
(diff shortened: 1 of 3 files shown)


willcAND (dismissed):

src/components/document-list/_macro.njk
```diff
@@ -2,9 +2,9 @@
...
         {% for document in params.documents %}
 
-        <li class="ons-document-list__item{{ ' ons-document-list__item--featured' if document.featured }}{{ ' ons-document-list__item--full-width' if document.fullWidth == true }}{{ ' ' + document.classes if document.classes else '' }}"{% if document.attributes %}{% for attribute, value in (document.attributes.items() if document.attributes is mapping and document.attributes.items else document.attributes) %} {{attribute}}="{{value}}"{% endfor %}{% endif %}>
+        <li class="ons-document-list__item{{ ' ons-document-list__item--featured' if document.featured }}{{ ' ons-document-list__item--full-width' if document.featured and document.fullWidth == true }}{{ ' ' + document.classes if document.classes else '' }}"{% if document.attributes %}{% for attribute, value in (document.attributes.items() if document.attributes is mapping and document.attributes.items else document.attributes) %} {{attribute}}="{{value}}"{% endfor %}{% endif %}>
```
alec-hamilton-ons, line 5: @willcAND is the `is mapping` Python syntax?

alec-hamilton-ons:
Might be worth updating the macro options file to reflect the changes you're introducing

src/components/document-list/_macro.njk
willcAND: Yeah, this is the line of code I raised a ticket for - need to leave it as is though for now, will cause breaking changes.

https://github.com/ONSdigital/design-system/issues/2671

willcAND (approved):

alec-hamilton-ons (approved):

Issue #2645, opened 2023-03-31
Missing icons size example


jrbarnes9:
The following example was mistakenly removed in #2454 

https://github.com/ONSdigital/design-system/blob/8a9e53db6e8f5acde5442ec58d21d133c3fb0bb8/src/foundations/style/icons/examples/icons-size/index.njk

PR #2729, opened 2023-07-17
Bugfix/2645/missing icon size example

willcAND:
#2645 

### How to review
Spin up locally, confirm example is loading as expected.


src/foundations/icons/example-icons-size.njk
```diff
@@ -0,0 +1,64 @@
+{% from "components/panel/_macro.njk" import onsPanel %}
+
+{{
+    onsPanel({
+        "type": 'success',
+        "id": 'success-id',
+        "iconType": 'check',
+        "iconSize": 'xxxl',
+        "body": '<h2 class="ons-u-fs-xxxl">ons-u-fs-xxxl</h2>'
```
alec-hamilton-ons, line 9: I think these should just say "xxxl" instead of prefixing the whole class because ultimately that is what you give to the `iconSize` param

backstop_data/bitmaps_reference/ds-vr-test__foundations_icons_example-icons-size_0_document_2_mobile.png
alec-hamilton-ons, line 1: Would it make sense to show some of the other icons here?

alec-hamilton-ons (changes requested):

src/foundations/icons/example-icons-size.njk
willcAND: Yeah I agree with this, will change.

alec-hamilton-ons (dismissed):

precious-onyenaucheya-ons (approved):

alec-hamilton-ons (approved):

Issue #1826, opened 2022-01-19
Inconsistent link param naming footer/header


rmccar:
Inconsistent param naming for the text for header nav links and footer links. I think this should be changed so they both use `text` for the link text param

Footer uses `text`:
```
"itemsList": [
    {
        "text": 'Cookies and privacy',
        "url": '#0'
    }
]
```

Header uses `title`:
```
"itemsList": [
    {
        "title": "Help",
        "url": "#0"
    }
]
```

PR #2741, opened 2023-07-20
Enhancement/1826/header footer param naming

willcAND:
#1826 

### How to review
Load up header examples, confirm all look correct.

### DIFF
--- a/src/components/header/_macro.njk
+++ b/src/components/header/_macro.njk
@@ -102,9 +102,9 @@
                                                         href="{{ item.url }}"
                                                         class="ons-header-service-nav__link"
                                                         {% if item.id %} id="{{ item.id }}"{% endif %}
-                                                    >{{ item.title }}</a>
+                                                    >{% if item.title %}{{ item.title }}{% elif item.text %}{{ item.text }}{% endif %}</a>
                                                 {% else %}
-                                                    {{ item.title }}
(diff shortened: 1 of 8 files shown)


src/components/header/_macro.njk
```diff
@@ -102,9 +102,9 @@
                                                         href="{{ item.url }}"
                                                         class="ons-header-service-nav__link"
                                                         {% if item.id %} id="{{ item.id }}"{% endif %}
-                                                    >{{ item.title }}</a>
+                                                    >{% if item.title %}{{ item.title }}{% elif item.text %}{{ item.text }}{% endif %}</a>
```
alec-hamilton-ons, line 105: Is this to make the changes not breaking?

willcAND: Correct :)

alec-hamilton-ons: Is this a good idea? Leaving legacy code in? Personally I feel like it would be better to just make the breaking change. Your call though.

willcAND: That was my initial thought yeah, however since it's in the header, I think it would affect a lot of users. Had a chat with Rich about it who raised the ticket initially and he thinks it's better to avoid a BC if we can for a time. Raised another ticket as a reminder to take it out later too.

alec-hamilton-ons (approved):

precious-onyenaucheya-ons (approved):

Issue #2176, opened 2022-05-06
Add more options to the 'Reply' component's textarea


jrbarnes9:
The reply component used in 'send and receive' messages pattern would benefit from more of the textarea component's options such as:
- character limit counter (`CharCheckLimit`)
- label hint description (`label.description`)
- number of rows (`rows`)

PR #2761, opened 2023-07-26
Enhancement/2176/options for reply component

willcAND:
#2176 

### How to review
Modify example, confirm new parameters are available inside `onsReply`. Tests pass.


rmccar (commented):
Looks good but maybe we should add the limit and description to one of the examples too

willcAND:
> Looks good but maybe we should add the limit and description to one of the examples too

I did try this originally, couldn't think of a good descriptive label for a reply message.. Will have another go and add back in.

alessioventuriniAND (approved):

rmccar (approved):

Issue #2732, opened 2023-07-18
Header menus on smaller viewports have an accessible name that does not match their visible label


emilymaylawes:
The mobile menu buttons on the header, 

![Image](https://github.com/ONSdigital/design-system/assets/131370908/c8b9d272-ec46-49b0-867f-ab409774b5dd)
![Image](https://github.com/ONSdigital/design-system/assets/131370908/31935f51-833f-4f8a-925d-cbabdf4df567)

Have accessible names, as denoted by their `aria-label`, respectively, "Toggle main navigation" and "Toggle section navigation". These accessible names do not contain the visible label, "Menu" and "Design system" which is a failure of [2.5.3 Label in name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html#:~:text=Understanding%20SC%202.5.,Label%20in%20Name%20(Level%20A)&text=For%20user%20interface%20components%20with,text%20that%20is%20presented%20visually.&text=A%20best%20practice%20is%20to,the%20start%20of%20the%20name.).
This is especially a problem for users who activate controls with speech, who may ask to "Click menu" or "Click design system" and nothing happens. It also can be confusing for screen reader users who have vision and can see the page, and see the disparity in the visual text and what is read out loud.

These must be updated to contain the visible label. 

A suggestion is "Toggle menu" and "Toggle Design System menu". Programmatically update the latter's accessible name dependent on the current section.

This is currently a failure on the Accessibility Statement.

This also exists on the example [Example header with sub navigation](https://service-manual.ons.gov.uk/design-system/components/header#example.example-header-external-with-sub-navigation.html) where the 'Menu' button is correctly named, but the section navigation does not contain the visible label "Design System".

It is suggested the Nunjucks options `ariaLabel` for the ToggleButton, ServiceLinks and Item are provided guidance to help users know that it must include the visible label, as`aria-label` replaces any visible text.

PR #2762, opened 2023-07-26
Bugfix/2732/header menu accessible name

willcAND:
#2732 

### How to review
Confirm header with sub nav example has correct aria labels.


rmccar (approved):

src/components/navigation/_macro.njk
```diff
@@ -68,7 +68,7 @@
                 "type": "button",
                 "variants": ["mobile", "dropdown"],
                 "attributes": {
-                    "aria-label": "Toggle section navigation",
+                    "aria-label": "Toggle " + params.navigation.currentPageTitle + " menu" if params.navigation.currentPageTitle else "Toggle section menu",
```
alec-hamilton-ons, line 71: Seems like the syntax here is wrong. Doesn't the `if` need to go first? If `params.navigation.currentPageTitle` doesn't exist then it will fall over I think.

src/components/navigation/_macro.spec.js
```diff
@@ -215,7 +215,7 @@ describe('macro: navigation', () => {
         variants: ['mobile', 'dropdown'],
         type: 'button',
         attributes: {
-          'aria-label': 'Toggle section navigation',
+          'aria-label': 'Toggle Main nav item 2 menu',
```
alec-hamilton-ons, line 218: Main doesn't need to be capitalised

alec-hamilton-ons (changes requested):

src/components/navigation/_macro.njk
willcAND: It does look a bit weird I agree, seems to be similar syntax used in other places when passing props. Also tested it and it works. SS below is from `example-header-external-with-sub-navigation.njk` with currentPageTitle removed.

Screenshot 2023-07-27 at 11 16 07

src/components/navigation/_macro.spec.js
willcAND: updated

alec-hamilton-ons (approved):

PR #2773, opened 2023-08-04
modfied generate npm-package file

precious-onyenaucheya-ons:
fixes the issue with arises when installing the design system as a dependency

### How to review
Follow the steps in the service manual to install the design system as an rpm package and check if it throws an error that it can't find the fonts folder


alessioventuriniAND (approved):

rmccar (commented):
I think the copy should either be removed from the `copyBaseSass` function or the function should be renamed. Also wouldn't it be better to move the fonts instead to keep the package as small as possible? I don't think that should break anything because currently they aren't being found in the static folder. Unless you guys can think of a reason to keep them in both places?

precious-onyenaucheya-ons:
> I think the copy should either be removed from the `copyBaseSass` function or the function should be renamed. Also wouldn't it be better to move the fonts instead to keep the package as small as possible? I don't think that should break anything because currently they aren't being found in the static folder. Unless you guys can think of a reason to keep them in both places?

I have copied the fonts to the sass folder to solve the issue we were having, Though, I don't see why we need to rename the function. 

I didn't want to break anything, that was the reason I left it in both places

rmccar:
> > I think the copy should either be removed from the `copyBaseSass` function or the function should be renamed. Also wouldn't it be better to move the fonts instead to keep the package as small as possible? I don't think that should break anything because currently they aren't being found in the static folder. Unless you guys can think of a reason to keep them in both places?
> 
> I have copied the fonts to the sass folder to solve the issue we were having, Though, I don't see why we need to rename the function.
> 
> I didn't want to break anything, that was the reason I left it in both places

I just think it should be renamed because it now doesn't just copy the scss, it copies the fonts too.

Yeah I understand you don't want to break anything but I don't think its a good idea to have multiple copies of the fonts in the npm package. As long as you test that you can import the package into a project with the fonts just in the fonts folder and that the project then builds and works correctly whereas before the copy it failed then I think we should be ok

rmccar:
Running `yarn npm-bundle` I've realised that we probably need to add `/js/` and `/layout/` to the .gitignore file under `npm package folders`

rmccar (approved):

PR #2774, opened 2023-08-07
autosuggest width aligned with input field

alessioventuriniAND:
The autosuggest suggestions were not aligned with the input field. This introduces a fix where the width of the input is equal to the width of the autosuggest results. 

### How to review
Set a different width to the two examples we have in the autosuggest component to confirm that the autosuggest results are aligned with the input width.

### DIFF
--- a/src/components/autosuggest/_macro.njk
+++ b/src/components/autosuggest/_macro.njk
@@ -36,7 +36,7 @@
     >
 
         {% set autosuggestResults %}
-            <div class="ons-autosuggest__results ons-js-autosuggest-results {% if params.width %} ons-input{% if params.type == 'number' or params.type == 'tel' %}-number{% endif %}--w-{{ params.width }}{% endif %} {% if not params.width %} ons-input--w-20 {% endif %}">
+            <div class="ons-autosuggest__results ons-js-autosuggest-results {% if params.width %} ons-input--w-{{ params.width }}{% else %} ons-input--w-20 {% endif %}">
                 <div id="{{ params.resultsTitleId }}" class="ons-autosuggest__results-title ons-u-fs-s">{{ params.resultsTitle }}</div>
                 <ul class="ons-autosuggest__listbox ons-js-autosuggest-listbox" title="{{ params.resultsTitle }}" role="listbox" id="{{ params.id }}-listbox" tabindex="-1"></ul>
             </div>
--- a/src/components/autosuggest/_macro.njk
+++ b/src/components/autosuggest/_macro.njk
@@ -36,7 +36,7 @@
     >
 
         {% set autosuggestResults %}
-            <div class="ons-autosuggest__results ons-js-autosuggest-results {% if params.width %}ons-input--w-{{ params.width }}{% else %}ons-input--w-20{% endif %}">
+            <div class="ons-autosuggest__results ons-js-autosuggest-results{% if params.width %} ons-input--w-{{ params.width }}{% else %} ons-input--w-20{% endif %}">
                 <div id="{{ params.resultsTitleId }}" class="ons-autosuggest__results-title ons-u-fs-s">{{ params.resultsTitle }}</div>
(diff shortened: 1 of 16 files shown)


src/components/autosuggest/_macro.njk
```diff
@@ -36,14 +36,13 @@
...
         {% set autosuggestResults %}
-            <div class="ons-autosuggest__results ons-js-autosuggest-results">
+            <div class="ons-autosuggest__results ons-js-autosuggest-results {% if params.width %} ons-input{% if params.type == 'number' or params.type == 'tel' %}-number{% endif %}--w-{{ params.width }}{% endif %} {% if not params.width %} ons-input--w-20 {% endif %}">
                 <div id="{{ params.resultsTitleId }}" class="ons-autosuggest__results-title ons-u-fs-s">{{ params.resultsTitle }}</div>
                 <ul class="ons-autosuggest__listbox ons-js-autosuggest-listbox" title="{{ params.resultsTitle }}" role="listbox" id="{{ params.id }}-listbox" tabindex="-1"></ul>
             </div>
             <div class="ons-autosuggest__instructions ons-u-vh ons-js-autosuggest-instructions" id="{{ params.id }}-instructions" tabindex="-1">{{ params.instructions }}</div>
             <div class="ons-autosuggest__status ons-u-vh ons-js-autosuggest-aria-status" aria-live="assertive" aria-atomic="true" role="status" tabindex="-1"></div>
         {% endset %}
-        <div class="ons-autosuggest__wrapper">
```
rmccar, line 46: If you are going to remove this will need to remove the css to go along with it. There is a `&__wrapper` class in the css.

src/components/autosuggest/_macro.njk
```diff
@@ -36,14 +36,13 @@
     >
 
         {% set autosuggestResults %}
-            <div class="ons-autosuggest__results ons-js-autosuggest-results">
+            <div class="ons-autosuggest__results ons-js-autosuggest-results {% if params.width %} ons-input{% if params.type == 'number' or params.type == 'tel' %}-number{% endif %}--w-{{ params.width }}{% endif %} {% if not params.width %} ons-input--w-20 {% endif %}">
```
rmccar, line 39:
```suggestion
            <div class="ons-autosuggest__results ons-js-autosuggest-results{% if params.width %} ons-input{% if params.type == 'number' or params.type == 'tel' %}-number{% endif %}--w-{{ params.width }}{% else %} ons-input--w-20{% endif %}">
```
I would probably do this, makes it a bit simpler and removes some extra spaces from the html.

But I'm also not sure we need the `params.type` stuff. Should always be text which I think is why `params.type` isn't currently passed in.

precious-onyenaucheya-ons (approved):

rmccar (changes requested):

src/components/autosuggest/example-autosuggest-country.njk
```diff
@@ -5,6 +5,7 @@
 
         {{ onsAutosuggest({
             "id": "country-of-birth",
+            "width": 10,
```
rmccar, line 8: I think this looks a bit too narrow in the example. To demonstrate this I would set it to 30 or something

rmccar (commented):
I think theres something weird happening with the husky files. I don't think they should be included in this PR, you'll probably need to merge main to resolve it

src/components/autosuggest/_macro.njk
```diff
@@ -36,14 +36,13 @@
     >
 
         {% set autosuggestResults %}
-            <div class="ons-autosuggest__results ons-js-autosuggest-results">
+            <div class="ons-autosuggest__results ons-js-autosuggest-results {% if params.width %} ons-input--w-{{ params.width }}{% else %} ons-input--w-20 {% endif %}">
```
rmccar, line 39:
```suggestion
            <div class="ons-autosuggest__results ons-js-autosuggest-results{% if params.width %} ons-input--w-{{ params.width }}{% else %} ons-input--w-20{% endif %}">
```
These spaces will add unwanted space to the class list unless removed

PR #2777, opened 2023-08-09
Fix/2775/grid examples moved into the ds repo

alessioventuriniAND:
The grid example were living in the service manual. They have been moved in the DS repo as we need to display the html tab in the example for ease of access to the code. 
A helper folder has been created within the components folder to store the grid helper component. This is not a component that we recommend people using. It has been created to keep the example code clean. 

### How to review
Check that the example are visible in the foundation path under grid.

### DIFF
--- a/src/foundations/grid/example-grid-12.njk
+++ b/src/foundations/grid/example-grid-12.njk
@@ -1,7 +1,7 @@
 ---
-layout: example
+layout: layouts/example
 ---
-{% from "helpers/design-system/grid.njk" import patternLibExampleGrid %}
+{% from "components/helpers/grid.njk" import patternLibExampleGrid %}
 
 <style>
     .ons-pl-grid-col {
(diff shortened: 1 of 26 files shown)


precious-onyenaucheya-ons (approved):

src/foundations/grid/example-grid-12.njk
```diff
@@ -0,0 +1,24 @@
...
+{% from "components/helpers/grid.njk" import patternLibExampleGrid %}
+
+<style>
+    .ons-pl-grid-col {
+        background: var(--ons-color-grey-5);
+        font-size: 0.8rem;
+        margin: 0 0 1rem;
+        padding: 1rem;
+    }
+</style>
```
rmccar, line 13: Can we not move this out into a scss file?

rmccar (approved):

PR #2779, opened 2023-08-10
add accessible tag for phase banner

### LINKED ISSUE #2462: Phase banner: Review most semantically appropriate and accessible tag for badge
Looking at the HTML for this component, the BETA itself is wrapped in a H3 tag. Given its position on the page, this means (from what I could make out on Voiceover) that it appears as the first heading. Would this potentially be confusing for users flicking through the headers to just hear 'Beta' and nothing else? 
Within the GOV.UK Design System, they seem to use a strong tag instead. Consequently it's not read out when navigating through the various headers on the page.

_Originally posted by @LaurenDowling in https://github.com/ONSdigital/design-system/discussions/8#discussioncomment-3753862_

precious-onyenaucheya-ons:
Fixes #2462 
Changed h3  tag to  strong  tag


alessioventuriniAND (approved):

rmccar (approved):
In the changes between the examples in the screenshots taken with backstop the line under the

PR #2789, opened 2023-08-22
Refactor summary hidden span text change

rmccar:
This PR changes and reverts some of the changes made in #2679.
This is because looking at the issue it addressed it didn't actually make the changes needed. The issue didn't call for getting rid of the `ariaLabel` param so I have reintroduced that and renamed it. Also there was some hardcoding of content that we want to avoid to make it as customisable as possible for the user and give them the option to translate the text into other languages.

Ive also added a `aria-hidden` span around the link text in order to make the text set in the `visuallyHiddenText` param to make more sense and allow the user to set exactly what will be read out by screen readers but also not repeat the link text.

***THIS IS STILL A BREAKING CHANGE***
Services upgrading that use `ariaLabel` will now need to rename that param to `visuallyHiddenText`.

### How to review
Test with voiceover and that the correct text is read out in the correct context


rmccar:
[Update DS to v65](https://trello.com/c/XLu69hsd/5840-update-ds-to-v65)

alessioventuriniAND (approved):

precious-onyenaucheya-ons (approved):

PR #2791, opened 2023-08-29
Force details open on print

rmccar:
When pages with details components are printed in eQ. The content for the details components is still hidden unless the details components are opened manually first. This solution will automatically open the details components when the print preview dialog is opened and then close it when the dialog is closed.

### How to review
Test one of the details or accordion examples and try printing the page. The details components should open when the dialog is open and close again when it is closed.


alessioventuriniAND (changes requested):
Would it be possible for you to discuss this change with Joe and Nikki and get their feedback/approval for this change as we are forcing this behaviour for all users. If they approve I am happy to approve the code too

rmccar:
> Would it be possible for you to discuss this change with Joe and Nikki and get their feedback/approval for this change as we are forcing this behaviour for all users. If they approve I am happy to approve the code too

Theres a few issues around this implementation. I need to do some more investigation but will have that conversation when I get further 👍

rmccar:
Going with a different plan. Will raise another PR when this work is prioritised.

PR #2794, opened 2023-09-04
update tabs title

### LINKED ISSUE #2710: [Bug] titleTag levels above tabs needs to be audited for heading level, as should not all be <h4>
In relation to https://github.com/ONSdigital/ons-service-manual-website/issues/281, I think a little bit more work needs to be done to audit through all pages of the design system and set the `titleTag` of the tab components displaying the individual component to get the right heading level. On some pages, there are multiple different requirements for heading levels. Some need to be `<h4>` if they are in variants, some need to be `<h3>` if they are under a subheading, some need to be `<h2>` if they are directly under the main page heading.

Probably a bit of spike/investigation/audit work to go through all pages to make sure the `titleTag` (hidden heading above the tabs, which then displays on mobile) the right heading level.

We want to avoid ever skipping heading levels to make the page structure correct and screen reader users able to navigate through in a more accurate manner.

precious-onyenaucheya-ons:
fixes #2710 
I have added an utility class to the tab title to set the font properties

### How to review
check if the tab title has class "ons-u-fs-m"  added to it

### DIFF
--- a/src/components/tabs/_macro.njk
+++ b/src/components/tabs/_macro.njk
@@ -11,7 +11,7 @@
         {% if params.noInitialActiveTab %}data-no-initial-active-tab="true"{% endif %}
     >
         {% set titleTag = params.titleTag | default("h2") %}
-        <{{ titleTag }} class="ons-tabs__title ons-u-fs-r--b ons-u-mt-no">{{ params.title }}</{{ titleTag }}>
+        <{{ titleTag }} class="ons-tabs__title ons-u-fs-r--b ons-u-mt-no ons-u-fs-m">{{ params.title }}</{{ titleTag }}>
         <div>
             <ul class="ons-tabs__list">
                 {%- for tab in params.tabs -%}


alessioventuriniAND (approved):

rmccar:
This PR doesn't address the issue. Allowing the user to add a class to change the size of the text on the screen doesn't fix the issue. This is about heading levels being skipped in the html of some pages, which means that sometimes for example there is a `<h4>` directly below a `<h2>` in the hierarchy. To fix this in this example the `<h4>` should be a `<h3>`. This is a much bigger card and will need some research and investigation on how best to solve the problem.

src/components/tabs/_macro.njk
```diff
@@ -11,7 +11,7 @@
         {% if params.noInitialActiveTab %}data-no-initial-active-tab="true"{% endif %}
     >
         {% set titleTag = params.titleTag | default("h2") %}
-        <{{ titleTag }} class="ons-tabs__title ons-u-fs-r--b ons-u-mt-no">{{ params.title }}</{{ titleTag }}>
+        <{{ titleTag }} class="ons-tabs__title ons-u-fs-r--b ons-u-mt-no {% if params.titleTagClass %}{{params.titleTagClass}}{% endif %}">{{ params.title }}</{{ titleTag }}>
```
rmccar, line 14:
```suggestion
        <{{ titleTag }} class="ons-tabs__title ons-u-fs-r--b ons-u-mt-no{% if params.titleTagClass %} {{ params.titleTagClass }}{% endif %}">{{ params.title }}</{{ titleTag }}>
```
This change fixes some styling and will not introduce spaces into the class list if there is no `titleTagClass`

src/components/tabs/_macro-options.md
```diff
@@ -2,6 +2,7 @@
 | ------------------ | --------------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
 | title              | string          | true     | The `h2` level heading for the tabs `section` element required to give context for screen readers. Visually hidden when tabs are visible. |
 | titleTag           | string          | false    | The HTML heading tag to wrap the title text in for it’s correct semantic order on the page. Will default to an `h2`                       |
+| titleTagClass      | string          | false    | Additional class to be added to titleTag                                                                                                  |
```
rmccar, line 5: Think we should rename `titleTagClass` to something like `titleClasses` or something because it isn't the tag the classes are being set on so I feel its more descriptive and fits more with what has been used elsewhere in the DS

src/components/tabs/_macro.njk
```diff
@@ -11,7 +11,7 @@
         {% if params.noInitialActiveTab %}data-no-initial-active-tab="true"{% endif %}
     >
         {% set titleTag = params.titleTag | default("h2") %}
-        <{{ titleTag }} class="ons-tabs__title ons-u-fs-r--b ons-u-mt-no">{{ params.title }}</{{ titleTag }}>
+        <{{ titleTag }} class="ons-tabs__title ons-u-fs-r--b ons-u-mt-no{% if params.titleTagClass %} {{ params.titleTagClass }}{% endif %}">{{ params.title }}</{{ titleTag }}>
```
rmccar, line 14:
```suggestion
        <{{ titleTag }} class="ons-tabs__title ons-u-fs-r--b ons-u-mt-no{% if params.titleClasses %} {{ params.titleClasses }}{% endif %}">{{ params.title }}</{{ titleTag }}>
```
After this change it looks all good

alessioventuriniAND (approved):

rmccar (approved):

PR #2801, opened 2023-09-11
Fix/2736 error container in pattern marked as hidden but has a focusable element

### LINKED ISSUE #2736: Error container underneath the example "What is your address" pattern has been marked as hidden for screen readers but has a focusable element inside it
The example [Non-editable address input](https://service-manual.ons.gov.uk/design-system/patterns/addresses/example-address-input-non-editable) in the pattern 'Ask users for addresses' has an error container that has been made hidden in the accessibility tree by `aria-hidden`. **Emily's comment:** Is this supposed to be there? Or is this a bug we need to separately raise, either way, the following still stands.

Image

Hidden elements should not have focusable children.

There is a second visually hidden `<div>` underneath this, which just has the rich text without the link, and a `role="status"` which likely is intended to be read as a status message to screen readers.

Image

The same content will be read out on sequential reading order, however the link is not parsed and therefore when reading out with control + option + arrow keys, the link is not included. The link will be accessed with the `tab` key, however it is worth noting that in a usability study, [not one participant used the tab key as their primary method of navigating](https://jessbudd.com/blog/screen-reader-usability-testing-observations/). The rotor functionality on VoiceOver, activated with control + option + U, or similar, can show users a list of links. After testing, this link was not included in this list and therefore likely will be missed entirely.

It is recommended to not use `aria-hidden` on this container, and place the live region roles and relevant attributes on the visible container instead, if possible.

It is worth investigating why this pattern was created in this way in the first place to understand user need.

precious-onyenaucheya-ons:
fixes #2736 
Removed aria-hidden from list warning element because it has a focusable child element. The aria-hidden is set  (within div element with role=status) only when the list warning  element is created.

The autosuggest results(ie title and list of results) has also been removed from the html content whenever there is an error.

Now the screen reader reads the message just once.  The screen reader also reads out the instructions on how to interact with it.  

### How to review
For all components and patterns that use the auto suggest, check if the aria-hidden tag has been removed from the list element with the focusable child element


rmccar (commented):
From the explanation in the issue its not completely clear exactly what needs to be done, maybe we need to confirm with Emily if possible but *I think* that its also suggesting we move the live region roles too etc too

alessioventuriniAND (approved):

rmccar:
Ive tested this with voiceover and comparing it to what is currently in the service manual after this change it now reads out what is in the error message twice, which I don't think is something we want.
I have also noticed that in the service manual it reads out the ways to interact with it twice which is something we might want to look into too.

rmccar:
In all the new screenshots that have just been added it highlights there is an extra thick line at the bottom of the error box. I think this may be because you have changed the list element to be a div. Also because of this change you now have a redundant unordered list with no list items in the html.
The more I look into this the more I think this solution isn't a change we want to make. Visually and aria hiding the status means it is no longer used at all so is there any point in having it at all? But also the message it reads out from the status makes much more sense when using a screenreader than the one we have now because it is meant to be a replacement for the one with the link.
Personally I think this needs more thought and discussion on what the solution for this should be and if that solution is better than what we already have. This is a bigger issue to fix than what it first seems.

src/components/autosuggest/autosuggest.ui.js
```diff
@@ -497,13 +504,12 @@ export default class AutosuggestUI {
   }
 
   createWarningElement(content) {
-    const warningListElement = document.createElement('li');
+    const warningDivElement = document.createElement('div');
```
rmccar, line 507: I think this should be renamed because its not very descriptive maybe `warningContainer` or something

rmccar (commented):
Has this been tested with the API?

src/components/autosuggest/autosuggest.ui.js
```diff
@@ -43,6 +43,7 @@ export default class AutosuggestUI {
     this.input = context.querySelector(`.${baseClass}-input`);
     this.resultsContainer = context.querySelector(`.${baseClass}-results`);
     this.listbox = this.resultsContainer.querySelector(`.${baseClass}-listbox`);
+    this.resultsTitleContainer = this.resultsContainer.querySelector(`.ons-autosuggest__results-title`);
```
rmccar, line 46: This should also use `${baseClass}` for consistency

```suggestion
    this.resultsTitleContainer = this.resultsContainer.querySelector(`.${baseClass}__results-title`);
```

rmccar (commented):
Few more minor comments

rmccar (approved):

alessioventuriniAND (approved):

PR #2802, opened 2023-09-14
Update chalk to v5

rmccar:
Updates chalk to v5

### How to review
Tests pass


rmccar:
Conversion of DS to esm. Needs more research. For now we will stay on v4

PR #2804, opened 2023-09-14
Update pr template, add pr checks and label checks

rmccar:

This PR adds the label and task checker github actions and updates the PR template. This will ensure we set all the correct data on a PR before it can be merged. Also updates the release template to include some new labels.

### How to review this PR

- Tasks pass
- Removing the label or unchecking a task causes the tasks to fail

### Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,18 @@
+<!-- ignore-task-list-start -->
+
 ### What is the context of this PR?
+
 Describe what you have changed and why, link to other PRs or Issues as appropriate.
 
-### How to review
+### How to review this PR
+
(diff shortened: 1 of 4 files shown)


alessioventuriniAND (approved):

precious-onyenaucheya-ons:
We should also check that a ticket has been linked to the PR

rmccar:
> We should also check that a ticket has been linked to the PR

I think whether that is required should be decided in a discussion on ways of working etc. For example this PR doesn't currently have a ticket as it was a bit ad hoc. Is that something we want to allow for? Or are we saying every PR needs to have a ticket? I think obviously the ticket should be linked if there is one but my opinion is I don't think that needs to be absolutely mandatory whereas the other things in this PR should be because they make a difference to our release documentation and keeping the board up to date with the work that is going on. Happy to discuss

PR #2817, opened 2023-09-28
2753 error panel needs extending to accept two errors for one input

### LINKED ISSUE #2753: Error panel needs extending to accept two errors for one input
On eQ we need the error panel component to be extended to accept more than one error if there are two things wrong with the answer at the moment the second error is truncated and not shown to the user. [Here](https://eq-runner-prototypes.netlify.app/prototypes/error/) is a link to a prototype with an example of this functionality.

precious-onyenaucheya-ons:

This fixes #2753.  This ticket ensures that the error panel accepts multiple errors for a single input. Users can now enter a list of errors for a single input. 

### How to review this PR

Check that the error panel component can accept multiple error message as well as a single one. 
### Checklist


src/components/error/_macro-options.md
```diff
@@ -0,0 +1,5 @@
+| Name       | Type                      | Required | Description                                                        |
+| ---------- | ------------------------- | -------- | ------------------------------------------------------------------ |
+| id         | string                    | true     | The HTML id attribute that the error summary link can anchor to    |
+| text       | string or array`<string>` | true     | The text describing the error                                      |
+| attributes | object                    | false    | HTML attributes (for example, data attributes) to add to the field |
```
rmccar, line 5: We dont want this file added as error isnt really a component the user can use on its own. Because of this, this has been documented in the content on the service manual

rmccar, line 5: We dont want this file added as error isnt really a component the user can use on its own. Because of this, this has been documented in the content on the service manual

src/components/error/_macro.njk
```diff
@@ -2,19 +2,30 @@
...
+                <strong>{{ item }}</strong>
+            </li>
+            {% endfor %}
+        {% else %}
+        <li class="ons-panel__error ons-list__item"
             {% if params.attributes %}{% for attribute, value in (params.attributes.items() if params.attributes is mapping and params.attributes.items else params.attributes) %}{{ attribute }}{% if value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}
         >
             <strong>{{ params.text }}</strong>
-        </p>
+        </li>
```
rmccar, line 19: This should just stay as a `<p>` and be moved out of the `<ul>`. There isnt any real need for this to be a list if there is only one item and it could cause breaking changes in tests

src/components/error/_macro.njk
```diff
@@ -2,19 +2,30 @@
```
rmccar, line 2: The formatting is a bit off in this file

src/components/error/_macro.spec.js
```diff
@@ -36,11 +36,23 @@ describe('macro: error', () => {
...
 
-    expect(
-      $('.ons-panel__error')
-        .text()
-        .trim(),
-    ).toBe('Example error text.');
+    expect($('.ons-panel__error').text().trim()).toBe('Example error text.');
+  });
+
+  it('has the provided `text` if an array was provided', () => {
```
rmccar, line 42: I think this test should have a more descriptive description. Something like "when the 'text' param is set to an array type, then all error messages in the array are shown".
I think we may need a card to look at the way the tests are structured and use Given-When-Then to write the descriptions or something similar

src/patterns/correct-errors/example-error-proto-multiple-error-per-input.njk
```diff
@@ -0,0 +1,96 @@
...
+        },
+        "year": {
+            "label": {
+                "text": "Year"
+            },
+            "name": "year",
+            "value": "2018"
+        }
+    })
+}}
```
rmccar, line 58: To show the functionality I think we only need one date input and just have it ask for date of birth or something would make it simpler as the extra date input isn't really adding anything. Or could just use a single regular input with 2 problems with the answer might be even better as this isn't only going to be used on components with more than one input

src/patterns/correct-errors/example-error-proto-multiple-error-per-input.njk
```diff
@@ -0,0 +1,96 @@
...
+                    {
+                        "text": "Enter a number between 1 to 12",
+                        "url": "#month-error",
+                        "variants": "inPageLink"
+                    },
+                    {
+                        "text": "Enter a date that is after 1 January 2019",
+                        "url": "#year-error",
+                        "variants": "inPageLink"
+                    }
```
rmccar, line 26: These links currently error because `"#month-error"` and `"#year-error"` don't exist on the page

rmccar (changes requested):

rmccar:
Might be something you want to check with Dina but I think the spacing has changed when making this a list compared to a panel with a single error. This extra margin-top that has been added because it is now a list might need removing and the margin bottom added at the bottom of the list might need increasing to match `<p>` for consistency

Screenshot 2023-09-28 at 11 17 25

src/components/error/_macro-options.md
precious-onyenaucheya-ons: I wasn't aware of this

Will remove

src/patterns/correct-errors/example-error-proto-multiple-error-per-input.njk
precious-onyenaucheya-ons: missed this

Will update

src/patterns/correct-errors/example-error-proto-multiple-error-per-input.njk
rmccar: The example now no longer makes sense because you have a period from but no period to.
I think an example based on this would be good as its a real eQ example of what would happen in a real questionnaire.

Screenshot 2023-09-28 at 13 47 00

Issue #1668, opened 2021-09-08
Skip link hover state underline


rmccar:
![image](https://user-images.githubusercontent.com/42928680/132520975-23b5af4e-5ae7-4764-9253-30a8272e3438.png)
Skip link underline is no longer getting thicker when hovered like other links like breadcrumbs

PR #2820, opened 2023-09-28
Increase underline to match other focussed elements

adi-unni:

Fixes: #1668

Skip link previously had a thin underline when focussed. As the components primary use case is for keyboard users, the route of navigation is through the tab index. When hitting tab, the element gets set from a non visible state, straight to focussed which is the last before an action. As with the rest of the components, the underline should be the thickest when focussed. The change increases the underline thickness to match with the rest of the components' focussed states. 

### How to review this PR

Press tab to bring up the component and check if the new skip-to-content component has a thicker underline.


src/components/skip-to-content/_skip.scss
```diff
@@ -14,7 +14,7 @@
 
   &:focus {
     background-color: var(--ons-color-focus);
-    box-shadow: 0;
+    box-shadow: 0;
```
rmccar, line 17:
```suggestion
    box-shadow: 0;
```
Should get rid of this extra spacing thats been added. Other than that looks good!

rmccar (approved):

Issue #2818, opened 2023-09-28
Focus states are too wide on card components


rmccar:
![Screenshot 2023-09-28 at 10 03 59](https://github.com/ONSdigital/design-system/assets/42928680/9d651cc5-4918-422e-bd27-f4c9c48341c5)

PR #2827, opened 2023-10-05
Change widths of components inside card to be their contents width

adi-unni:

Fixes: #2818

Issue was within the card component. Previously, card's components would scale to the width of the whole card, but now scale to just their contents. Additionally, incorrect margin's were being applied to headings within the card, so removing that bring design back in line with other components when tabbing. 

### How to review this PR

Pressing tab on the card components should reveal new boundaries for focussed states within the card.

Before and After for components changed when focussed:

Screenshot 2023-10-05 at 10 52 07
Screenshot 2023-10-05 at 10 53 29

Screenshot 2023-10-05 at 10 52 22
Screenshot 2023-10-05 at 10 53 42

### Checklist

### DIFF
--- a/src/components/card/_card.scss
+++ b/src/components/card/_card.scss
@@ -13,14 +13,6 @@
     }
   } 
 
-  &__image--reorder {
-    order: 1;
-  }
-
-  &__title--reorder {
-    order: 2;
-  }
-
   &__link:hover {
     text-decoration-thickness: 3px;
   }
--- a/src/components/card/_card.scss
+++ b/src/components/card/_card.scss
@@ -7,11 +7,11 @@
     flex-direction: column;
     margin-bottom: 1rem;
     width: fit-content;
-  }
 
-  &__remove--margin {
-    margin-bottom: 0;
-  }
+    & > .ons-card__title {
+      margin-bottom: 0;
+    }
+  } 
 
   &__image--reorder {
     order: 1;
--- a/src/components/card/_card.scss
+++ b/src/components/card/_card.scss
@@ -5,6 +5,12 @@
   &__link {
     display: flex;
     flex-direction: column;
+    margin-bottom: 1rem;
+    width: fit-content;
+  }
+
+  &__remove--margin {
+    margin-bottom: 0;
   }
 
   &__image--reorder {
--- a/src/components/card/_macro.njk
+++ b/src/components/card/_macro.njk
@@ -10,7 +10,7 @@
       {%- if params.url -%}
         <a href="{{ params.url }}" class="ons-card__link">
       {%- endif -%}
-          <h{{ titleSize }} class="ons-card__title ons-card__title--reorder {{ params.titleClasses | default('ons-u-fs-m')}}" id="{{ params.id }}">
+          <h{{ titleSize }} class="ons-card__title ons-card__title--reorder {{ params.titleClasses | default('ons-u-fs-m')}} ons-card__remove--margin" id="{{ params.id }}">
               {{ params.title }}
           </h{{ titleSize }}>
           {% if params.image.smallSrc %}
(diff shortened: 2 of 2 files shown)


src/components/card/_card.scss
```diff
@@ -5,6 +5,12 @@
   &__link {
     display: flex;
     flex-direction: column;
+    margin-bottom: 1rem;
+    width: fit-content;
+  }
+
+  &__remove--margin {
+    margin-bottom: 0;
   }
```
rmccar, line 14:
```
  &__link {
    display: flex;
    flex-direction: column;
    margin-bottom: 1rem;
    width: fit-content;

    & > .ons-card__title {
      margin-bottom: 0;
    }
  }
```
I would do something like this instead because inside the link will always be a title. Then we can remove the `ons-card__remove--margin` class from the html too.

src/components/card/_macro.njk
```diff
@@ -24,15 +24,15 @@
...
+        <h{{ titleSize }} class="ons-card__title ons-card__title--reorder {{ params.titleClasses | default('ons-u-fs-m')}} ons-card__remove--margin" id="{{ params.id }}">
           {{ params.title }}
-        {%- if params.url -%}
-          </a>
-        {%- endif -%}
-      </h{{ titleSize }}>
+        </h{{ titleSize }}>
+      {%- if params.url -%}
+        </a>
+      {%- endif -%}
```
rmccar, line 35: Now you've made this change to reorder the `<a>` and `<h#>` there is a lot more repeated code here so you could simplify this and move the `{%- if params.image -%}` and put it above the `{% if params.image.smallSrc %}` line. Then you can remove a lot of the code around it.

src/components/card/_macro.njk
```diff
@@ -10,7 +10,7 @@
       {%- if params.url -%}
```
rmccar, line 10: Have a word with Dina about whether there is any need to use the card component without a link. Because in the docs it says `url` is required. If it is required theres no need for these ifs at all and then in the css change we should just be able to remove the `margin-bottom` line. If there is a need for it then we should update the documentation to say its not required.

rmccar (changes requested):

src/components/card/_macro.njk
alessioventuriniAND: I think the IDS are currently using the card component in their homepage without the urls. I think it might be best to update the documentation and make them optional.

rmccar: In that case, if we are allowing this we should add an example where there is no url. Then we can cover that case with the visual tests

alessioventuriniAND: Oh wait a second! I think I confused the url with the itemlist param. No you always have a link as a header. Please ignore my comment and go with Richard first suggestion. Speak with Dina and then the rest lol

src/components/card/_card.scss
adi-unni: Change has been made, I had tried this previously, but my styles wasn't applying so had to go with other approach, however this code is clearly working :)

src/components/card/_macro.njk
```diff
@@ -5,36 +5,18 @@
...
       </h{{ titleSize }}>
-
-    {%- endif -%}
+      {%- if params.image -%}
+        {% if params.image.smallSrc %}
+          <img class="ons-card__image ons-u-mb-s ons-card__image--reorder" height="200" width="303"{% if params.image.largeSrc %} srcset="{{ params.image.smallSrc }} 1x, {{ params.image.largeSrc }} 2x"{% endif %} src="{{ params.image.smallSrc }}" alt="{{ params.image.alt }}" loading="lazy">
+        {% elif params.image == true or params.image.placeholderURL %}
+          <img class="ons-card__image ons-u-mb-s ons-card__image--reorder" height="100" width="303" srcset="{{ params.image.placeholderURL if params.image.placeholderURL }}/img/small/placeholder-card.png 1x, {{ params.image.placeholderURL if params.image.placeholderURL }}/img/large/placeholder-card.png 2x" src="{{ params.image.placeholderURL if params.image.placeholderURL }}/img/small/placeholder-card.png" alt="" loading="lazy">
+        {% endif %}
+      {%- endif -%}
```
rmccar, line 18: Here we should move the header element below the image ones and then remove all the "reorder" css classes as they aren't needed

adi-unni: Made the appropriate changes

rmccar (approved):

PR #2878, opened 2023-10-31
update analytics.js file to be compatible with GA4

precious-onyenaucheya-ons:

The design system currently uses Universal Analytics  to set up events for google analytics. UA is deprecated and replaced by GA4.  This ticket was created to update the events defined in analytics.js so it is compatible with GA4.
### How to review this PR
Check that events data is gotten from the GA dashboard 

### Checklist


rmccar (commented):
Some minor comments before I review the changes to the JS

src/components/tabs/_macro.njk
```diff
@@ -15,7 +15,7 @@
         <div>
             <ul class="ons-tabs__list">
                 {%- for tab in params.tabs -%}
-                    <li class="ons-tab__list-item"><a href="#{{ tab.id if tab.id else tab.title|trim|replace(' ','-')|lower}}" class="ons-tab" data-ga="click" data-ga-category="tabs" data-ga-action="Show: {{ tab.title }}" data-ga-label="Show: {{ tab.title }}">{{ tab.title }}{% if tab.hiddenSpan %}<span class='ons-u-vh'>{{tab.hiddenSpan}}</span>{% endif %}</a></li>
+                    <li class="ons-tab__list-item"><a href="#{{ tab.id if tab.id else tab.title|trim|replace(' ','-')|lower}}" class="ons-tab" data-ga-category="tabs" data-ga-action="Show: {{ tab.title }}" data-ga-label="Show: {{ tab.title }}"{% if tab.attributes %}{% for attribute, value in (tab.attributes.items() if tab.attributes is mapping and tab.attributes.items else tab.attributes) %} {{ attribute }}{% if value %}="{{ value }}"{% endif %} {% endfor %}{% endif %}>{{ tab.title }}{% if tab.hiddenSpan %}<span class='ons-u-vh'>{{tab.hiddenSpan}}</span>{% endif %}</a></li>
```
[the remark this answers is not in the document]
alessioventuriniAND: things are left the same but what we are going to tell the users is that in order for them to start tracking elements they need to defin the data-ga attributes. So I think it's better if we don't add it from the get go.

src/components/details/_macro.njk
```diff
@@ -11,14 +11,14 @@
...
-                        })
+            {% set titleTag = params.titleTag | default("h2") %}
+            <{{ titleTag }} class="ons-details__title">{{ params.title }}</{{ titleTag }}>
+            <span class="ons-details__icon">
+                {% from "components/icon/_macro.njk" import onsIcon %}
+                {{-
+                    onsIcon({
+                        "iconType": "chevron"
+                    })
                     -}}</span>
```
rmccar, line 22:
```suggestion
                -}}</span>
```

src/components/tabs/_macro.njk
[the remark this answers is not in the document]
rmccar: Ok in that case do we also want to remove the category and action ones also then?

src/js/analytics.js
```diff
@@ -1,22 +1,26 @@
 import domready from './domready';
 
-export let trackEvent = (event, data) => {
-  console.log(`[Analytics disabled] Event: ${event}`); // eslint-disable-line no-console
-  console.log(data); // eslint-disable-line no-console
+export let trackEvent = (data) => {
```
rmccar, line 3: `data` here isn't being used

src/js/analytics.js
```diff
@@ -1,22 +1,26 @@
...
+  console.log('analitycs script connected'); // eslint-disable-line no-console
```
rmccar, line 4: "analytics" typo

src/js/analytics.js
```diff
@@ -1,22 +1,26 @@
...
 };
+console.log('connected');
+setTimeout(() => {
+  if (typeof window.google_tag_manager !== 'undefined') {
+    console.log('GTM active');
```
rmccar, line 9:
```suggestion
    console.log('GA active');
```
I think this should be GA because we might not use GTM in some services

src/js/analytics.js
```diff
@@ -1,22 +1,26 @@
...
 
-export let trackEvent = (event, data) => {
-  console.log(`[Analytics disabled] Event: ${event}`); // eslint-disable-line no-console
-  console.log(data); // eslint-disable-line no-console
+export let trackEvent = (data) => {
+  console.log('analitycs script connected'); // eslint-disable-line no-console
 };
+console.log('connected');
+setTimeout(() => {
+  if (typeof window.google_tag_manager !== 'undefined') {
```
rmccar, line 8: Same with this, what if we aren't using GTM?

src/js/analytics.js
rmccar, line 3: Think this change should be reverted because if GA is on this `trackEvent` is overwritten so it should be disabled at this stage

src/js/analytics.js
```diff
@@ -25,27 +29,26 @@ export default function initAnalytics() {
...
       window.clearInterval(interval);
     }
   }, 200);
 
-  [...document.querySelectorAll('[data-ga=error]')].map(trackElement);
-
   document.body.addEventListener('click', ({ target }) => {
     if (target.getAttribute('data-ga') === 'click') {
-      trackElement(target);
+      trackElement(target, target.getAttribute('data-ga'));
```
rmccar, line 41: In this case wouldn't this always be "click"?

src/components/tabs/_macro.njk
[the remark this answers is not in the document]
alessioventuriniAND: I'll remove all of them :)

src/js/analytics.js
```diff
@@ -1,22 +1,26 @@
 import domready from './domready';
 
-export let trackEvent = (event, data) => {
-  console.log(`[Analytics disabled] Event: ${event}`); // eslint-disable-line no-console
-  console.log(data); // eslint-disable-line no-console
+export let trackEvent = (data) => {
+  console.log('analitycs script connected'); // eslint-disable-line no-console
 };
+console.log('connected');
```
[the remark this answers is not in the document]
alessioventuriniAND: No I though it could be good for people to know that data ga tracking is enabled? If we don't like it I can delete it though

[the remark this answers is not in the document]
alessioventuriniAND: Removed :)

src/js/analytics.js
```diff
@@ -1,19 +1,16 @@
...
-    console.log('GTM active');
-    window.dataLayer = window.dataLayer || [];
-    trackEvent = (data) => {
-      console.log('Data sent to Data Layer');
-      window.dataLayer.push({ data });
-    };
-  }
-}, 300);
+if (window.google_tag_manager !== 'undefined') {
+  console.log('GTM active');
```
rmccar, line 7: Didn't we think that this would be better to be worded like this?
```suggestion
  console.log('Google analytics active');
```

src/js/analytics.js
```diff
@@ -1,19 +1,16 @@
 import domready from './domready';
 
-export let trackEvent = (data) => {
-  console.log('analitycs script connected'); // eslint-disable-line no-console
+export let trackEvent = () => {
+  console.log('Google analytics not connected'); // eslint-disable-line no-console
 };
-console.log('connected');
-setTimeout(() => {
```
rmccar: I think this might be necessary because the user could be on the page for a bit before accepting the cookies

rmccar (approved):

PR #2879, opened 2023-10-31
CTA button component invokes :active with keyboard

adi-unni:

Changed styles and added javascript to the button so that link button which is a a ref styled as a button now has the same behaviours as the "button" buttons. When pressing the link-button in the CTA and otherwise, it now depresses and comes back.

### How to review this PR

Navigate to Call to Action button and click on the button with mouse and by using by using keyboard space or enter. Visual functionality should be the same between using a mouse and using a keyboard.

### Checklist

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -78,13 +78,15 @@ $button-shadow-size: 3px;
     }
   }
 
-  &:focus:hover:not(:active) &__inner {
+  &:focus:hover:not(:active, .active) &__inner {
     background: var(--ons-color-focus-dark);
   }
 
   // When down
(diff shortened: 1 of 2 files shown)


rmccar:
Theres currently an issue with the colour of the arrow when the button is active and pressed down

Screenshot 2023-11-02 at 09 50 11

adi-unni:
> Theres currently an issue with the colour of the arrow when the button is active and pressed down
> 
> Screenshot 2023-11-02 at 09 50 11

rmccar (approved):

precious-onyenaucheya-ons (approved):

PR #2887, opened 2023-11-06
Add max width to paragraphs

precious-onyenaucheya-ons:
fixes https://github.com/ONSdigital/ons-service-manual-website/issues/64
I have set max width to 685px for paragraphs, lists and headings in the design system

### How to review this PR

Check that the width of all paragraphs, list and headings in the DS does not exceed 685ox.
### Checklist


rmccar (commented):
I've noticed a few things looking at the visual test results that look odd to me now, and that we might need to do something about. Might be something to discuss with Joe to see what he thinks.

Panels and interviewers notes look too wide for the content and probably also need to be limited:
Screenshot 2023-11-06 at 12 48 57

Screenshot 2023-11-06 at 12 49 12

Featured articles now look a bit too narrow for the page and should probably still be full width:
Screenshot 2023-11-06 at 12 49 59

Should text areas still be this wide?:
Screenshot 2023-11-06 at 12 49 42

Issue #2869, opened 2023-10-24
Details components not showing contents when printed


rmccar:
On eQ currently when using details components to hold some text that text is not being printed unless the details component is already open when printing.

I have already tested one solution of opening all the print dialogs when the print dialog is opened (https://github.com/ONSdigital/design-system/pull/2791) but there is issues with this solution as they would then stay open unless you close them all when the print dialog is closed, which would close any ones that were already open. It would be complicated to fix that. 

The proposed solution is to revert the details components to their original implementation of using js to open and close them. A lot of the work done in this PR (https://github.com/ONSdigital/design-system/pull/2013) would need to be reverted. This should allow us to use the print.css to show/hide the contents on print instead of having the functionality hidden behind the html details element. For non js users the contents would be displayed on the page as paragraphs that couldn't be closed as was before this PR change went in. We need to make sure we take into account any accessibility issues brought up in the PRs. And this will need DAC testing again.

JNATION85:
Is this good enough to add the the print.css file

@media print {
  details {
    open: true;
  }
}

rmccar:
> Is this good enough to add the the print.css file
> 
> @media print { details { open: true; } }

Nah unfortunately it isn't that simple, needs some js to get it to work

PR #2944, opened 2023-11-16
Revert details component to JS version to fix printing issue

rmccar:
Fixes: #2869 

This reverts the details component to use the previous JS implementation. This allows us to fix an issue of details components not showing the contents when printed. It will mean that users that have JS turned off will now just see the title and contents as a heading and paragraph and not have a functioning details component.

Because we have switched back to the JS implementation there is a class `ons-details--initialised` set by the JS when JS is on that I was able to use to set print styles on when printing.

This will need a DAC test but I believe this should cover all the current DAC recommendations we have at the moment.

Ive also made the accordion example a bit wider because it was looking a bit narrow and wrapping.

### How to review this PR

- Check details and accordion components look and work as before with JS turned on
- Check details and accordion components now display just a heading and paragraph when printed with JS turned on
- Check details and accordion components now display a heading and paragraph for users with JS turned off
- Check the components still work with screen readers

### Checklist


alessioventuriniAND (commented):
This all looks great! I would only add a test to confirm that only the clicked onsDetail component opens if more than one are displayed in the same page. Apart from that this looks good to me :)

alessioventuriniAND (approved):

precious-onyenaucheya-ons (approved):

rmccar:
[Bug: Preview questions not opening sections on save/print answers](https://trello.com/c/D9PUSJOA/5832-bug-preview-questions-not-opening-sections-on-save-print-answers)

Issue #2945, opened 2023-11-17
Selecting an input prefix/suffix should show the tooltip and put focus on the input


adi-unni:
GDS have evidence that some users try to click their input prefix/suffix. ONS input prefixes/suffixes are styled similar to a secondary buttons which would further increase the affordance and the likelihood of this behaviour. Putting focus on the input when they are selected would replicate the standard label behaviour.

PR #2947, opened 2023-11-17
Selecting an input prefix/suffix should should focus the associated input

adi-unni:

Fixes: #2945

GDS user research found that users often misclicked the icon to the left of an input. As an enhancement, this change changes the focus from the `<abbr>` element to the associated input.

### How to review this PR

Click on the little pound sign (the prefix/suffix) and check that the focus get set to the input.

### Checklist


src/components/input/_macro.njk
```diff
@@ -76,7 +76,7 @@
                     {% set prefixClass = " ons-input-type--prefix" %}
                 {% endif %}
 
-                <span class="ons-input-type{{ prefixClass }}">
+                <span class="ons-input-type{{ prefixClass }}{% if params.prefix or params.suffix %} ons-js-input-container-abbr {% endif %}">
```
rmccar, line 79: Need to remove this space or we will introduce spaces into the class list
```suggestion
                <span class="ons-input-type{{ prefixClass }}{% if params.prefix or params.suffix %} ons-js-input-container-abbr{% endif %}">
```

src/components/input/_macro.spec.js
```diff
@@ -3,7 +3,7 @@
 import * as cheerio from 'cheerio';
 
 import axe from '../../tests/helpers/axe';
-import { renderComponent, templateFaker } from '../../tests/helpers/rendering';
+import { renderComponent, templateFaker, setTestPage } from '../../tests/helpers/rendering';
```
rmccar, line 6: I don't think adding this import is needed

rmccar (approved):
Looks good, although make sure you get 3 reviews on this one because also worked on this a bit

precious-onyenaucheya-ons (approved):

alessioventuriniAND (approved):

PR #2958, opened 2023-11-24
Trim spaces and line breaks beside icons

### LINKED ISSUE #2873: HTML not being minimised causing extra whitespace
HTML not being minimised is causing extra whitespace such as in this example here. You can see there is a gap between the text and the padding that I don't think should be there. This is also an issue on header service links.
I found this on [this](https://service-manual.ons.gov.uk/design-system/components/header/example-header-external-for-surveys) example on the service manual 

Screenshot 2023-10-25 at 12 40 39

precious-onyenaucheya-ons:
Fixes #2873 
Currently there extra spaces beside icons in the Design System. Following the suggestion made in the issue linked, I made use of the `--` syntax in Nunjucks to trim any trailing whitespace or line breaks.

I have also updated the external link component to use the onsIcon rather than `<svg>`. Tests to cover this scenario have been added as well.

### How to review this PR

Check components which include an icon if there is an extra white space. Use the image in the issue linked to the PR as a reference

### Checklist


rmccar:
The padding around the pagination component and breadcrumbs component now needs updating. As this issue only affects places where an SVG is placed after text and not before the padding is no longer consistent between the two.

Screenshot 2023-11-29 at 15 30 10

Screenshot 2023-11-29 at 15 30 30

I also think we need to run these changes past Dina to see if shes happy with the padding or prefers the padding before the change. If she prefers that latter than I still think we should fix this but just increase the actual margin.

Also we are no longer minifying the HTML, just removing line breaks before svg elements so the PR title and description should be updated.

There are other benefits to minification are we happy to not minify? We need to research and discuss this because we could easily have the minification you had before alongside the regex which could be the best of both.

rmccar:
As part of this PR the external link component can now be updated to use the icon component instead of hardcoding the svg. The external link icon might need to be added to the icon macro to do that

rmccar (commented):
Screenshot 2023-12-06 at 09 47 47

The changes seem to have increased the space between the external link icon and the link itself

precious-onyenaucheya-ons:
> Screenshot 2023-12-06 at 09 47 47
> 
> The changes seem to have increased the space between the external link icon and the link itself

This is because of  the span that contains `&nbsp;` which was added. So I'd imagine the space is expected

rmccar:
> > Screenshot 2023-12-06 at 09 47 47
> > The changes seem to have increased the space between the external link icon and the link itself
> 
> This is because of the span that contains `&nbsp;` which was added. So I'd imagine the space is expected

Yes but the space has increased which is not what we want to happen. The `&nbsp` has always been there to stop the line breaking at that point. Ideally we want the space to stay the same including keeping the `&nbsp` but use the icon component. The other option is to look into other ways of stopping the line breaking at that point. But overall we don't want the space between the icon and text to increase

src/components/icon/_macro.njk
```diff
@@ -1,91 +1,91 @@
...
         </svg>
-    {% elif params.iconType == "search" %}
+    {%- elif params.iconType == "search" -%}
         <svg class="{% if params.iconSize %}ons-icon ons-icon--{{ params.iconSize }}{% else %}ons-icon{% endif %}{{ iconClasses }}" viewBox="0 0 12 12" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" fill="currentColor">
             <path d="M11.86 10.23 8.62 6.99a4.63 4.63 0 1 0-6.34 1.64 4.55 4.55 0 0 0 2.36.64 4.65 4.65 0 0 0 2.33-.65l3.24 3.23a.46.46 0 0 0 .65 0l1-1a.48.48 0 0 0 0-.62Zm-5-3.32a3.28 3.28 0 0 1-2.31.93 3.22 3.22 0 1 1 2.35-.93Z"/>
         </svg>
-    {% elif params.iconType == "sort-sprite" %}
-        <svg id="sort-sprite{% if params.id %}-{{ params.id | lower }}{% endif %}" class="ons-icon{{ iconClasses }}" viewBox="0 0 12 19" xmlns="http://www.w3.org/2000/svg" focusable="false" fill="currentColor">
+    {%- elif params.iconType == "sort-sprite" -%}
+        <svg id="sort-sprite{%- if params.id -%}-{{ params.id | lower }}{%- endif -%}" class="ons-icon{{ iconClasses }}" viewBox="0 0 12 19" xmlns="http://www.w3.org/2000/svg" focusable="false" fill="currentColor">
```
rmccar, line 65:
```suggestion
        <svg id="sort-sprite{% if params.id %}-{{ params.id | lower }}{% endif %}" class="ons-icon{{ iconClasses }}" viewBox="0 0 12 19" xmlns="http://www.w3.org/2000/svg" focusable="false" fill="currentColor">
```

src/components/external-link/_macro.njk
```diff
@@ -1,10 +1,9 @@
...
         <span class="ons-external-link__text">
         {{- params.linkText | safe -}}
-        </span><span class="ons-external-link__icon">&nbsp;<svg class="ons-icon" viewBox="0 0 12 12" xmlns="http://www.w3.org/2000/svg" focusable="false" aria-hidden="true">
-        <path d="M13.5,9H13a.5.5,0,0,0-.5.5v3h-9v-9h3A.5.5,0,0,0,7,3V2.5A.5.5,0,0,0,6.5,2h-4a.5.5,0,0,0-.5.5v11a.5.5,0,0,0,.5.5h11a.5.5,0,0,0,.5-.5v-4A.5.5,0,0,0,13.5,9Z" transform="translate(-2 -1.99)"/>
-        <path d="M8.83,7.88a.51.51,0,0,0,.71,0l2.31-2.32,1.28,1.28A.51.51,0,0,0,14,6.49v-4a.52.52,0,0,0-.5-.5h-4A.51.51,0,0,0,9,2.52a.58.58,0,0,0,.14.33l1.28,1.28L8.12,6.46a.51.51,0,0,0,0,.71Z" transform="translate(-2 -1.99)"/>
-    </svg></span><span class="ons-external-link__new-window-description ons-u-vh"> ({{- params.newWindowDescription | default("opens in a new tab") -}})</span></a>
+        </span><span class="ons-external-link__icon">&nbsp;{{- onsIcon({"iconType": 'external-link'}) -}}
+        </span><span class="ons-external-link__new-window-description ons-u-vh"> ({{- params.newWindowDescription | default("opens in a new tab") -}})</span></a>
```
rmccar, line 8:
```suggestion
        </span><span class="ons-external-link__new-window-description ons-u-vh">({{- params.newWindowDescription | default("opens in a new tab") -}})</span></a>
```

__snapshots__/layout/_template.spec.js.snap
```diff
@@ -1615,10 +1449,10 @@ exports[`base page template matches the full configuration snapshot 1`] = `
                     
                             
     <a href="#0" class="ons-external-link" target="_blank" rel="noopener">
-        <span class="ons-external-link__text">UK Statistics Authority</span><span class="ons-external-link__icon">&nbsp;<svg class="ons-icon" viewBox="0 0 12 12" xmlns="http://www.w3.org/2000/svg" focusable="false" aria-hidden="true">
```
rmccar, line 1618: I think this `aria-hidden="true"` `fill="currentColor"` change needs to be reverted. Just edit the icon in the icon component

rmccar (commented):
Couple of minor things

rmccar (approved):

alessioventuriniAND (approved):

Issue #2835, opened 2023-10-17
[Bug]: ons-grid__col--sticky CSS functionality is broken


AlanHONS:
### What happened?

Attempting to use `ons-grid__col--sticky` CSS class did not result in the expected behaviour

### What was the expected outcome?

Properly wrapped  `ons-grid__col--sticky` content was expected to switch from original/relative to floating/fixed position onscreen, depending on whether that element's original position was onscreen or not.

### How is this problem impacting your service?

Design of planned IDS Website page wants to make use of a floating 'Back to top' link using pure CSS functionality already in the design system. Without this functionality user's will how no quick way to return to alternate page and jump link choices.

### What list of steps can be followed to reproduce the bug or issue?

After discovering issue initially within out IDS Website prototype I created a new simplified example page under latest [65.2.5](https://github.com/ONSdigital/design-system/releases/tag/65.2.5) of the design system at [simplified code example page](https://github.com/ONSdigital/ixd-ons-prototype-playpen/blob/main/src/prototypes/sticky/index.njk)

Right column content on that page should be sticky and stay onscreen when the page is scrolled. However doesn't currently.

Joe Nation found that root cause is that base page parent DIV of `.ons-page` has a CSS style of `overflow-x: hidden;`. 
Which prevents child right column CSS `position: sticky;` from working when applied via `ons-grid__col--sticky@m`

Joe Nation recommended fix is to override `overflow-x: hidden;` to `overflow-x: inherit;` or similar on `.ons-page`. Which does restore `ons-grid__col--sticky` CSS class functionality. 
But needs investigation of why `overflow-x: hidden;` is being applied to `.ons-page`

### Example

https://github.com/ONSdigital/ixd-ons-prototype-playpen/blob/main/src/prototypes/sticky/index.njk

Screenshot of page + applied fix when overriding `.ons-page` > `overflow-x: hidden;`
![image](https://github.com/ONSdigital/design-system/assets/114412177/f75b1c2d-3ea0-4005-b2d0-63ea0e30083f)

### What version of the ONS Design System are you using?

65.2.5

### What devices are you seeing the problem on?

- [X] Desktop
- [X] Mobile
- [ ] Tablet

### Device details

ONS PC and ONS Macbook

### What operating systems are you seeing the problem on?

- [X] Windows
- [X] macOS
- [ ] Android
- [ ] iOS
- [ ] Other

### Operating system details

Windows 10 and macOS Sonoma 14.0

[...]

PR #2960, opened 2023-11-27
Grid sticky functionality fixed in full pages

adi-unni:

Fixes: #2835 

Previously, the ons grid had the functionality to make columns sticky i.e. have the object have fixed positioning on the screen. This worked when the grid was not contained within any page however was broken when the class was applied to an object within a full page such as the base page template. The change fixes this to work within the full page and adds an additional visual regression test for testing the sticky class in full pages

Additional documentation about this class is also needed and the new VR example can be used for demonstration see ticket: #2969

### How to review this PR

Navigate to the new regression test within the table component and check that the contents section on the page remains in place whilst scrolling.

### Checklist


alessioventuriniAND (approved):

src/components/table-of-contents/example-table-of-contents-sticky-full-page.njk
```diff
@@ -0,0 +1,114 @@
...
+        <div class="ons-grid ons-js-toc-container">
+            <div class="ons-grid__col ons-grid__col--sticky@m ons-col-4@m">
+                {{
+                    onsTableOfContents({
+                        "title": 'Contents',
+                        "ariaLabel": 'Sections in this page',
+                        "itemsList": [
+                            {
+                                "url": '#section1',
+                                "text": 'What is the benson?'
```
rmccar, line 47: benson?

src/components/table-of-contents/example-table-of-contents-sticky-full-page.njk
```diff
@@ -0,0 +1,114 @@
...
+title: "Example: Table of contents on full page"
+layout: ~
+---
+{% from "components/table-of-contents/_macro.njk" import onsTableOfContents %}
+
+{% extends "layout/_template.njk" %}
+
+{% set pageConfig = {
+    "header": {
+        "title": "What is the census?"
```
rmccar, line 11: Might need to check with Oscar but I think census should have a upper case C

rmccar (commented):
Other than these two minor content things, looks good

src/components/table-of-contents/example-table-of-contents-sticky-full-page.njk
adi-unni: Good spotting. Needed a change of title and accidentally left it there. Was probably thinking about sleeping and benson for beds :)

rmccar:
Actually maybe we should add a test to this. One that loads the page example, scrolls to the bottom and checks that the table of contents is still in view. Will stop the bug breaking again.

adi-unni:
> Actually maybe we should add a test to this. One that loads the page example, scrolls to the bottom and checks that the table of contents is still in view. Will stop the bug breaking again.

Okay sure, I will work on that test and add it in

src/components/table-of-contents/example-table-of-contents-sticky-full-page.njk
rmccar: Cool cheers. Yeah I noticed that, some upper case and some lower case. I feel like it could be something that might need an upper case C but I could be wrong

alessioventuriniAND (approved):

alessioventuriniAND:
All look sgood to me :) One note for next time , I have noticed that your branch doesn't follow the convention stated in the docs [here](https://confluence.ons.gov.uk/display/SDC/Ways+of+Working+details). You can leave this branch names as it is but let's make sure to all stick to the rule defined in the docs :)

src/components/table-of-contents/toc.spec.js
```diff
@@ -52,10 +52,120 @@ describe('script: table-of-contents', () => {
...
+                </section>
+            </div>
+        </div>
+    </div>
+    `,
+      'main',
+    );
+  });
+
+  it('renders a fixed table of contents in full page and scrolls', async () => {
```
rmccar, line 148: This line should change to make the description make more sense. Maybe something like, "when the ons-grid__col--sticky is used table of contents is rendered in a fixed position staying in view on the page when scrolling to the bottom"

rmccar:
Other than my comments still needing to be addressed and the tests failing this now looks good

src/components/table-of-contents/example-table-of-contents-sticky-full-page.njk
adi-unni: Haven't heard back yet, will spin this off into another ticket

src/components/table-of-contents/toc.spec.js
adi-unni: Understood the sentiment, however I think that test descriptions should be kept to within 70 characters for conciseness and clarity. How about "when ons-grid__col--stick is used, ToC stays fixed while scrolling"

rmccar: I understand that, but we haven't yet discussed or set any char limits or standards for our test descriptions. To be fair that is a bit better but I don't think thats quite descriptive enough which I would say is more important that keeping it under 70 characters. If you do want to keep it that short then you could break off the when bit into a `describe` but really that is used for grouping which isn't really needed here. We have a card to look into rewriting our test descriptions anyway. I do prefer something like "when the "ons-grid__col--sticky" class is used, then the ToC stays in a fixed position while scrolling" because its more explicit.

rmccar (approved):

Issue #2955, opened 2023-11-23
[Bug]: Section Navigation list item classes


alessioventuriniAND:
### What happened?

When adding a class to the ItemList within the Sections of the SectionNavigation component the below is added to the class attribute of the listItem 
<img width="456" alt="Screenshot 2023-11-23 at 11 06 22" src="https://github.com/ONSdigital/design-system/assets/112873190/dff843a7-6dc9-4913-ae1f-b952cee14f70">
This is cause by this line of code 
` <li class="ons-section-nav__item{% if item.classes %} ' ' + {{ item.classes }}{% endif %}{% if isCurrent == true %} ons-section-nav__item--active{% endif %}">`
Which should be 
` <li class="ons-section-nav__item{% if item.classes %}{{ ' ' + item.classes }}{% endif %}{% if isCurrent == true %} ons-section-nav__item--active{% endif %}">`

### What was the expected outcome?

<img width="519" alt="Screenshot 2023-11-23 at 11 10 24" src="https://github.com/ONSdigital/design-system/assets/112873190/0635cac5-620c-43e7-8853-25832178b98f">

### How is this problem impacting your service?

We need to deploy a new template on the IDS website and this is currently blocking it. 

### What list of steps can be followed to reproduce the bug or issue?

Add a class to the listItem of the Sections parameter of the sectionNavigation component. 

### What version of the ONS Design System are you using?

66.0.1

### What devices are you seeing the problem on?

Desktop, Mobile, Tablet

PR #2961, opened 2023-11-30
Updated Section Navigation list item classes

SriHV:
This fixes : #2955 
This PR addresses the spacing issue when adding a class to the ItemList within the Sections of the SectionNavigation component the below is added to the class attribute of the listItem.
image

### How to review this PR

To test this would be to add a class to one of the sections in the section navigation example and check that the "+ ' '" is not added in the html on your branch and compare that to `main` and see that it is added there.

### Checklist

### DIFF
--- a/src/components/access-code/_macro.njk
+++ b/src/components/access-code/_macro.njk
@@ -2,7 +2,7 @@
 {% from "components/panel/_macro.njk" import onsPanel %}
 
 {% macro onsAccessCode(params) %}
-    {% set groupSize = params.groupSize | default(5) %}
+    {% set groupSize = params.groupSize | default(4) %}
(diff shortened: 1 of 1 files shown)


src/components/access-code/_macro.njk
```diff
@@ -2,7 +2,7 @@
 {% from "components/panel/_macro.njk" import onsPanel %}
 
 {% macro onsAccessCode(params) %}
-    {% set groupSize = params.groupSize | default(4) %}
```
rmccar, line 5: Was this change added by accident?

rmccar (approved):

precious-onyenaucheya-ons (approved):

PR #2977, opened 2023-12-12
add search functionality to ds

### LINKED ISSUE #2872: Implement component search in the DS front end
Since moving the documentation out of the DS and into craft CMS the front end for the DS is now a bare bones list of links. Would make it a lot quicker to use for devs if we enabled the search functionality of the service manual to get to components more quickly. And removing the padding around full page examples in this front end would make it easier to compare examples.

- The search will work the same as the one on the service manual front page and return suggestions based on what you type in.
- Search will need to work for components, patterns and foundations as the one on the service manual does.
- The autosuggest will return the link to the component page that has the list of example links.
- The search box will need to be added either next to or underneath the Design System title on the root page of the DS.

precious-onyenaucheya-ons:
Fixes #2872 
This ticket implements search functionality in the DS local frontend.
### How to review this PR

Check that there is a search input at the root of the DS local frontend. And that it implements autosuggest as well.
### Checklist


lib/dev-server.js
```diff
@@ -30,7 +32,11 @@ const port = process.env.TEST_PORT ?? 3010;
   });
 
   app.get('/', async (req, res) => {
-    const output = renderPageList(pages, title);
+    const nunjucksEnvironment = createNunjucksEnvironment();
+    const PROJECT_PATH = process.cwd();
+    const pageSearch = renderSearch(`${PROJECT_PATH}/templates/layout/_dsTemplate.njk`, nunjucksEnvironment);
```
rmccar, line 37: This path either seems to be wrong or `_dsTemplate.njk` is missing

lib/dev-server.js
```diff
@@ -30,7 +32,11 @@ const port = process.env.TEST_PORT ?? 3010;
...
+    const output = renderPageList(pages, title, pageSearch);
+    await buildSearchIndex();
```
rmccar, line 39: Config for the search should probably be added in config.indexPage.js because that is where the rest of the page is configured

lib/render-page-list.js
```diff
@@ -1,4 +1,4 @@
-export default function renderPageList(pages, title) {
+export default function renderPageList(pages, title, output = '') {
```
rmccar, line 1: 'output' isn't a good variable name here, its very undescriptive

src/static/examples/data/search-index.json
```diff
@@ -0,0 +1 @@
...,{"en":"save-and-sign-out","url":"/patterns/save-and-sign-out","category":""},{"en":"send-a-web-form","url":"/patterns/send-a-web-form","category":""},{"en":"send-and-reply-to-messages","url":"/patterns/send-and-reply-to-messages","category":""},{"en":"survey-interstitial-page","url":"/patterns/survey-interstitial-page","category":""},{"en":"survey-section-introduction-page","url":"/patterns/survey-section-introduction-page","category":""},{"en":"address-output","url":"/foundations/address-output","category":""},{"en":"grid","url":"/foundations/grid","category":""},{"en":"headings","url":"/foundations/headings","category":""},{"en":"highlighting","url":"/foundations/highlighting","category":""},{"en":"icons","url":"/foundations/icons","category":""},{"en":"links","url":"/foundations/links","category":""},{"en":"lists","url":"/foundations/lists","category":""},{"en":"page-template","url":"/foundations/page-template","category":""},{"en":"paragraphs","url":"/foundations/paragraphs","category":""},{"en":"quote","url":"/foundations/quote","category":""},{"en":"text-indent","url":"/foundations/text-indent","category":""},{"en":"type-scale","url":"/foundations/type-scale","category":""}]
```
rmccar, line 1: Shouldn't this file be git ignored because it should be created automatically?

rmccar (changes requested):
Unable to spin up the DS on this branch. It falls over because theres a missing `_dsTemplate.njk` file. Im not sure if this is meant to be a problem with the path or a file that hasn't been pushed

rmccar (commented):
Be nice to have a label on the search box. I know its only the dev front end but at the moment its just an input with no explanation what it is for. Also the results are pushing the links down the page

Screenshot 2023-12-13 at 10 44 32

src/layout/_dsTemplate.njk
```diff
@@ -0,0 +1,34 @@
...
+        "searchButton": {
+            "visuallyHideButtonText": true,
+            "text": 'Search',
+            "iconType": 'search'
+        }
+    })
+ %}
+
+{{ onsAutosuggest({
+    "id": "search input-autosuggest",
```
rmccar, line 20: Shouldn't have spaces in ids and I think something like this would be better

```suggestion
    "id": "component-pattern-autosuggest",
```

rmccar:
> Be nice to have a label on the search box. I know its only the dev front end but at the moment its just an input with no explanation what it is for. Also the results are pushing the links down the page
> 

Even just adding a placeholder to the input instead of a label would be good

src/layout/_dsTemplate.njk
```diff
@@ -0,0 +1,34 @@
+{% from "components/input/_macro.njk" import onsInput %}
```
rmccar, line 1:
```suggestion
```

src/layout/_dsTemplate.njk
```diff
@@ -0,0 +1,33 @@
+{% from "components/autosuggest/_macro.njk" import onsAutosuggest %}
+
+{% set inputWithSearch = {
+         "id": 'search-field',
+         "width": '20',
+        "label": {
+            "text": "Search this service",
```
rmccar, line 7:
```suggestion
            "text": "Search patterns and components",
```
Think this is better wording for this

src/layout/_dsTemplate.njk
```diff
@@ -0,0 +1,33 @@
...
+
+{% set inputWithSearch = {
+         "id": 'search-field',
+         "width": '20',
+        "label": {
+            "text": "Search",
+            "id": "search-field-label",
+            "classes": "ons-u-pl-m"
+        },
+        "classes": "ons-input-search ons-input-search--icon ons-input-search--dark",
```
rmccar, line 11:
```suggestion
        "classes": "ons-input-search ons-input-search--icon ons-input-search--dark ons-u-mb-s",
```
Adding this class improves the spacing

rmccar (approved):

alessioventuriniAND (approved):

PR #2982, opened 2023-12-28
Subnavigation rendered only when needed

precious-onyenaucheya-ons:

I have introduced a new parameter, removeHorizontalSubNav, to the navigation component. The presence of the sub-navigation HTML is now contingent on the removeHorizontalSubNav parameter being set to false. I have provided an example to illustrate this functionality. Additionally, I have written tests to cover this specific scenario.

### How to review this PR

Check that the html is not rendered when the removeHorizontalSubNav is set to true

### Checklist


src/components/navigation/_macro.njk
```diff
@@ -58,21 +58,23 @@
...
-                                {% if item.title %}{{ item.title }}
-                                {% elif item.text %}{{ item.text }}
-                                {% endif %}
-                            </a>
-                        </li>
-                    {% endfor %}
-                </ul>
-            </div>
-        </nav>
+        {% if not params.navigation.subNavigation.hideDisplay %}
```
alessioventuriniAND, line 61: I think this parameter needs to be more descriptive as it's needed to hide only the horizontal subnavigation that is displayed at desktop size and not on mobile view. The mobile subnavigation is still displayed. 

Maybe we could call this parameter `hideHorizontalSubNav` or maybe `hideDesktopSubNav` ?

alessioventuriniAND (changes requested):

src/components/navigation/_macro.njk
rmccar: Maybe even something like `hideHorizontalSubNav` or something because we aren't hiding it

src/components/navigation/navigation.spec.js
```diff
@@ -230,3 +290,19 @@ describe('script: navigation', () => {
...
+
+describe.each([['sub', EXAMPLE_NAVIGATION_WITH_HIDDEN_SUBNAVIGATION, '.ons-navigation--sub']])(
+  'level: %s navigation',
+  (_, params, navEl) => {
+    describe('when hideDisplay is set to true', () => {
+      beforeEach(async () => {
+        await setTestPage('/test', renderComponent('header', params));
+      });
+
+      it('does not render the sub-nav element', async () => {
```
rmccar, line 302: We arent not rendering it, we are completely removing it

src/components/navigation/navigation.spec.js
```diff
@@ -230,3 +290,19 @@ describe('script: navigation', () => {
     },
   );
 });
+
+describe.each([['sub', EXAMPLE_NAVIGATION_WITH_HIDDEN_SUBNAVIGATION, '.ons-navigation--sub']])(
+  'level: %s navigation',
+  (_, params, navEl) => {
```
rmccar, line 296: Do we need the `navEl` var here considering we are only using it in one place?

src/components/navigation/navigation.spec.js
```diff
@@ -86,6 +86,66 @@ const EXAMPLE_NAVIGATION_WITH_SUBNAVIGATION = {
...
+                },
+              ],
+            },
+          ],
+        },
+      ],
+    },
+  },
+};
+
```
rmccar, line 148: We can probably simplify this config a bit more considering we are only testing for its existence in the html

src/components/navigation/_macro.njk
```diff
@@ -58,21 +58,23 @@
...
+                                    {% if item.title %}{{ item.title }}
+                                    {% elif item.text %}{{ item.text }}
+                                    {% endif %}
+                                </a>
+                            </li>
+                        {% endfor %}
+                    </ul>
+                </div>
+            </nav>
+            {% endif %}
```
rmccar, line 77: Indentation needs fixing here

src/components/navigation/navigation.spec.js
```diff
@@ -86,6 +86,66 @@ const EXAMPLE_NAVIGATION_WITH_SUBNAVIGATION = {
   },
 };
 
+const EXAMPLE_NAVIGATION_WITH_HIDDEN_SUBNAVIGATION = {
```
rmccar, line 89: Would be good to rename this because the sub navigation isn't being hidden its being removed

src/components/navigation/navigation.spec.js
SriHV: We are using it here, should it be removed?
  it('removes the sub-nav element', async () => {
        const hasSubNavEl = (await page.content()).includes(navEl);
        expect(hasSubNavEl).toBe(false);
      });
    });

rmccar: Im just saying do we need it as a variable? Its only used in one place so we could just set it there instead of passing it in

src/components/navigation/navigation.spec.js
```diff
@@ -230,3 +290,19 @@ describe('script: navigation', () => {
     },
   );
 });
+
+describe.each([['sub', EXAMPLE_NAVIGATION_WITH_HIDDEN_SUBNAVIGATION, '.ons-navigation--sub']])(
+  'level: %s navigation',
+  (_, params, navEl) => {
+    describe('when hideDisplay is set to true', () => {
+      beforeEach(async () => {
```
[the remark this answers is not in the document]
SriHV: I have updated it to `before` but  let me know if you want to move it :). I am not completely sure whether to keep or move it,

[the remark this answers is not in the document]
rmccar: Thinking about it I think it would be simpler to just add it to the test. But I cant see any of your changes, can you push them up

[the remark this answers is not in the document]
SriHV: Can you check now. I have pushed the changes.

src/components/navigation/navigation.spec.js
rmccar: We are still passing the '.ons-navigation--sub' class in on line 294. It should be removed from there too

src/components/navigation/navigation.spec.js
```diff
@@ -291,13 +291,17 @@ describe('script: navigation', () => {
   );
 });
 
-describe.each([['sub', EXAMPLE_NAVIGATION_WITH_SUBNAVIGATION_REMOVED, '.ons-navigation--sub']])(
-  'level: %s navigation',
-  (_, params) => {
-    it('removes the sub-nav element when hideHorizontalSubNav is set to true', async () => {
-      await setTestPage('/test', renderComponent('header', params));
+describe.each([['sub']])('level: %s navigation', () => {
```
rmccar, line 294:
```suggestion
describe.each()('level: sub navigation', () => {
```
Another change we can make here to simplify things

src/components/navigation/_macro.njk
rmccar: Sorry this should have said `removeHorizontalSubNav`

rmccar: Just need to add `removeHorizontalSubNav` to the `_macro-options` file

SriHV: I have added it in navigation section. Let me know if its ok

src/components/header/_macro-options.md
```diff
@@ -63,13 +63,14 @@
...
+| id                     | string                    | true     | The HTML `id` of the `<nav>` element. Used for the `aria-controls` attribute for the menu button displayed on small viewports. |
+| ariaLabel              | string                    | false    | The `aria-label` attribute added to the `<nav>` element. Defaults to “Main menu”.                                              |
+| itemsList              | array`<Item>`             | true     | Settings for an array of [list items](#item) for each navigation link                                                          |
+| currentPath            | string or array`<string>` | true     | The path of the current active page. Multiple paths can be provided using an array to highlight nested navigation.             |
+| toggleButton           | array`<ToggleButton>`     | true     | Settings for the navigation [menu toggle button](#togglebutton) displayed on small viewports                                   |
+| removeHorizontalSubNav | boolean                   | false    | Set to “true” to remove the sub navigation                                                                                     |
```
rmccar, line 73:
```suggestion
| removeHorizontalSubNav | boolean                   | false    | Set to “true” to remove the sub navigation from the HTML                                                                                    |
```

rmccar (approved):

alessioventuriniAND (approved):

Issue #1599, opened 2021-07-20
Back to top link


JovdP:
## What

Back to top link component

## Why

To Enable users to get back to the table of contents at the top of the page

> Explain why you think this feature or content should be added to the ONS Design System.
>
> - What evidence do you have that your ONS service needs it?
> - What evidence do you have that it meets the needs of the users of that service?
> - Have you checked if it’s already in the ONS Design System?
> - Have you considered creating a new variant of an existing feature instead of something new?

## Supporting material

Research from 2018 prototypes (DP to provide)

> Include links to any examples, research, prototypes or code that support your proposal.

## Contacts

> The Design System working group has a weekly meeting. We will invite you to one of these meetings to give a show and tell about your proposal and answer any questions. Please tell us who to invite to that meeting. We recommend involving at least your product manager, user researcher and designer.

Display: Take users back to top

armstrongb:
Current 'default state' of the back to top link:
![BackToTop-base](https://user-images.githubusercontent.com/22910336/126503379-25772e12-25e9-4763-b4de-fd8af559ed86.png)
>A key line was added based on research conducted around Area Profiles. The key line helps to highlight the navigation item.

Originating designer: Will Richards
Source file: Figma
Page context used in research (attached)

![DatasetLandingPage-996](https://user-images.githubusercontent.com/22910336/126504252-d82bba33-69f6-4ddd-841b-a2c61013afeb.png)

jrbarnes9:
@armstrongb Do you think this link could be "sticky" to avoid the visual repetition? I think we could probably still ensure it's focusable after each content block.

emilymaylawes:
Accessibility considerations:
- Must be an anchor link that targets an anchor # to the top of the page. Either the <h1> of the page or the <main> content (alternative to skip to content). In that case, keyboard focus is managed.
- Must be the last link of the <main> content so right at the end of the focus order.
- If sticky to the bottom or anywhere on the page, must not obscure any other content on all screen viewports and not obscure focus.
- Must not rely on JavaScript
- Should only have one on the page, rather than repeated
- If `scroll-behaviour` property in CSS is used, ensure we turn this off with `@prefers-reduced-motion` media query

adi-unni:
#### Technical side:

- It can be added as an overlay by adjusting the z-index
- It can be added as a component to be added at the bottom of the page
- It can be added to fit within the page contents on the navigation sidebar
[...]

PR #2990, opened 2024-01-09
⬆️  Back To Top Link  ⬆️ 

adi-unni:

Fixes: #1599 

The **back-to-top link** is a requested component for the Design System. The button takes you from wherever you are on the page to the top of the page or wherever you specify. 

The default options for both are _"Back to Top"_ taking you to _"#top"_ however, this can be modified to you to the contents with the description _"Scroll to contents"_ for example. To modify the text of the button, edit the _description parameter._ Similarly, to modify the jump location, edit the _anchor parameter._ When making a custom point to anchor, make sure to add an `id` to the element you want to jump to which matches the `id` in the anchor parameter. 

As per the requirements, the button will appear towards the bottom of the screen when the user scrolls two window heights down from the anchor div. When the user reaches the link on the page, it will take the form of the link on the page.

Note: Position of the text of the button when sticky is based on the position when it is a link on the on the page. This in turn is based on the container it is in.

### How to review this PR

Check out the back to top button examples and make sure they are working correctly. Also test resizing the window to make sure the BTT links margins move accordingly

### Checklist


balibirchlee:
On the second example the highlighting stays on after I use the back to top link, I think that after it goes back to top we would want the highlight to disappear?

This only happens on the full page version - not the first version.

image
image

balibirchlee:
Another small thing is where the page pulls you back to when linking to a specific point in the page - usually when I've seen section links used they put the bit you want to go back to at the natural reading height on the screen. Not sure if this might cause other issues or has already been considered!!

So as a user I'd expect it to go to a position roughly like this, rather than the 'Contents' title being right at the top of the page

image

balibirchlee (commented):
Sorry, should have added my comments here!

.github/workflows/check-labels.yml
```diff
@@ -17,6 +17,6 @@ jobs:
     steps:
       - uses: docker://onsdigital/github-pr-label-checker:v1.2.7
         with:
-          one_of: Breaking changes,Accessibility,Bug,Documentation,Dependencies,Enhancement,Example
+          one_of: Breaking changes,Accessibility,Bug,Documentation,Dependencies,Enhancement,Example,Component
```
rmccar, line 20: If we are adding component we should also add pattern here
```suggestion
          one_of: Breaking changes,Accessibility,Bug,Documentation,Dependencies,Enhancement,Example,Component,Pattern
```

src/components/back-to-top/_back-to-top.scss
```diff
@@ -0,0 +1,36 @@
+.ons-back-to-top {
+  > a > span {
```
rmccar, line 2: Instead of using `a` here and in the other places you could use `&__link`. Makes the code a bit more descriptive and otherwise that class doesn't seem to be being used.
You could even add a class to the span something like `ons-back-to-top__description`

src/components/back-to-top/btt.dom.js
```diff
@@ -0,0 +1,12 @@
+import domready from '../../js/domready';
+
+async function btt() {
```
rmccar, line 3: Think we should have a more descriptive name here rather than btt

src/components/back-to-top/btt.spec.js
```diff
@@ -0,0 +1,145 @@
+import { renderComponent, setTestPage } from '../../tests/helpers/rendering';
+
+describe('script: back-to-top', () => {
+  beforeEach(async () => {
+    await setTestPage(
+      '/test',
+      `
+        <div class="ons-back-to-top-track ons-container">
+        <div class="ons-grid">
+            <div class="ons-grid__col ons-col-6@m ">
```
rmccar, line 10:
```suggestion
            <div class="ons-grid__col ons-col-6@m">
```

src/components/back-to-top/btt.spec.js
```diff
@@ -0,0 +1,145 @@
...
+                Sit amet mattis vulputate enim nulla aliquet. Quis commodo odio aenean sed adipiscing diam donec adipiscing tristique. Vulputate odio ut enim blandit volutpat. Elit pellentesque habitant morbi tristique senectus et netus. Aliquet lectus proin nibh nisl condimentum id. A iaculis at erat pellentesque adipiscing commodo elit at. Quis ipsum suspendisse ultrices gravida dictum fusce. Sit amet mauris commodo quis imperdiet massa tincidunt. Adipiscing elit ut aliquam purus. A diam maecenas sed enim ut sem viverra aliquet eget. Enim neque volutpat ac tincidunt vitae. Ultricies leo integer malesuada nunc vel risus commodo viverra maecenas. Amet nisl suscipit adipiscing bibendum. Nunc pulvinar sapien et ligula ullamcorper malesuada proin. Nulla facilisi cras fermentum odio eu feugiat.
+            </div>
+            </div>
+        </div>
+        <div id="contents" class="ons-container">
+            <h1> Contents </h1>
+        </div>
+        <div class="ons-grid">
+            <div class="ons-grid__col ons-col-6@m ">
```
rmccar, line 32:
```suggestion
            <div class="ons-grid__col ons-col-6@m">
```

src/components/back-to-top/btt.spec.js
```diff
@@ -0,0 +1,145 @@
...
+                Sit amet mattis vulputate enim nulla aliquet. Quis commodo odio aenean sed adipiscing diam donec adipiscing tristique. Vulputate odio ut enim blandit volutpat. Elit pellentesque habitant morbi tristique senectus et netus. Aliquet lectus proin nibh nisl condimentum id. A iaculis at erat pellentesque adipiscing commodo elit at. Quis ipsum suspendisse ultrices gravida dictum fusce. Sit amet mauris commodo quis imperdiet massa tincidunt. Adipiscing elit ut aliquam purus. A diam maecenas sed enim ut sem viverra aliquet eget. Enim neque volutpat ac tincidunt vitae. Ultricies leo integer malesuada nunc vel risus commodo viverra maecenas. Amet nisl suscipit adipiscing bibendum. Nunc pulvinar sapien et ligula ullamcorper malesuada proin. Nulla facilisi cras fermentum odio eu feugiat.
+            </div>
+            </div>
+        </div>
+        <div id="contents" class="ons-container">
+            <h1> Contents </h1>
```
rmccar, line 29:
```suggestion
            <h1>Contents</h1>
```

src/components/back-to-top/btt.spec.js
```diff
@@ -0,0 +1,145 @@
...
+                Lectus nulla at volutpat diam ut venenatis tellus. Tellus rutrum tellus pellentesque eu tincidunt tortor. Purus sit amet volutpat consequat mauris nunc congue. Dignissim cras tincidunt lobortis feugiat vivamus at. Ac felis donec et odio pellentesque diam volutpat commodo. Arcu dui vivamus arcu felis. Pulvinar proin gravida hendrerit lectus a. Venenatis lectus magna fringilla urna porttitor rhoncus dolor purus. Maecenas sed enim ut sem viverra aliquet eget. Lacus laoreet non curabitur gravida arcu ac tortor. Laoreet sit amet cursus sit amet dictum. Maecenas accumsan lacus vel facilisis volutpat est velit.
+                </div>
+            </div>
+            <div class="ons-grid__col ons-col-6@m">
+            <div class="ons-pl-grid-col">
```
rmccar, line 20: Few places where you have 2 divs like this where the second one isn't indented enough.
This and the other changes should be in the njk file too

src/components/back-to-top/btt.spec.js
```diff
@@ -0,0 +1,145 @@
...
+                Sit amet mattis vulputate enim nulla aliquet. Quis commodo odio aenean sed adipiscing diam donec adipiscing tristique. Vulputate odio ut enim blandit volutpat. Elit pellentesque habitant morbi tristique senectus et netus. Aliquet lectus proin nibh nisl condimentum id. A iaculis at erat pellentesque adipiscing commodo elit at. Quis ipsum suspendisse ultrices gravida dictum fusce. Sit amet mauris commodo quis imperdiet massa tincidunt. Adipiscing elit ut aliquam purus. A diam maecenas sed enim ut sem viverra aliquet eget. Enim neque volutpat ac tincidunt vitae. Ultricies leo integer malesuada nunc vel risus commodo viverra maecenas. Amet nisl suscipit adipiscing bibendum. Nunc pulvinar sapien et ligula ullamcorper malesuada proin. Nulla facilisi cras fermentum odio eu feugiat.
+            </div>
+            </div>
+        </div>
+        <div class="ons-grid">
+            <div class="ons-grid__col ons-col-6@m ">
```
rmccar, line 51:
```suggestion
            <div class="ons-grid__col ons-col-6@m">
```

src/components/back-to-top/btt.spec.js
```diff
@@ -0,0 +1,145 @@
...
+                Lectus nulla at volutpat diam ut venenatis tellus. Tellus rutrum tellus pellentesque eu tincidunt tortor. Purus sit amet volutpat consequat mauris nunc congue. Dignissim cras tincidunt lobortis feugiat vivamus at. Ac felis donec et odio pellentesque diam volutpat commodo. Arcu dui vivamus arcu felis. Pulvinar proin gravida hendrerit lectus a. Venenatis lectus magna fringilla urna porttitor rhoncus dolor purus. Maecenas sed enim ut sem viverra aliquet eget. Lacus laoreet non curabitur gravida arcu ac tortor. Laoreet sit amet cursus sit amet dictum. Maecenas accumsan lacus vel facilisis volutpat est velit.
+                </div>
+            </div>
+
```
rmccar, line 60:
```suggestion

```

src/components/back-to-top/example-back-to-top.njk
```diff
@@ -0,0 +1,77 @@
...
+        </div>
+        </div>
+    </div>
+
+    {{ onsBackToTop (
+    {
+        "description" : "Scroll to contents",
+        "anchor" : "contents"
+    }
+    )}}
```
rmccar, line 72:
```suggestion
    {{ onsBackToTop ({
        "description" : "Scroll to contents",
        "anchor" : "contents"
    }) }}
```

src/components/back-to-top/example-full-page-back-to-top.njk
```diff
@@ -0,0 +1,197 @@
+---
+title: "Example: Long Page"
+layout: ~
+---
+{% extends "layout/_template.njk" %}
+{% from "components/list/_macro.njk" import onsList %}
+{% from "components/back-to-top/_macro.njk" import onsBackToTop %}
+{% from "components/image/_macro.njk" import onsImage %}
```
rmccar, line 8: Import not being used

src/components/back-to-top/example-full-page-back-to-top.njk
```diff
@@ -0,0 +1,197 @@
...
+{% from "components/related-content/_macro.njk" import onsRelatedContent %}
+{% from "components/share-page/_macro.njk" import onsSharePage %}
```
rmccar, line 10: Import not being used

src/components/back-to-top/example-full-page-back-to-top.njk
```diff
@@ -0,0 +1,197 @@
...
+                "text": 'News'
+            }
+        ]
+    },
+    "footer": {}
+} %}
+
+
+
+{% block main %}
```
rmccar, line 35: Contents of this block should be indented

rmccar:
Think this focus state might be a bit big. Also I think because of this we are losing the black underline that is used on other focused elements

Screenshot 2024-02-14 at 13 44 56

rmccar (commented):
Just some minor things I have noticed. Other than these seems good!

rmccar:
Just realised we are also missing a `_macro-options.md` file for this

src/components/back-to-top/btt.dom.js
adi-unni: Table of Content uses toc, do we need to change that too?

rmccar: Maybe, just when initially looking at your code I saw "btt" and thought it could be more obvious what that function was for

adi-unni: sure, ill change it for back to top right now, and we can ask around in a tech session for change of ToC

adi-unni:
> Another small thing is where the page pulls you back to when linking to a specific point in the page - usually when I've seen section links used they put the bit you want to go back to at the natural reading height on the screen. Not sure if this might cause other issues or has already been considered!!
> 
> So as a user I'd expect it to go to a position roughly like this, rather than the 'Contents' title being right at the top of the page
> 
> image

Talked to Joe about this and the decision was made to keep it as is. Doing this would confuse the user where they have jumped to if there was a say a one line bit of text above the jump location.

adi-unni:
> Think this focus state might be a bit big. Also I think because of this we are losing the black underline that is used on other focused elements
> 
> Screenshot 2024-02-14 at 13 44 56

Talked to Joe about the width and the width being full / almost full width is fine and as intended. To match the skip-to-content component, the text underline is being added back in on focus.

rmccar:
> > Think this focus state might be a bit big. Also I think because of this we are losing the black underline that is used on other focused elements
> > Screenshot 2024-02-14 at 13 44 56
> 
> Talked to Joe about the width being full / almost full width it is deemed to be as intended. To match the skip-to-content component, the text underline is being added back in on focus.

If you want to match skip-to-content wouldn't you want to have the whole thing highlighted?

src/components/back-to-top/_back-to-top.scss
```diff
@@ -0,0 +1,37 @@
...
+  &__sticky {
+    bottom: -1px;
+    position: fixed;
+    z-index: 10;
+    background: var(--ons-color-grey-15);
+    width: 100%;
+    height: fit-content;
+    left: 0;
+
+    > a {
```
rmccar, line 25: You can use `.ons-back-to-top__link` here

src/components/back-to-top/_back-to-top.scss
```diff
@@ -0,0 +1,37 @@
...
+  &__description {
+    margin-left: 0.2rem;
+  }
+
+  &__enabled {
+    background: none;
+    width: fit-content;
+    padding: 0.5em 0;
+
+    > a {
```
rmccar, line 11: and you can use .ons-back-to-top__link here, but I don't think this bit is having any affect

src/components/back-to-top/_macro.njk
```diff
@@ -0,0 +1,15 @@
+{% from "components/icon/_macro.njk" import onsIcon %}
+
+{% macro onsBackToTop(params) %}
+    <div class="ons-back-to-top ons-js-btt ons-back-to-top__enabled">
```
rmccar, line 4: `ons-js-btt` should be `ons-js-back-to-top` to be consistent with the other class names

rmccar (commented):
Few minor things here and I think the formatting of your examples needs looking at. Theres also a few of my previous comments that haven't been addressed

src/components/back-to-top/_macro.spec.js
```diff
@@ -0,0 +1,60 @@
+/** @jest-environment jsdom */
+
+import * as cheerio from 'cheerio';
+
+import axe from '../../tests/helpers/axe';
+import { renderComponent, templateFaker } from '../../tests/helpers/rendering';
```
rmccar, line 6: Import not being used

```suggestion
import { renderComponent } from '../../tests/helpers/rendering';
```

src/components/back-to-top/backToTop.spec.js
```diff
@@ -0,0 +1,145 @@
...
+    const backToTopEnabled = await page.$eval('.ons-back-to-top', (node) => node.classList.contains('ons-back-to-top__enabled'));
+    expect(backToTopEnabled).toBe(true);
+  });
+
+  it('changes width when the window is resized', async () => {
+    await page.setViewport({ width: 1000, height: 800 });
+    await page.evaluate(() => {
+      window.scrollTo(0, window.innerHeight * 2);
+    });
+    await page.waitForTimeout(250);
```
rmccar, line 105: `waitForTimeout` has been deprecated, do we need these waits?

If we really do then we can use `await new Promise((r) => setTimeout(r, 250));`

src/components/back-to-top/backToTop.spec.js
```diff
@@ -0,0 +1,145 @@
...
+    const previousWidth = await page.evaluate(() => {
+      const node = document.querySelector('.ons-back-to-top > a');
```
rmccar, line 107: These selectors can be replaced with `.ons-back-to-top__link`

rmccar:
Id recommend adding the eslint extension to your VScode and changing the "render whitespace" setting to "all" which will allow you to easily spot some of these things

src/components/back-to-top/backToTop.spec.js
adi-unni: Just done a lot of testing and you're correct. It does seem like they're redundant unless the browser operation of puppeteer is slow on the tests in Github, however assuming theyve been running without for so long it's probably not needed there either.

rmccar: They have been running, by deprecated I mean that it works but is deemed an older way of doing something that is on the way out. It wasn't immediately obvious to me that we needed them, although you could be right that we do need them when running them on actions. But what we can do is take them out and make a commit, see if the tests pass. Run them a few times to be sure.

adi-unni:
> > > Think this focus state might be a bit big. Also I think because of this we are losing the black underline that is used on other focused elements
> > > Screenshot 2024-02-14 at 13 44 56
> > 
> > 
> > Talked to Joe about the width being full / almost full width it is deemed to be as intended. To match the skip-to-content component, the text underline is being added back in on focus.
> 
> If you want to match skip-to-content wouldn't you want to have the whole thing highlighted?

Joe has confirmed this is what the intended behaviour should be. I will work on adding this in

src/components/back-to-top/example-back-to-top.njk
```diff
@@ -0,0 +1,37 @@
...
+        </div>
+        <div class="ons-grid__col ons-col-6@m">
```
rmccar, line 19:
```suggestion
        <div class="ons-grid__col ons-col-6@m">
```

src/components/back-to-top/backToTop.spec.js
```diff
@@ -0,0 +1,139 @@
...
+                Sit amet mattis vulputate enim nulla aliquet. Quis commodo odio aenean sed adipiscing diam donec adipiscing tristique. Vulputate odio ut enim blandit volutpat. Elit pellentesque habitant morbi tristique senectus et netus. Aliquet lectus proin nibh nisl condimentum id. A iaculis at erat pellentesque adipiscing commodo elit at. Quis ipsum suspendisse ultrices gravida dictum fusce. Sit amet mauris commodo quis imperdiet massa tincidunt. Adipiscing elit ut aliquam purus. A diam maecenas sed enim ut sem viverra aliquet eget. Enim neque volutpat ac tincidunt vitae. Ultricies leo integer malesuada nunc vel risus commodo viverra maecenas. Amet nisl suscipit adipiscing bibendum. Nunc pulvinar sapien et ligula ullamcorper malesuada proin. Nulla facilisi cras fermentum odio eu feugiat.
+              </div>
+            </div>
+          </div>
+          <div id="contents" class="ons-container">
+              <h1>Contents</h1>
```
rmccar, line 29:
```suggestion
            <h1>Contents</h1>
```

src/components/back-to-top/backToTop.spec.js
adi-unni: Ran the tests a couple times and seems like they are failing without some sort of wait. Adding the new Timeout commands in.

rmccar: Makes sense

rmccar (commented):
The "backToTop" files should be renamed "back-to-top" to be consistent with other filenames of the same type

adi-unni:
> > > > Think this focus state might be a bit big. Also I think because of this we are losing the black underline that is used on other focused elements
> > > > Screenshot 2024-02-14 at 13 44 56
> > > 
> > > 
> > > Talked to Joe about the width being full / almost full width it is deemed to be as intended. To match the skip-to-content component, the text underline is being added back in on focus.
> > 
> > 
> > If you want to match skip-to-content wouldn't you want to have the whole thing highlighted?
> 
> Joe has confirmed this is what the intended behaviour should be. I will work on adding this in

An update on this. This would not be a good solution for screens which are wide / have wide windows. Joe has requested additional UX work to be done to make a decision on this before it is implemented.

src/components/back-to-top/example-back-to-top.njk
```diff
@@ -0,0 +1,37 @@
+{% from "components/back-to-top/_macro.njk" import onsBackToTop %}
```
rmccar, line 1:
```suggestion
{% from "components/back-to-top/_macro.njk" import onsBackToTop %}
```

src/components/back-to-top/_macro.njk
```diff
@@ -0,0 +1,17 @@
...
+                <div>
+                    {{-
+                        onsIcon({
+                            "iconType": 'arrow-up',
+                            "iconSize": 'xs'
+                        })
+                    -}}
+                    <span class="ons-back-to-top__description">{{ params.description | default('Back to top') }}</span>
+                </div>
+            </a>
```
rmccar, line 15:
```suggestion
        <a href="#{{ params.anchor | default('top') }}" class="ons-back-to-top__link">
            <div>
                {{-
                    onsIcon({
                        "iconType": 'arrow-up',
                        "iconSize": 'xs'
                    })
                -}}
                <span class="ons-back-to-top__description">{{ params.description | default('Back to top') }}</span>
            </div>
        </a>
```

src/components/back-to-top/_back-to-top.scss
```diff
@@ -9,7 +9,7 @@
     padding: 0.5em 0;
 
     > .ons-back-to-top__link {
-      padding: 0;
+      padding: 0%;
```
rmccar, line 12: You wouldn't add a percent sign here it would just be `0`. But I dont think this css is doing anything anyway so can be removed. There doesn't seem to be any padding being added to that element without this

adi-unni: Aah, I accidentally left it in when there was padding. Now there is margin padding applied to another element.

src/components/back-to-top/example-back-to-top.njk
```diff
@@ -0,0 +1,37 @@
...
+        </div>
+    </div>
+	{{
```
rmccar, line 31: Some of this file uses tabs to indent, to keep it consistent we use spaces because tabs can be different widths depending on how they are rendered but spaces are always the same size. This way we can keep the indenting consistent. This line here is one example where I know the format looks correct in vscode but here it looks off because the tab has been rendered as 8 spaces. Id recommend checking "insert spaces when pressing tab" in your vscode settings

rmccar (approved):

alessioventuriniAND (approved):

Issue #2988, opened 2024-01-08
Longer titles too close to button


rmccar:
When using the header component if the header is particularly long and has words that have a space after a certain number of characters so that the title wraps at that point then its possible for the title to be too close to the button

<img width="1089" alt="Screenshot 2024-01-08 at 12 46 56" src="https://github.com/ONSdigital/design-system/assets/42928680/88e11a49-89ef-409e-b98e-2f09c57355a6">

JNATION85:
![Image](https://github.com/ONSdigital/design-system/assets/41283876/f271ee2c-5ba5-4b33-8093-d1eeddf4c5c4)

rmccar:
Technical implementation of this would be to remove the `ons-grid--gutterless` class from the header container grid on 172 in the header macro if there is a exit button specified.

PR #2998, opened 2024-01-19
adds conditional logic to add gutter between title and button solving…

balibirchlee:
… issue

Added conditional logic so gutterless scss tag is only present when there is no button provided in the params. When there is a button, the gutter provides space between it and the title fixing the issue. 

https://github.com/ONSdigital/design-system/issues/2988

### How to review this PR

Pull, yarn start, and then change the header params for the 'example-header-external-for-surveys' - if you make the header super long with no spaces then save it, you'll see that there is still a gap between the title and the button! 

### Checklist

### DIFF
--- a/src/components/header/_macro.njk
+++ b/src/components/header/_macro.njk
@@ -169,7 +169,7 @@
         </div>
         <div class="ons-header__main">
             <div class="ons-container{{ ' ons-container--full-width' if params.fullWidth }}{{ ' ons-container--wide' if params.wide }}">
-                <div class="ons-grid ons-grid--gutterless ons-grid--flex ons-grid--between ons-grid--vertical-center ons-grid--no-wrap">
+                <div class="ons-grid {{ 'ons-grid--gutterless' if not params.button }} ons-grid--flex ons-grid--between ons-grid--vertical-center ons-grid--no-wrap">
                     <div class="ons-grid__col ons-col-auto ons-u-flex-shrink">
                         {% if params.titleLogo.large %}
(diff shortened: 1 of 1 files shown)


src/components/header/_macro.njk
```diff
@@ -169,7 +169,7 @@
         </div>
         <div class="ons-header__main">
             <div class="ons-container{{ ' ons-container--full-width' if params.fullWidth }}{{ ' ons-container--wide' if params.wide }}">
-                <div class="ons-grid ons-grid--gutterless ons-grid--flex ons-grid--between ons-grid--vertical-center ons-grid--no-wrap">
+                <div class="ons-grid {{ 'ons-grid--gutterless' if not params.button }} ons-grid--flex ons-grid--between ons-grid--vertical-center ons-grid--no-wrap">
```
rmccar, line 172:
```suggestion
                <div class="ons-grid{{ ' ' + 'ons-grid--gutterless' if not params.button }} ons-grid--flex ons-grid--between ons-grid--vertical-center ons-grid--no-wrap">
```
Theres a space before and after this so if the class isn't set then it results in a double space in the class list. This syntax will stop this

rmccar:
Think we should add a test for this

balibirchlee:
> Think we should add a test for this

test added - just added one to check that the gutterless class is present if there's no buttton, but if preferred i could add another that tests the reverse too?

rmccar (commented):
We should also have to opposite of this test, where the class isn't added if the button is set

rmccar (approved):

PR #3036, opened 2024-02-05
Fix/3004/icon in footer not clickable

### LINKED ISSUE #3004: [Bug]: Icon in the template footer are not clickable
### What happened?

The icon in the footer are not clickable. This has been highlighted as an accessibility issue in the Dac report of the IDS website. 

### What was the expected outcome?

Icon in the footer to be clickable. 

### How is this problem impacting your service?

accessibility failing 

### What list of steps can be followed to reproduce the bug or issue?

You can look into the IDS live website and when trying to click on the icons in the footer they are not clickable. 

### Example

_No response_

### What version of the ONS Design System are you using?

67.0.4

### What devices are you seeing the problem on?

Desktop, Mobile, Tablet

### Device details

_No response_

### What operating systems are you seeing the problem on?

_No response_

### Operating system details

_No response_

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge, Microsoft Internet Explorer

### Browser details

_No response_

balibirchlee:

As the custom poweredBy accepts HTML and the IDS website is using this to render a more complicated set up than a simple icon image - it has been agreed that IDS would instead add the links to their existing html. Any other implementation would result in a breaking change, requiring either IDS to simplify their existing footer (i.e. providing only an image) or us to scope an enhancement to the current footer modelled after IDS's current set up.

Whether ourselves or IDS feel that this enhancement should be made is up for discussion, but is beyond the scope of this ticket.

As such, the ticket has been adjusted to focus on providing links for the existing default ONS icons which are responsive to the language selection (i.e. the Welsh language version which renders the Welsh variant of the default icon has a link to the Welsh site, and vice versa for the English language version).

fixes #3004 

### How to review this PR

Run locally and head to the footer component, checking **example footer cymraeg** for the Welsh language version and **example footer default** for the English language version.

### Checklist


src/components/footer/_macro.njk
```diff
@@ -158,9 +158,11 @@
...
                         {% endif %}
                         {% if not params.poweredBy %}
-                            <div class="ons-footer__poweredby ons-u-mb-m">
-                                {{ onsLogo | safe }}
-                            </div>
+                            <a class="ons-footer__poweredByLink" {% if lang == "cy" %}href="https://cy.ons.gov.uk/"{% else %}href="https://www.ons.gov.uk/"{% endif %}>
+                                <div class="ons-footer__poweredby ons-u-mb-m">
+                                    {{ onsLogo | safe }}
+                                </div>
+                            </a>
```
rmccar, line 165: Is this something we want to make mandatory? And the link not configurable by the user like we have in other places? I get that its always the same logo but I was expecting this to be controlled by params.
Feels like something we may need to have a discussion on with the rest of the team, I just wanted to check to see if there were any use cases for when a service might not want this

src/components/footer/_footer.scss
```diff
@@ -49,4 +49,8 @@
       margin-bottom: 0.5rem !important;
     }
   }
+
+  &__poweredByLink {
+    display: contents;
+  }
```
rmccar, line 55: What was the thinking of using `display: contents` here? I wasn't sure if it was adding anything

rmccar (commented):
Along with these comments Ive noticed the link doesn't have a focus state and isn't reachable via the keyboard

alessioventuriniAND (approved):

src/components/footer/_macro.njk
alessioventuriniAND: I think if the user uses the powered by parameter they can pass their html with their links. If we had a parameter that allowed you to pass in a Logo and we managed the layout and positioning of it like the header that we would also give the url parameter. But in this case it does look like the only icons that are made available in the footer are chosen by us so we might as well wrap them with the correct link. But maybe we can talk about making the footer's logo placements more aligned with the header component? @balibirchlee I think this is what we discussed right? It would be a separate ticket to improve our docs and code to fully manage the logos placements in the footer.

rmccar: Yeah fair enough, I just thought it would offer the user more flexibility but mainly I just wanted to make sure there wasn't a use case for it or not and it wouldn't be much work to parametrise it as it will enforce this change on every service.

balibirchlee: sorry for delay on this - illness etc.

So this is covered in the text under 'What is the context of this PR?' - I'm happy to create a ticket for it, so it doesn't get lost?

rmccar:
Have you pushed your latest changes? I cant see any extra commits since I last looked

SriHV (approved):

Issue #2159, opened 2022-05-03
Update legacy 'type' parameters to align with 'variants'


jrbarnes9:
Some components use parameters such as 'type' to set the different variants. This ticket is to update those so all components use the 'variants' parameter. 

Components include:
- Input (`params.type` - sets `type` attr. but “text” is used for number type)
- Panel (`params.type` & `params.spacious`)
- Status (`params.type` & `params.size`)
- Character check limit (`params.type`)

precious-onyenaucheya-ons:
Additional clarification and a technical discussion are needed for the implementation of this change in the Panel and Status components. Each of these components currently employs parameters with two different use cases.

yatesn:
Consider breaking down per component

daniellecorke:
To be started after the unnecessary conditionals work.

PR #3047, opened 2024-02-22
Update type param to variant on various components

rmccar:

Fixed: #2159 
Updates the panel, char check limit, status and message components to replace their type params with a variant param instead to bring them more in line with other components
Also fixes some formatting issues

Decision has been made to only make the changes for the `variant` params and not the `size` and `spacious` ones. It was felt that adding the size ones in would be confusing because some combinations of variants would work while others wouldn't. We will look to standardise these size attributes in another PR.

Also I have decided not to make the change to the input component as it makes sense for a `type` param set the `type` attribute and the param doesn't affect any visual difference to the input.

*BREAKING CHANGE*
Services using these components will now need to rename the `type` param to `variant`:
- Panel
- Message
- Status

### How to review this PR

- changes make sense
- vr tests pass
- check the panel, char check limit, status and message components still look and work as expected

### Checklist


alessioventuriniAND (approved):

src/components/char-check-limit/_macro-options.md
```diff
@@ -1,6 +1,6 @@
...
+| variant                    | string  | false                        | Set to “check” to add the character limit checker to the input                                                                                                                                     |
 | id                         | string  | true                         | The HTML `id` of the component                                                                                                                                                                     |
 | limit                      | integer | true                         | The maximum number of characters allowed in the input                                                                                                                                              |
 | charCountPlural            | string  | true (unless `type` is set)  | The string displayed when multiple characters can be entered before the limit is reached. Set `{x}` in the string to be replaced with the number, for example “You have {x} characters remaining”. |
```
adi-unni, line 6: Need to update this as type no longer exists along with the rest of the references to type?
```suggestion
| charCountPlural            | string  | true (unless `variant` is set)  | The string displayed when multiple characters can be entered before the limit is reached. Set `{x}` in the string to be replaced with the number, for example “You have {x} characters remaining”. |
```

adi-unni (changes requested):

src/components/char-check-limit/_macro-options.md
rmccar: I dont quite understand what you mean, "Type" has been updated to "variant" here?

rmccar: Good spot! Done!

adi-unni (approved):

PR #3053, opened 2024-02-26
Description list not structured properly for screen readers

### LINKED ISSUE #3027: List structure not properly represented to screen reader users (WCAG 2.2 Level A - Accessibility)
### Background
The eQ service identified this issue within the ONS Design System following their DAC audit report (18/01/2024). 

### Issue 
List structure not properly represented to screen reader users within the markup.

Image
The summary list of answers is marked up in a way which does not convey its structure properly to screen reader users. Although **this appears as a single list to sighted users**, it is represented **as three separate definition lists to blind users**, who may find it difficult to interpret the information as a result. This **issue occurs because each row within the summary list is marked up within its own `<dl>` element**. 

### Screen reader user comments
When viewing the details that could be changed in the ‘Calculated distance on foot and driving’ section I became aware that each item was in its own list rather than all being in one list, given they all related to the same content.
This made the information time consuming to navigate as I had to navigate past the list markup for each separate piece of content. Making all the content into one list will ensure a much more logical user experience.

### Suggested solution
Ensure that list structure is properly represented to screen reader users so that the information can be understood in the same way as by sighted users. In this case, **we recommend merging the three definition lists into a single list with three items**. 
This can be achieved by grouping each question within a `div` element appearing as the direct descendent of the `dl` element, to allow for further styling as per [Summary list – GOV.UK Design System](https://design-system.service.gov.uk/components/summary-list/).

### Tasks
- Investigate whether the change can be made and any impacts
- Decide the best solution
- Implement the solution
- Update eQ with progress

### Info
- [eQ DAC reports/docs](https://confluence.ons.gov.uk/display/SDC/eQ+DAC+Reports)
- Issue ID: DAC_List_Structure_01
- DAC report pages: 14 to 19
- WCAG Reference: 1.3.1 Info and Relationships (Level A): 
  - [Understanding Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) 
  - [How to Meet Info and Relationships](https://www.w3.org/WAI/WCAG22/quickref/#info-and-relationships)

precious-onyenaucheya-ons:

Fixes #3027 
The current structure of the summary component utilizes multiple `<dl> `elements, causing the screen reader to interpret them as separate list elements, which does not pass accessibility tests.

This PR addresses the issue by restructuring the Summary component so that the list is enclosed within a single <dl> element. This change ensures that the screen reader recognizes it as one list rather than multiple lists.

### How to review this PR

Check that the summary list is wrapped in a `<dl` element and is being read out correctly by the screen reader
### Checklist


src/components/summary/_macro.spec.js
```diff
@@ -307,8 +307,17 @@ describe('macro: summary', () => {
...
     it('passes jest-axe checks', async () => {
       const $ = cheerio.load(renderComponent('summary', EXAMPLE_SUMMARY_BASIC));
+      const results = await axe($.html(), {
+        rules: {
+          dlitem: {
+            enabled: false,
+          },
+          'definition-list': {
+            enabled: false,
+          },
```
rmccar, line 317: I think these rules can be added using a variable because they are used so many times. Would also probably be good to add a comment as to why we are adding them

alessioventuriniAND: Apart from this comment all looks good to me :)

alessioventuriniAND (approved):

src/components/summary/_macro.spec.js
precious-onyenaucheya-ons: done

src/components/summary/_macro.spec.js
```diff
@@ -303,20 +303,23 @@ const EXAMPLE_SUMMARY_MULTIPLE_GROUPS = {
   ],
 };
 
+// Disable specific axe rules to accommodate the testing scenario
```
rmccar, line 306: I dont think this is explains why we are adding these rules because if we come back to this in some time we wont know why we made the change. We want something like "disable some axe definition list rules - solving a DAC issue (https://github.com/ONSdigital/design-system/pull/3053) caused these tests to fail. We would need a big refactor to the way summaries and errors work to get these to pass but these failing is not seen as something that is a big problem accessibility wise so we have decided to remove the testing of them here".

precious-onyenaucheya-ons: updated

src/components/summary/_macro.spec.js
```diff
@@ -303,12 +303,26 @@ const EXAMPLE_SUMMARY_MULTIPLE_GROUPS = {
   ],
 };
 
+// To address a DAC issue, we've disabled specific axe definition list rules causing test failures.
+// While resolving it would require a significant refactor, the failures are deemed non-critical for accessibility,
+// leading to their removal in this context. [https://github.com/ONSdigital/design-system/issues/3027]
```
rmccar, line 308: Can you add something to outline and explain this change in the PR description too

rmccar (approved):
Other than this looks good

Issue #3025, opened 2024-02-01
Exploring Unnecessary Conditionals: Mutually-Exclusive


balibirchlee:
Component contains references to params.autosuggestResults which are not in the markdown file.

Actions
- [ ] Establish if this is legacy code or else is created elsewhere - double bracket to see if the param has a value.
- [ ] Remove if not necessary.
- [ ] Ensure testing remains intact / update as required.

PR #3076, opened 2024-03-07
Exploring Unnecessary Conditionals: Mutually-Exclusive

SriHV:
https://github.com/ONSdigital/design-system/issues/3025 .... Added new params that were missing in _macro_options.md

### How to review this PR
Check that params mentioned in examples are present in readme

### Checklist


rmccar (changes requested):
`autosuggestResults` and the fieldset params are missing

rmccar (changes requested):
So the way we have the params in other components are that the initial table is set up for params passed in with `param.<paramName>` then all the other tables like the fieldset one you have done here are for params passed in as an object to another one. Here I think they should all be in the same table. Check https://service-manual.ons.gov.uk/design-system/components/input#example.example-input-text.nunjucks-macro-options to see what I mean

src/components/mutually-exclusive/_macro-options.md
```diff
@@ -1,8 +1,18 @@
...
+| legendClasses                    | string                               | false                                        | Classes to apply to the legend element                                                                                                                                                                     |
+| legendIsQuestionTitle            | boolean                              | true (unless `legend` is set)                | Creates an `h1` inside the `legend`. Use when there is only a single fieldset on the page                                                                                                                  |
+| legendClasses                    | string                               | false                                        | Classes to apply to the legend element                                                                                                                                                                     |
```
rmccar, line 8: This one needs to be removed because you have it twice

SriHV: oh sry might have copied it by mistake

rmccar (approved):

Issue #3016, opened 2024-02-01
Exploring Unnecessary Conditionals: Details


balibirchlee:
Component contains references to params.accordian and params.group which are not in the markdown file.

Actions
- [ ] Establish if this is legacy code or else is created elsewhere - double bracket to see if the param has a value.
- [ ] Remove if not necessary.
- [ ] Ensure testing remains intact / update as required.Component contains a reference to 'params.borderlessParent' which is not in the markdown file.

PR #3088, opened 2024-03-12
Exploring Unnecessary Conditionals: Details

SriHV:

#3016....  Added two new params that were missing in `_macro_options.md`

### How to review this PR

Check that params declared in the component are present in `_macro_options.md`

### Checklist

### DIFF
--- a/src/components/details/_macro-options.md
+++ b/src/components/details/_macro-options.md
@@ -10,3 +10,5 @@
 | attributes        | object  | false    | HTML attributes (for example, data attributes) to add to the details                                                |
 | headingAttributes | object  | false    | HTML attributes (for example, data attributes) to add to the details header element                                 |
 | contentAttributes | object  | false    | HTML attributes (for example, data attributes) to add to the details content element                                |
+| isAccordion       | boolean | false    | Checks if the functionaltiy needs to be accordion. If true, adds the appropriate class                              |
+| group             | string  | false    | Used for accordian component where mutilple details components are grouped                                          |


src/components/details/_macro-options.md
```diff
@@ -10,3 +10,5 @@
 | attributes        | object  | false    | HTML attributes (for example, data attributes) to add to the details                                                |
 | headingAttributes | object  | false    | HTML attributes (for example, data attributes) to add to the details header element                                 |
 | contentAttributes | object  | false    | HTML attributes (for example, data attributes) to add to the details content element                                |
+| isAccordion       | boolean | false    | Checks if the functionaltiy needs to be accordion. If true, adds the appropriate class                              |
+| group             | string  | false    | Used for accordian component where mutilple details components are grouped                                          |
```
rmccar, line 14: I think these two are only used through the accordion component and so do not need to be listed here. Because you wouldn't ever need to use them when using the details component on its own

SriHV: okay got it, so only params that I need to list are the ones that are independent of other components. Can I close this PR and ticket if changes are not required?

rmccar: Yeah so the user only needs to know about params they would use and pass in themselves. These two are ones that are passed in from the accordion component but the user wouldn't pass them in to the details component. So yeah this can be closed.

SriHV:
Closed it as changes are not required

Issue #3017, opened 2024-02-01
Exploring Unnecessary Conditionals: Document List


balibirchlee:
Component contains conditionals around the .text .iso .fileType and .fileSize parameters - which are required parameters.

Actions
- [ ] Establish if this is superfluous or necessary code.
- [ ] Remove if not necessary.
- [ ] Ensure testing remains intact / update testing as required.

PR #3090, opened 2024-03-12
Exploring Unnecessary Conditionals: Document List

SriHV:

#3017 .... Updated params `required` field in `_macro_options.md`

### How to review this PR

Remove params in any of the example and if the page functions normally the param required status can be false.

### Checklist


rmccar (commented):
These aren't necessarily required as in it will break without them but the reason why these have been marked as required is more down to the fact the component doesn't look correct without them being set

Issue #3116, opened 2024-04-02
Fix repeatedly failing VR tests


adi-unni:
Some VR examples repeatedly fail the tests (such as where the component is sticky) so the decision has been made to omit them from the testing procedure.

adi-unni:
Not doing for now, will investigate other techniques to fix VR tests or use a different tool entirely in #3142

PR #3117, opened 2024-04-02
Add functionality to exclude examples from VR tests

adi-unni:

Fixes: #3116

Previously the sticky components would be failing the VR tests repeatedly. Looking at the backstop JS artifacts showed that the example was not being fully rendered and no fix was found to prevent this. The decision was made to stop certain examples from being included. I have added a change where you can edit the contents of the exclude list to remove certain examples from the VR tests.

Important note: If you are making a change to Table of Contents or a change to any global CSS. you must also check the ToC stick components separately. 

### How to review this PR

Check that the VR tests no longer run on example-table-of-contents-sticky and example-table-of-contents-sticky-full-page examples. Also sense check the code and please give suggestions if you have any enhancements.

### Checklist


balibirchlee (approved):

alessioventuriniAND (approved):
You might want to update the reference files and push to git lfs too.

alessioventuriniAND (approved):

rmccar (changes requested):
Isn't there a better way of fixing this? I don't like removing examples from tests. I can see you've added warning comments to the affected files but I can see this still getting forgotten. It will be very easy to forget to check this when changing global css. If we haven't already I think we need to look into why the examples aren't being rendered correctly, is it a race condition?

adi-unni:
> Isn't there a better way of fixing this? I don't like removing examples from tests. I can see you've added warning comments to the affected files but I can see this still getting forgotten. It will be very easy to forget to check this when changing global css. If we haven't already I think we need to look into why the examples aren't being rendered correctly, is it a race condition?

Alessio had looked into fixing this issue for a week in Feb and tried multiple different solutions with no result. I'll give it a try today, but if I can't, then this will be a temporary fix to prevent other PRs from being blocked while I work on a solution that separates CI tests from local tests so that all local tests run but omits tests in CI.

rmccar:
> > Isn't there a better way of fixing this? I don't like removing examples from tests. I can see you've added warning comments to the affected files but I can see this still getting forgotten. It will be very easy to forget to check this when changing global css. If we haven't already I think we need to look into why the examples aren't being rendered correctly, is it a race condition?
> 
> Alessio had looked into fixing this issue for a week in Feb and tried multiple different solutions with no result. I'll give it a try today, but if I can't, then this will be a temporary fix to prevent other PRs from being blocked while I work on a solution that separates CI tests from local tests so that all local tests run but omits tests in CI.

I still don't think removing test coverage is the answer, which includes removing tests in CI. There must be another way to fix this

Issue #3115, opened 2024-03-27
[Bug]: Incomplete Analytics Events Tracking


MebinAbraham:
### What happened?

Currently, some elements aren't being tracked accurately due to an issue in the analytics JS logic [here](https://github.com/ONSdigital/design-system/blob/main/src/js/analytics.js#L35-L39). The JavaScript checks whether the clicked target has a data-ga attribute of click. However, in cases where elements such as anchors contain nested elements like `<span>`, the target identified is the `<span>` itself, leading to the event not triggering as the attributes are set on the anchor.

```html
<a
  href="/questionnaire/non-mandatory-checkbox/?return_to=final-summary&amp;return_to_answer_id=non-mandatory-checkbox-answer#non-mandatory-checkbox-answer"
  class="ons-summary__button"
  data-qa="non-mandatory-checkbox-answer-edit"
  data-ga="click"
  data-ga-category="Summary"
  data-ga-action="Edit click"
>
  <span class="ons-summary__button-text" aria-hidden="true">
    Change
  </span>
  <span class="ons-u-vh">
    Change your answer for: What extra toppings would you like?
  </span>
</a>;
```

In the example above, the attributes are set on the anchor tag but the target the DS gets back when a click happens is `span` hence the tracking doesn’t work.

### What was the expected outcome?

All click events with the appropriate `data-ga` attributes should be tracked via GA.

### How is this problem impacting your service?

Missing vital analytics data.

### What list of steps can be followed to reproduce the bug or issue?

This specific issue can be seen when using the summary component where the rows have links i.e `Change` however, this is likely to be an issue on all elements that have nested elements where the attributes have been set on the parent element.

To recreate:
- Enable analytics
- Navigate to a page with elements containing nested elements like <span> within <a> tags. (Summary component)
- Click on row action element (e.g Change link) and observe the analytics event tracking aren't firing.

### Example

Can be tested on the summary component with action links.
![image](https://github.com/ONSdigital/design-system/assets/35296336/6692b1e6-492a-40ec-8f11-ee459cdf4709)

### What version of the ONS Design System are you using?

68.0.2

### What devices are you seeing the problem on?

Desktop, Mobile, Tablet

### Device details

N/A

### What operating systems are you seeing the problem on?

macOS

### Operating system details

N/A

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge, Microsoft Internet Explorer, Other

### Browser details

N/A

PR #3125, opened 2024-04-09
Introduced tracking for parent element if child is target of click event. 

alessioventuriniAND:

This fixes : #3115 

Some elements when clicked on return their child element as the target of the click event rather than the parent element which has all the data-ga attributes set. This fix will check for the parent of the element if the target element that has been clicked on doesn't have any data-ga attributes. 

### How to review this PR

It will be much easier to reach out to me to review this PR as I can show you how it works with GA analytics connected. 

### Checklist


adi-unni (approved):
It has been mentioned that if the data-ga tag is double nested then the fix would not function as intended but after reviewing the current state of the DS, there are no examples of which this would break case. If this edge case comes up in the future, a fix will be deployed then.

src/js/analytics.js
```diff
@@ -13,6 +13,7 @@ if (window.google_tag_manager !== undefined) {
 }
 
 export const trackElement = (el, type) => {
+  console.log(el);
```
precious-onyenaucheya-ons, line 16: remove this console.log please

SriHV (approved):

src/js/analytics.js
alessioventuriniAND: Done :)

precious-onyenaucheya-ons (approved):

Issue #2411, opened 2022-08-18
Selecting an input prefix/suffix should show the tooltip and put focus on the input


jrbarnes9:
Selecting an input prefix/suffix should show the tooltip and put focus on the input 

GDS have evidence that some users try to click their input prefix/suffix. ONS input prefixes/suffixes are styled similar to a secondary buttons which would further increase the affordance and the likelihood of this behaviour. Putting focus on the input when they are selected would replicate the standard label behaviour. Additionally showing the tooltip when they are selected (if is `title` applicable) would also help the discoverability of the full form label for mobile/sighted users.

![Screenshot 2022-08-18 at 15 04 28](https://user-images.githubusercontent.com/43346934/185414901-7506faaf-8d10-47d4-9e99-6559b285d1b0.png)

_Originally posted by @jrbarnes9 in https://github.com/ONSdigital/design-system/discussions/1383#discussioncomment-3423317_

PR #3130, opened 2024-04-09
show tooltip when hovering over input with prefix/suffix

### LINKED ISSUE #2946: Show tooltip when hovering over input with prefix/suffix
Showing the tooltip of an input when they are selected (if is title applicable) would also help the discoverability of the full form label for mobile/sighted users. Currently tooltip is only visible when the user hovers over the `<abbr>` tag using a mouse, however this functionality would not override, but compliment it by displaying the tooltip when the input is selected. Originally from ticket #2411

precious-onyenaucheya-ons:
fixes #2946 

This pull request introduces the functionality to display tooltips when hovering over inputs with both prefix and suffix. Additionally, corresponding tests have been added to ensure proper functionality.

### How to review this PR

Verify that inputs with either a prefix or suffix display the tooltip when hovered over.

### Checklist


alessioventuriniAND (approved):

adi-unni:
Scope of original ticket has changed. According to #2411 the ticket wanted the title text to be displayed when the input was in focus, not just when the input area is hovered over. As I understand, this change is not possible without reworking the component and is not backed up with user research for need so happy to approve in this current state

adi-unni (approved):

rmccar (changes requested):
Im not sure this is needed? Has this ticket been raised because something has been missed on the original issue? The original issue was asking for the tooltip to be added when hovering over the `<abbr>` this is used to give meaning to the abbreviation in the prefix/suffix and they usually have titles for this purpose. I don't think it was asking for the tooltip to also be added to the input.

Issue #3133, opened 2024-04-10
[Bug] Button in Radios component moves out of alignment when zoom is not 100%


daniellecorke:
### Background
This issue was reported via the Radios discussion board #56.

### Issue
[Radios component](https://service-manual.ons.gov.uk/design-system/components/radios)
When zoom is not 100%, the user reported that the button moves out of alignment. The user reported that this issue appears to mainly affect MacOS users, however I was able to replicate it without MacOS. 

### Example

![Image](https://github.com/ONSdigital/design-system/assets/102512949/4e3f7bee-9bed-462a-bb1d-fca7cc1f3826)

### Tasks
- Investigate the issue and identify a potential fix. 
- If possible, implement a fix for the issue. 
- PM to update the user of the outcome.

daniellecorke:
#2652 - previous ticket completed.

PR #3147, opened 2024-04-16
Fixed misalignment of circle inside radio input

alessioventuriniAND:

This fixes : #3133 

I have changed the way the circle inside the radio input is populated by removing the current after pseudo element and changing the background colour of the input itself to black on click or focus. This will fix the issue of misalignment. 
I have tried changing the positioning of the after element but it did get out of alignment at different zoom levels (90% and below and between 125% and 150%).

### How to review this PR

Open the netlify preview and check all the radios example. Confirm that the inside circle remain centered at every zoom level. 

### Checklist

### DIFF
--- a/src/components/radios/_radio.scss
+++ b/src/components/radios/_radio.scss
@@ -11,7 +11,6 @@
     &::after {
       display: none;
     }
-    &:focus,
     &:checked {
       background-color: var(--ons-color-input-border);
     }
--- a/src/components/radios/_radio.scss
+++ b/src/components/radios/_radio.scss
@@ -9,15 +9,11 @@
     box-shadow: inset 0 0 0 3px var(--ons-color-input-bg);
 
     &::after {
+      display: none;
+    }
+    &:focus,
+    &:checked {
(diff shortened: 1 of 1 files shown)


src/components/radios/_radio.scss
```diff
@@ -9,15 +9,11 @@
     box-shadow: inset 0 0 0 3px var(--ons-color-input-bg);
 
     &::after {
+      display: none;
+    }
+    &:focus,
```
rmccar, line 14: Once this line is removed for the focus issue we talked about then it looks good to me

alessioventuriniAND: removed :)

rmccar: Cool think you've managed to add a load of commits from other PRs though

alessioventuriniAND: very weird! I have just merged main into my branch

rmccar (approved):

balibirchlee (approved):

Issue #2138, opened 2022-04-13
Rename `pageURL` parameter of share-page component to `pageUrl`


kruncher:
### Tasks

1. Rename pageURL parameter of the [share page component](https://service-manual.ons.gov.uk/design-system/components/share-page#example.example-share-page.nunjucks-macro-options) to pageUrl

daniellecorke:
This ticket has been added to a larger ticket to cover small tasks that result in breaking changes.

rmccar:
This should be extended to cover other params that are named similarly with "URL" in the param name

PR #3166, opened 2024-04-30
Update param pageURL to pageUrl for share page component

adi-unni:

Fixes: #2138

The page url parameter across the DS was set as `pageUrl` but in share page component was `pageURL`. Updating param to `pageUrl` for consistency.

### How to review this PR

Check the tests pass and the new param works.

### Checklist


alessioventuriniAND (approved):

rmccar (commented):
I don't mind if you want to do it in another PR or this one but if we are making this change I think we should also do this for the other params that are named in this way. There is `logoURL` in the header component, `overviewURL` in naviagion, `pageURL` in share-page but there are others too. These should also be added to the issue.

balibirchlee (approved):

rmccar:
> I don't mind if you want to do it in another PR or this one but if we are making this change I think we should also do this for the other params that are named in this way. There is `logoURL` in the header component, `overviewURL` in naviagion, `pageURL` in share-page but there are others too. These should also be added to the issue.

Added the rest in #3188

adi-unni:
Change being completed in #3188

Issue #2822, opened 2023-09-28
Research options to refactor components tests


rmccar:
Our test descriptions could be better and more descriptive and meaningful. We should investigate the best strategies for this such as using Given-When-Then and then implement it to improve our automated tests.

- Research test frameworks like cucumber and get the pros and cons of a few options.
- Write a document on the the options with examples that we can then discuss in a tech session.

alessioventuriniAND:
We need to discuss this as a team and then create a ticket for each component so that the refactor can take place over the course of different sprints and completed by different developers.

balibirchlee:
Findings are in [this presentation doc](https://docs.google.com/presentation/d/1LwdM1ksluxNo28PXRETVX05nE5hN2KZ7KduCycAksCA/edit#slide=id.g2b863b078e7_0_125)

This will be picked up again once there is enough time for dev team to work on it & create actions points and new cards as needed.

balibirchlee:
New ticket has been made to create a proof of concept off the back of the research and suggested proposal - thus this ticket is now complete! 

[Follow-Up Ticket](https://github.com/ONSdigital/design-system/issues/3120)

PR #3168, opened 2024-04-30
Header Test Document Refactor

balibirchlee:

Follows the investigation and proposal done for ONSdigital/design-system#2822, making alterations as necessary to the proposed structure in response to our specific testing requirements.

Fixes: ONSdigital/design-team#59 

The conclusion of this ticket will spawn tickets for the following:
- Fully documenting the new test doc structure: clear approach and important info for refactoring other test docs/writing new tests.
- Refactoring all remaining components (each component will be a separate ticket).

### How to review this PR
Quite a few things to think about with this one!

**Functionality:**
- Run the test suite using: yarn test --testNamePattern="FOR: Header"

**Does this testing refactor meet the following success criteria:**
- Clearly structured.
- Follows naming conventions.
- Consistent.
- Concise.
- Suited to our codebase.
- Compatible with existing testing framework.

**Look at the refactored code in relation to:**
- The original test document.
- The macro file.
- The macro options file.

**Other Considerations (discussed in tech session 30/04):**
- Multiple set-ups - double check you're happy with the layout for this.
- Consistent testing - see if there's any tests that you think should be further split out or condensed into one!

### Checklist


src/components/header/_macro.spec.js
```diff
@@ -6,523 +6,348 @@ import axe from '../../tests/helpers/axe';
...
-                  logoImage: '<img src="a-logo.svg">',
-                },
-                logo2: {
-                  logoImage: '<img src="a-second-logo.svg">',
-                },
-                logo3: {
-                  logoImage: '<img src="a-third-logo.svg">',
-                },
+    });
+    describe('WHEN: multilpleLogos & mastheadLogo:Large parameters are both provided', () => {
```
rmccar, line 204:
```suggestion
    describe('WHEN: multipleLogos & mastheadLogo:Large parameters are both provided', () => {
```
Typo

src/components/header/_macro.spec.js
```diff
@@ -691,17 +456,37 @@ describe('macro: header', () => {
           ],
         },
       });
-
-      expect(buttonSpy.occurrences[0]).toHaveProperty('variants', ['mobile', 'text-link', 'text-link-inverse']);
+      test('THEN: renders button with coresponding variants', () => {
```
rmccar, line 459:
```suggestion
      test('THEN: renders button with corresponding variants', () => {
```
Typo

src/components/header/_macro.spec.js
```diff
@@ -828,155 +643,143 @@ describe('macro: header', () => {
...
-    it('gets the language if set', () => {
-      $('.ons-js-autosuggest-input').val('Eng');
+      const mockData = [
+        { en: 'England' },
+        { en: 'Wales' },
+        { en: 'Scotland' },
+        { en: 'United States of America' },
+        { en: 'United States Virgin Islands' },
+        { en: 'Åland Islands' },
+      ];
```
rmccar, line 759: Could set this a level up and then we wouldn't need to set it twice

src/components/header/_macro.spec.js
```diff
@@ -567,110 +392,50 @@ describe('macro: header', () => {
...
+          },
+        });
       });
     });
-
-    it('renders a button with correct variant if `internal`', () => {
+    describe('WHEN: one item is provided to the itemsList parameter', () => {
+      const $ = cheerio.load(renderComponent('header', EXAMPLE_HEADER_SERVICE_LINKS_SINGLE));
+      test('THEN: renders without multilink display class on desktop', () => {
+        expect($('.ons-header__links .ons-grid__col').hasClass('ons-u-d-no@xxs@')).toBe(false);
```
rmccar, line 429: The second breakpoint is missing off the `ons-u-d-no@xxs@` class

src/components/header/_macro.spec.js
```diff
@@ -6,523 +6,348 @@ import axe from '../../tests/helpers/axe';
...
+
+describe('FOR: Header', () => {
+  describe('GIVEN: Params: none', () => {
+    describe('WHEN: All params are at default state', () => {
+      const $ = cheerio.load(
+        renderComponent('header', {
+          ...EXAMPLE_HEADER_BASIC,
+        }),
+      );
+      test('THEN: Jest-Axe tests pass', async () => {
```
rmccar, line 27:
```suggestion
      test('THEN: jest-axe tests pass', async () => {
```
This library is lowercase

src/components/header/_macro.spec.js
```diff
@@ -6,523 +6,348 @@ import axe from '../../tests/helpers/axe';
...
+  });
+  describe('GIVEN: Params: mastheadLogo', () => {
+    describe('WHEN: default params are provided', () => {
       const faker = templateFaker();
       const iconsSpy = faker.spy('icon');
-
       faker.renderComponent('header', EXAMPLE_HEADER_BASIC);
 
-      expect(iconsSpy.occurrences[1].altText).toBe('Office for National Statistics logo');
+      test('THEN: renders default large logo on desktop', () => {
```
rmccar, line 91: I think we should specify "small screen" and "large screen" here rather than "mobile" and "desktop" just because there are more types of screens these days than mobile and desktop

src/components/header/_macro.spec.js
```diff
@@ -6,523 +6,348 @@ import axe from '../../tests/helpers/axe';
...
+  describe('GIVEN: Params: button: SignOutButton', () => {
+    describe('WHEN: no button parameters are provided', () => {
+      const $ = cheerio.load(
+        renderComponent('header', {
+          ...EXAMPLE_HEADER_BASIC,
+        }),
+      );
+      test('THEN: renders title grid with gutterless class', () => {
+        const titleGridDiv = $('.ons-header__main .ons-container .ons-grid');
+        expect($(titleGridDiv).hasClass('ons-grid--gutterless')).toBe(true);
```
rmccar, line 303: Think the wording on this test needs to be a bit more clear that there is no button expected to be rendered and maybe we need to add a test to show that

src/components/header/_macro.spec.js
```diff
@@ -567,110 +392,50 @@ describe('macro: header', () => {
...
+            'aria-expanded': 'false',
+          },
+        });
       });
     });
-
-    it('renders a button with correct variant if `internal`', () => {
+    describe('WHEN: one item is provided to the itemsList parameter', () => {
+      const $ = cheerio.load(renderComponent('header', EXAMPLE_HEADER_SERVICE_LINKS_SINGLE));
+      test('THEN: renders without multilink display class on desktop', () => {
```
rmccar, line 428:
```suggestion
      test('THEN: renders without multiple link display class on desktop', () => {
```
Think this wording is better

alessioventuriniAND (commented):
I would like to review this once you have addressed Richard comments. So far it looks great and works like a treat. Once the comments are addressed I'll go over it again before approving it.

src/components/header/_macro.spec.js
balibirchlee: This is actually the case in the original code as well - not entirely sure what the desired second break point is

src/components/header/_macro.spec.js
balibirchlee: we are technically not testing for this though - that would be a separate test I think! We are just checking that if we don't provide any parameters for a button that the rest of title grid displays properly

src/components/header/_macro.spec.js
balibirchlee: done!

src/components/header/_macro.spec.js
balibirchlee: done!

src/components/header/_macro.spec.js
balibirchlee: done!

src/components/header/_macro.spec.js
rmccar: Yeah true, fair enough

src/components/header/_macro.spec.js
rmccar: I think I see what its doing here looking at the tests above it, is its supposed to be checking for the existence of either the `ons-u-d-no@xxs@m` or `ons-u-d-no@xxs@xs` classes. I think we should probably just do that explicitly because its a bit confusing like this. Especially now the tests are a bit further apart. There is also the possibility that this isn't actually working and the `hasClass` function is looking specifically for `ons-u-d-no@xxs@` which it will never find but I haven't checked that in the docs

rmccar: Its not meant to find either. Thats the test. If one service link/language is in the masthead then it doesn't change to a menu button on small screens because it doesn't have either of those classes. Then if there are two or 3 or more links then it will change to a button at different break points. The `ons-u-d-no` utility class sets display to none for the button to appear instead. So what we need to do is to change this test to test that either class explicitly doesn't exist not `ons-u-d-no@xxs@`.

balibirchlee: I couldn't work out if it was supposed to be M or XS was the issue - but Alessio helped find it is supposed to be M.

rmccar: It needs to be both. We need to update the test to check that both `ons-u-d-no@xxs@m` and `ons-u-d-no@xxs@xs` do not exist in the class list

balibirchlee: btw this has been added!

alessioventuriniAND (approved):
This loos so much better than the current setup. I do think we need to get an epic created for all the standardisation work we are doing and this would be part of that.

rmccar (approved):
Looks good but when we do it for the rest we will need to update the documentation in the readme, the commands to run tests there will now be wrong

precious-onyenaucheya-ons (approved):

PR #3226, opened 2024-05-31
remove references to pattern lib

### LINKED ISSUE #3215: Remove references to the `patternlib`
There is still a number of references to the Pattern Library which is what the DS was before it was the DS, mainly around the `patternLibExampleGrid` macro and the patternLib-img folder. These can be changed so that they reference the DS instead.

precious-onyenaucheya-ons:

Fixes #3215

This PR focuses on renaming all references to the pattern lib within the project.

### How to Review this PR
Ensure that all renamed files continue to function as expected.
Confirm that all tests pass successfully.
Verify that all example files work correctly.

### Checklist

- [ ] I have selected the correct Assignee
- [ ] I have linked the correct Issue


alessioventuriniAND (approved):

SriHV (approved):

ci/generate-npm-package.js
```diff
@@ -27,7 +27,7 @@ async function copyAssets() {
     await fs.ensureDir(folder);
 
     try {
-      const files = (await fs.readdir(builtPath)).filter((path) => !path.includes('patternlib'));
+      const files = (await fs.readdir(builtPath)).filter((path) => !path.includes('ds'));
```
rmccar, line 30: The rest of the changes are fine, Im just wondering if "ds" here might be a bit too generic and could cause issues if we have a folder at some point called something that has the letters "ds" in it. For example if there was a folder called "cards" this would pick it up. So I was thinking it would be good to at least change this to "ds-" which will help with this. Will also need to be changed on line 40

rmccar (changes requested):

ci/generate-npm-package.js
precious-onyenaucheya-ons: updated

ci/generate-npm-package.js
```diff
@@ -27,7 +27,7 @@ async function copyAssets() {
         await fs.ensureDir(folder);
 
         try {
-            const files = (await fs.readdir(builtPath)).filter((path) => !path.includes('patternlib'));
+            const files = (await fs.readdir(builtPath)).filter((path) => !path.includes('ds-img'));
```
rmccar, line 30:
```suggestion
            const files = (await fs.readdir(builtPath)).filter((path) => !path.includes('ds-'));
```
This is what I was suggesting, it allows us to still have folders that start with this if we need to bundle other folders in there at some point which is what the original "patternlib" version did rather than specifically fully hardcoding one folder name

rmccar (approved):

PR #3262, opened 2024-07-10
update logic in related-content macro

### LINKED ISSUE #3236: Duplicate class on related links social example
Because of how the logic works in the related links and list components the example with the social links and icons(https://service-manual.ons.gov.uk/design-system/components/related-content/example-related-content-social-media) has the `ons-list--bare` class added twice.

The logic needs updating so it is only added once but this will have implications on other lists that don't use the social icons because this logic is also used by them so there may be some refactoring needed, this will have to be taken into considerations when fixing this bug.

download

precious-onyenaucheya-ons:

Fixes #3236 
The related links social example had the ons-list--bare class added twice. This PR updates the logic for the related content to ensure that the right class is added. Additionally, the documentation has been updated to include **variants** as a parameter.

### How to review this PR

Verify that the example in question no longer has the duplicate class added.
Ensure all tests pass.
Confirm that the documentation has been updated accordingly.

### Checklist

### DIFF
--- a/src/components/related-content/_macro-options.md
+++ b/src/components/related-content/_macro-options.md
@@ -14,3 +14,4 @@
 | iconType     | string                                                      | false    | Adds an icon to all the list items when set to the name of one of the [available icons](/foundations/icons#a-to-z). Requires `iconPosition` to be set. |
 | iconPosition | string                                                      | false    | Sets the position of the icon to “before” or “after” each list item in the row                                                                         |
 | iconSize     | string                                                      | false    | Icon size can be set to match the size of the list item text as detailed in the [typography type scale](/foundations/typography/#type-scale).          |
+| variants     | array or string                                             | false    | An array of values or single value (string) to adjust the list items using available variants: “bare”, “dashed”, “inline”, “social”, and “languages”   |
--- a/src/components/related-content/_macro.njk
+++ b/src/components/related-content/_macro.njk
@@ -16,7 +16,7 @@
                     <nav class="ons-related-content__navigation" aria-labelledby="{{ row.id }}">
                         {{
                             onsList({
-                                "variants": 'bare',
+                                "variants": row.variants if row.variants else 'bare',
                                 "iconType": row.iconType,
                                 "iconPosition": row.iconPosition,
                                 "iconSize": row.iconSize,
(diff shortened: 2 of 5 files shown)


src/components/related-content/_macro.njk
```diff
@@ -16,7 +16,7 @@
                     <nav class="ons-related-content__navigation" aria-labelledby="{{ row.id }}">
                         {{
                             onsList({
-                                "variants": 'bare',
+                                "variants": row.variants if row.variants else 'bare',
```
rmccar, line 19: If variants isn't set and this defaults to bare then the bare class will again be set twice. You can check this by removing the variants param from the related links example

rmccar (changes requested):

src/components/related-content/_macro.njk
precious-onyenaucheya-ons: I have addressed this

alessioventuriniAND (approved):

balibirchlee (approved):

SriHV (approved):

src/components/related-content/_macro-options.md
```diff
@@ -14,3 +14,4 @@
 | iconType     | string                                                      | false    | Adds an icon to all the list items when set to the name of one of the [available icons](/foundations/icons#a-to-z). Requires `iconPosition` to be set. |
 | iconPosition | string                                                      | false    | Sets the position of the icon to “before” or “after” each list item in the row                                                                         |
 | iconSize     | string                                                      | false    | Icon size can be set to match the size of the list item text as detailed in the [typography type scale](/foundations/typography/#type-scale).          |
+| variants     | array or string                                             | false    | An array of values or single value (string) to adjust the list items using available variants: “bare”, “dashed”, “inline”, “social”, and “languages”   |
```
rmccar, line 17: This can be removed

precious-onyenaucheya-ons: removed

rmccar (approved):

rmccar:
Because this has changed quite a bit and Im the only one that has reviewed it since the changes Ive requested some re-reviews

alessioventuriniAND (approved):
Can you please update the body of the PR to also include the change you have made around incorporating the social style in the native behaviour of the component and remove the social variant?

SriHV (approved):

Issue #3263, opened 2024-07-11
Focus states broken on internal header


rmccar:
Focus states aren't correctly displaying on internal headers, this would be an accessibility issue for users with impaired sight because there isn't enough contrast. The logo should be black with a yellow background when focused and the link should have black text with black underline when focused like other logos and links are styled.

To reproduce:
Navigate to the [internal header](https://service-manual.ons.gov.uk/design-system/components/header) example and focus on the ons logo or sign out link (using a full size screen)

<img width="393" alt="Screenshot 2024-07-11 at 14 26 43" src="https://github.com/ONSdigital/design-system/assets/42928680/63c9b012-5882-4c8e-9a4a-e4d0f310acb4">
<img width="478" alt="Screenshot 2024-07-11 at 14 26 34" src="https://github.com/ONSdigital/design-system/assets/42928680/783ea21b-a52e-45b3-afcb-f4b2cdf83814">

e-livingstone:
Discuss with @Majsol

PR #3269, opened 2024-07-16
Fix focus states on internal and neutral header

alessioventuriniAND:

This fixes : #3263 

The focus of the logo is now aligned with the rest of our headers variants and the service link focus state too. 

The focus hover state now removes any text-underline styling from the service links . This has also been applied to the neutral header. 

### How to review this PR

Move to the internal header example and tab through the elements and confirm that the focus state has the expected styling. 

### Checklist


balibirchlee (commented):
looks good to me and all works as expected

balibirchlee (approved):

SriHV (approved):

precious-onyenaucheya-ons (approved):

rmccar (changes requested):
Screenshot 2024-07-17 at 10 57 13

The focused hover isn't right here. I think it should either be black and not white or not have any change like the title link. Might need to check with Dina, if its the latter this will also need to be fixed in the neutral header too

rmccar (approved):

PR #3278, opened 2024-07-30
Move foundation examples to Components folder

precious-onyenaucheya-ons:

Fixes https://github.com/ONSdigital/design-team/issues/52
It relocates the examples of list, quote, text-indent, and address output from the foundations folder to the component folder. This change ensures that the macro options file and examples file are co-located for easier accessibility. Additionally, the grid examples have been moved and a new folder has been created for them.

### How to review this PR

Verify that all tests and visual tests have been updated correctly.

### Checklist


src/scss/main.scss
```diff
@@ -27,7 +27,7 @@
 @import '../components/fieldset/fieldset';
 @import '../components/footer/footer';
 @import '../components/header/header';
-@import '../components/helpers/grid';
+@import '../components/helpers/grid/grid';
```
rmccar, line 30: I think we should either remove the helpers folder or move the grid files into foundations. Grid isn't really a component

backstop_data/bitmaps_reference/ds-vr-test__components_helpers_grid_example-grid-8_0_document_0_desktop.png
rmccar, line 1: This grid examples are broken

rmccar (changes requested):

src/scss/main.scss
alessioventuriniAND: I agree I think one should go back to foundation as it's not a component. Reason I am saying this is not a component is that we don't expect users to use this grid macro but the grid classes.

alessioventuriniAND (commented):
Could you also run the reference command for the visual regression tests? It will remove all the old references that are not longer needed like all the screenshots of the same component but with the old name.

backstop_data/bitmaps_reference/ds-vr-test__components_helpers_grid_example-grid-8_0_document_0_desktop.png
precious-onyenaucheya-ons: i have fixed the issue

rmccar (approved):

SriHV (approved):

PR #3288, opened 2024-08-08
Update Fuse.js configuration to adjust search threshold

### LINKED ISSUE #3282: [3 days SPIKE] Increase number of characters in the search field
**Goal**
In our Community backlog, user raised a concern about the number of characters in the list which can be searched.
We need to investigate if it is possible to increase this number.

**Background**

Community backlog comment [link](https://github.com/ONSdigital/design-system/discussions/894#discussioncomment-10134657)

**Context**
Currently we only use 30 characters string. We will have to investigate if it is possible to increase it.

**Consideration**
We are using Fuse library, so we need to check if it is feasible to do the change.

**Output**
TBC

### LINKED ISSUE #3347: Extend search functionality for Autosuggest component
As IDS Developer
I want to increase the number of characters in Autosuggest component
So that search performed by my users includes all the characters which are after 20 (which is current limit)

**Description**
IDS Team highlighted to us a need for the different approach in Autosuggest component. The current limit of 20 characters doesn't allow IDS users to perform correct search in the organisation field. As per accessibility guidance, they adding the abbreviations of the organisations at the end of the sting. Based on their UR majority of the users will search by using the abbreviations, but current set up of the component do not allow it especially that the longest string of characters they need is 86. 
A new variant of the Autosuggest component would give users flexibility in setting up the parameter determining the length of characters as per their needs within the limit 0-1, where 0.2 = 20 characters and 0.6 = 60 characters.
Number of hits presented to the users will stay the same = 10.

**Acceptance Criteria**
1. Parameter limiting the number of characters during fuzzy search is editable for the users.
2. User can only set up value between 0-1.
3. Autosuggest search is adjusting the search based on the parameter set up. If parameter is 0.3 only 30 characters is available, but if user changes to 0.5 search will include 50 characters.
4. Documentation how to use this new functionality is added to Service Manual.

**Visualisation/Input from UCD**
N/a

**Linked User stories/epics**
Community backlog comment https://github.com/ONSdigital/design-system/discussions/894#discussioncomment-10134657

Technical details (optional)

MVP (optional)

precious-onyenaucheya-ons:

This PR fixes #3347 

This PR updates the Fuse.js configuration to address an issue where only the first 30 characters in the list seem to be searched.

The previous threshold setting of 0.2 was too restrictive, causing the search to miss relevant entries when queries became longer. With this update, users can now set the search threshold between 0 and 1, allowing for greater flexibility and improving search accuracy across different query lengths.

### How to review this PR

1. Run the search functionality with the updated configuration.
2. Perform a search for "South Georgia and the South Sandwich Islands" with the updated threshold.
3. Verify that the search result appears when searching for "South Sa" and similar queries.

### Checklist


rmccar:
The description now needs updating with the latest changes

rmccar (changes requested):

rmccar:
I think there may be an issue with this not working as expected, I upped the threshold to 0.8 in the example and was still getting the same results. So you may need to make sure that is being passed into the config correctly

rmccar (dismissed):

rmccar:
Im not sure this is defaulting to 0.2, when removing the setting of the param from the example the results are a lot stricter. Also the attribute is still being set just without a value which I dont think we should be doing

rmccar (dismissed):

admilne:
@precious-onyenaucheya-ons @rmccar 

Sorry, I ran out of time to look at this on Friday after running into some git permission issues Precious is aware of but I wanted to take a look before you have stand up in the morning. I'm not sure if I've misunderstood the ticket and the requirements - it wasn't entirely clear for me but this doesn't look like it's working to me. Could be I'm not looking at it correctly but once I hit around 32 characters in the search box it no longer returns the results I'd expect. I've taken some screen shots to illustrate what I experienced.

- "South Georgia and the South San" - returns what i'd expect
- "South Georgia and the South Sand" - doesn't seem to return anything
- "South Georgia and the South Sandw" - gives me some results, but doesn't seem to be based on anything I've searched for
- "South Georgia and the South Sandwich Island" - still appears to give me results not based on my search
- "South Georgia and the South Sandwich Islands" - suddenly seems to kick back in to life and return an expected result.

I tried playing with the thresholds but this didn't seem to make any difference to the issues I've described above.

Screenshot 2024-10-27 at 22 06 23
Screenshot 2024-10-27 at 22 06 34
Screenshot 2024-10-27 at 22 06 51
Screenshot 2024-10-27 at 22 07 09
Screenshot 2024-10-27 at 22 07 20

src/components/autosuggest/_macro-options.md
```diff
@@ -1,19 +1,20 @@
...
+| resultsTitleId      | string                               | true     | ID for the results title. The ID is used in the results `aria-labelledby` to provide context for the results                                                                 |
+| input               | `Input` [_(ref)_](/components/input) | true     | Configuration object for the input                                                                                                                                           |
+| language            | string                               | false    | The ISO 639-1 Code will override the default language in page. Please note that only 'en', 'cy' and 'ni' is currently supported                                              |
+| resultsThreshold    | integer                              | false    | Option to adjust the search threshold and fuzziness. Accepts a range from 0 to 1, where 0 provides the closest match and 1 allows for more distant matches. Defaults to 0.2. |
```
adi-unni, line 20: Just being very nit-picky but it isn't technically an integer. Maybe "float" would be more appropriate?

adi-unni:
I'd also like to second @admilne's findings. I've also done some additional testing with some of my own examples. 
I created a new JSON file with only examples of search terms which are longer than 32 characters. It seems like the search threshold doesn't change the results when typing. Whats more worrysome is the fact that words outside of the 32 character soft limit aren't picked up by fuse.

image

In this picture, you can see `this` is a search term. You can also see the word `characters` in the search terms which is the 54th character in the field

image

Characters does not return anything in this scenario

This might seem like an edge case but say there was a scenario where a user wanted to match this: 

_Ultra HD 4K Smart Television Model XYZ1234567890_

based on the model number which is outside of 32 characters

You can test my changes in: [feature/3282/increase-search-threshold-adi](https://github.com/ONSdigital/design-system/tree/feature/3282/increase-search-threshold-adi)

src/components/autosuggest/autosuggest.ui.js
```diff
@@ -300,7 +300,10 @@ export default class AutosuggestUI {
             this.customResultsThreshold != null && this.customResultsThreshold >= 0 && this.customResultsThreshold <= 1
                 ? this.customResultsThreshold
                 : 0.2;
-        const results = await runFuse(sanitisedQuery, data, this.lang, threshold);
+
+        const distance = threshold >= 0.6 ? 500 : threshold >= 0.4 ? 300 : 100;
```
rmccar, line 304: I don't think chaining ternaries together is a good idea, its not very readable. You are probably better off just using if else

rmccar:
> * South Georgia and the South Sand

After the changes that have been made with the distance param these issues seem to have fixed by upping the threshold to 0.4 in my testing. But the issues Andrew highlighted still seem to be an issue

rmccar:
> I'd also like to second @admilne's findings. I've also done some additional testing with some of my own examples. I created a new JSON file with only examples of search terms which are longer than 32 characters. It seems like the search threshold doesn't change the results when typing. Whats more worrysome is the fact that words outside of the 32 character soft limit aren't picked up by fuse.
> 
> image
> In this picture, you can see `this` is a search term. You can also see the word `characters` in the search terms which is the 54th character in the field
> 
> image
> Characters does not return anything in this scenario
> 
> This might seem like an edge case but say there was a scenario where a user wanted to match this:
> 
> _Ultra HD 4K Smart Television Model XYZ1234567890_
> 
> based on the model number which is outside of 32 characters
> 
> You can test my changes in: [feature/3282/increase-search-threshold-adi](https://github.com/ONSdigital/design-system/tree/feature/3282/increase-search-threshold-adi)
> 
> **Edit: Actually in even more testing, back to the original example (testing in countries-multiple), if you type "sandwich islands", you'd expect "South Georgia and the South Sandwich Islands" to appear but nothing appears**
> 
> Edit 2: Even weirder, if you search for "thi" in the search box in my example, you get matches for "thi" beyond 32 characters? Not sure if this is because its previously matched "thi" at the beginning.
> 
> image

These issues seem to have fixed by upping changing the distance to 500 in my testing without changing the threshold at all. So maybe we need to parameterise the distance not the threshold. But the issues Andrew highlighted still seem to be an issue. There might be an issue with the length of the query string entered which would cause the issues Andrew has found. This all needs bit more testing to make sure we get a balance between the distance and threshold. Testing with ignoreLocation might also be something we want to look into.

adi-unni (approved):

rmccar:
Ive found an issue with the component search on the DS. It finds the list of components when you start typing the name of a component but no longer goes to that page when you click the suggestion. There is this error in the console:
Screenshot 2024-10-31 at 09 48 00

src/components/autosuggest/autosuggest.ui.js
```diff
@@ -363,18 +368,21 @@ export default class AutosuggestUI {
...
                     listElement.setAttribute('role', 'option');
-                    if (result.item.category) {
+                    if (resultItem.category) {
                         innerHTML =
                             innerHTML +
-                            `<span class="ons-autosuggest__category ons-u-lighter ons-u-fs-s ons-u-db">${result.item.category}</span>`;
+                            `<span class="ons-autosuggest__category ons-u-lighter ons-u-fs-s ons-u-db">${resultItem.category}</span>`;
                     }
                     listElement.innerHTML = innerHTML;
+                    console.log(listElement);
```
rmccar, line 385: Think this has been left in by accident

SriHV (approved):

admilne (approved):
LGTM and looks like it's working as expected

Issue #3311, opened 2024-08-22
[Feature]: Add ability to set the colour of check icon in summary lists


rmccar:
### What feature would you like to add to the ONS Design System?

Add ability to set the colour of check icon in summary lists

### Why should this new feature be added to the Design System?

This will allow the check icon to appear as green as what is currently happening in eQ when using the summary component for the same purpose

rmccar:
A suggested solution to this would be to add the ability to pass icon classes from the list component to the icon component and then add some css in the list component which would change the colour to green if a class like `ons-icon--check-green` was added

PR #3318, opened 2024-08-23
Updated color of check icon for summary variant in list component to green

SriHV:

#3311...

As per the ticket, the defualt color of check icon for the summary variant in list component is changed to green. 

### How to review this PR

Add the check icon in example-summary-list and check that the icon's color is in green

### Checklist

### DIFF
--- a/src/components/list/_list.scss
+++ b/src/components/list/_list.scss
@@ -63,9 +63,6 @@
     &__prefix,
     &__suffix {
         font-family: $font-mono;
-        .ons-icon--check-green {
-            fill: var(--ons-color-leaf-green);
-        }
     }
 
     &--prefix,
@@ -109,6 +106,12 @@
             @extend .ons-u-pt-s;
             @extend .ons-u-pb-s;
             @extend .ons-u-m-no;
+        }
+        .ons-list__prefix,
+        .ons-list__suffix {
+            &--icon-check .ons-icon {
+                fill: var(--ons-color-leaf-green) !important;
+            }
         }
     }
 }
--- a/src/components/list/_list.scss
+++ b/src/components/list/_list.scss
@@ -63,8 +63,7 @@
     &__prefix,
     &__suffix {
         font-family: $font-mono;
-        .ons-icon {
-            &--check-green {
+        .ons-icon--check-green {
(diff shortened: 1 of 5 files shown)


src/components/list/_macro-options.md
```diff
@@ -8,6 +8,7 @@
 | iconPosition | string            | false    | Sets position of icon to “before” or “after” each list item                                                                                            |
 | iconType     | string            | false    | Adds an icon to all the list items when set to the name of one of the [available icons](/foundations/icons#a-to-z). Requires `iconPosition` to be set. |
 | iconSize     | string            | false    | Icon size can be set to match the size of the list item text as detailed in the [typography type scale](/foundations/typography/#type-scale).          |
+| iconClasses  | string            | false    | Adds on option to pass icon classes fron list component.                                                                                               |
```
rmccar, line 11: Maybe a description of "Classes to add to the icon when using a list with icons" would fit in with other descriptions a bit more

src/components/list/_list.scss
```diff
@@ -63,6 +63,11 @@
     &__prefix,
     &__suffix {
         font-family: $font-mono;
+        .ons-icon {
+            &--check-green {
```
rmccar, line 67:
```suggestion
        .ons-icon--check-green {
```

This may as well be this because its not adding anything by splitting them out

rmccar:
Actually, Ive had another thought that might provide a better way of configuring it for the user. We add some logic into the macro so that if the user sets the icon to "check-green" then we add the class to the icon automatically rather than the user having to set it themselves, that will then also remove the need for the new param

alessioventuriniAND (commented):
I am not sure about this approac. IconType is supposed to receive the string name of the icon type that we have available in the icon set list in the service manual. Check-green is not one of the listed icon types. 

Otherwise we can simply add the check-green variant to the icon list without making any changes to the list component.

alessioventuriniAND (approved):

rmccar (approved):

rmccar:
Snapshot needs updating

rmccar:
Also the PR title and description need updating because the solution has changed

PR #131, opened 2019-02-05
111 add resize

boxadesign:

PR #3320, opened 2024-08-28
Refactor Accordion component test file to new format

balibirchlee:

Fixes [#131 ](https://github.com/ONSdigital/design-team/issues/131) in the design team issues board

I've also added some tests that I thought were missing

### How to review this PR

Ask yourself....

- Do we need any additional tests?
- Are the tests descriptions clear?
- Does the order of the test document make sense?
- Is it internally consistent?
- Is it consistent with the [guidelines for test refactoring](https://confluence.ons.gov.uk/display/SDC/Testing+Refactor+Guidance)?
- run this test command to view the test results:
- yarn test --testNamePattern="FOR: Macro: Accordion"

### Checklist


src/components/accordion/_macro.spec.js
```diff
@@ -4,64 +4,65 @@ import * as cheerio from 'cheerio';
...
-};
-
-const EXAMPLE_ACCORDION = {
-    ...EXAMPLE_ACCORDION_WITH_TWO_ITEMS,
-    allButton: {
-        open: 'Open label',
-        close: 'Close label',
-    },
-};
-
```
alessioventuriniAND, line 29: The only thing I am noticing is that we are not longer testing the accordion component with the allButton parameter(so the one buttone that opens all accordions) against axe. Was this intentional?

src/components/accordion/_macro.spec.js
```diff
@@ -4,64 +4,65 @@ import * as cheerio from 'cheerio';
...
-describe('macro: accordion', () => {
-    it('passes jest-axe checks', async () => {
-        const $ = cheerio.load(renderComponent('accordion', EXAMPLE_ACCORDION));
-
-        const results = await axe($.html());
-        expect(results).toHaveNoViolations();
+import { EXAMPLE_ACCORDION } from './_test_examples';
+
+describe('FOR: Accordion', () => {
+    describe('GIVEN: Params: required ', () => {
```
rmccar, line 10:
```suggestion
    describe('GIVEN: Params: required', () => {
```

src/components/accordion/_macro.spec.js
```diff
@@ -173,9 +180,56 @@ describe('macro: accordion', () => {
...
+                    ...EXAMPLE_ACCORDION,
+                    saveState: true,
+                }),
+            );
+            test('THEN: renders with saveState attribute', () => {
+                expect($('.ons-details--accordion').attr('saveState'));
+            });
+        });
+    });
+    describe('GIVEN: Params: open ', () => {
```
rmccar, line 212:
```suggestion
    describe('GIVEN: Params: open', () => {
```

src/components/accordion/_macro.spec.js
rmccar: In main `EXAMPLE_ACCORDION` is only used for the jest-axe test. The rest of the time the `allButton` param is added again without using `EXAMPLE_ACCORDION`. This PR still tests those but also adds them so its all covered and fixes the near redundant example set out here with the `allButton` param. I think the only think that is missing is the fact we now don't use the `allButton` version for the jest-axe test. We should add this to the test.

rmccar (changes requested):

src/components/accordion/_test_examples.js
```diff
@@ -0,0 +1,15 @@
+export { EXAMPLE_ACCORDION };
+
+const EXAMPLE_ACCORDION = {
```
rmccar, line 3:
```suggestion
export const EXAMPLE_ACCORDION = {
```

balibirchlee: this is neater for one case but if things are added in the future then wouldn't we want the original set up? that's how it's set up in other files i believe

balibirchlee: we can move to this - but it won't be consistent with some of the others already merged, i preferred it like this initially because it is the same format as the import statement - making it easier to ensure you are exporting and importing the exact same thing

alessioventuriniAND (approved):

src/components/accordion/_macro.spec.js
```diff
@@ -4,64 +4,82 @@ import * as cheerio from 'cheerio';
...
-
-describe('macro: accordion', () => {
-    it('passes jest-axe checks', async () => {
-        const $ = cheerio.load(renderComponent('accordion', EXAMPLE_ACCORDION));
-
-        const results = await axe($.html());
-        expect(results).toHaveNoViolations();
+import { EXAMPLE_ACCORDION } from './_test_examples';
+
+describe('FOR: Accordion', () => {
```
rmccar, line 9:
```suggestion
describe('FOR: Macro: Accordion', () => {
```
Think we should make this change to specify between macro and script tests. Making this change and the same for scripts will allow us to still specify if we want to run all script tests, all macro tests, all tests or the tests for a single component

rmccar (approved):

PR #3382, opened 2024-10-07
Rename items summary param for compatibility with Jinja

rmccar:

`items` is a reserved word in python so this was causing an error when we used the param in a Python environment. Renaming is `itemsList` which is what we have used in other components solves this.

### How to review this PR

Test that the summary component in both nunjucks and jinja environments, make sure it works as expected

### Checklist


precious-onyenaucheya-ons (commented):
You missed to update in the hub and spoke example pattern

precious-onyenaucheya-ons (approved):

alessioventuriniAND (approved):

PR #3398, opened 2024-10-10
Fix font size for section navigation header

### LINKED ISSUE #3399: Horizontal section navigation header has a larger font size than other sections
A bug has been introduced in the section navigation component where when using the horizontal variant, the active heading is being rendered in a larger font size than the links for the other sections. This heading needs to be brought back in line with the links for the other sections

Screenshot 2024-10-10 at 10 31 46

precious-onyenaucheya-ons:

Fixes #3399 
This PR updates the font sizes of the section navigation item header to 18px.

### How to review this PR

Check that the section nigivation header font is updated

Check that the visual test is updated  
### Checklist


src/components/section-navigation/_section-navigation.scss
```diff
@@ -45,6 +45,10 @@
 
     &__item-header {
         @extend .ons-u-mb-no;
+
+        font-size: 1.125rem;
+        font-weight: 700;
+        line-height: 1.6;
```
rmccar, line 51: I think a better solution here would be to extend the `ons-u-fs-r--b` class instead.

The font-weight on line 29 and the margin on line 71 can also be removed as they aren't having any effect in this instance

src/components/section-navigation/_section-navigation.scss
```diff
@@ -45,6 +44,8 @@
 
     &__item-header {
         @extend .ons-u-mb-no;
+
```
rmccar, line 47:
```suggestion
```
Would be a bit neater without this line but thats a minimal thing. Looks good other than that

rmccar (approved):

PR #3408, opened 2024-10-28
Refactor summary component to fix lighthouse test

rmccar:

Fixes: https://github.com/ONSdigital/design-team/issues/219

Lighthouse tests on summary component are failing because there is more than one level of `<div>` inside the `<dl>` when only one level is valid before the `<dt>` and `<dd>` elements. This PR refactors the summary component so that there is only one `<div>` level between the `<dl>` and `<dt>` and `<dd>` elements. As part of this summary row titles needed to be changed from being marked up as a `<div>` to a `<dt>`.

The setting and testing of the `ons-summary__row--has-values` class was also removed as this class no longer seems to be used for anything.

### How to review this PR

- Check the lighthouse tests in the github actions and see that the result for the test for the hub and spoke examples is now at least 100.
- Also run the test locally on this branch in chrome and on the main branch to see that this issue is now fixed

### Checklist


src/components/summary/_macro.njk
```diff
@@ -35,96 +35,91 @@
...
+                                        <dt class="ons-summary__row-title ons-summary__row-title--no-group-title ons-u-fs-r">
                                             {{- row.title -}}
-                                        </div>
+                                        </dt>
                                     {% endif %}
 
                                     {% for item in row.itemsList %}
-                                        <div
-                                            class="ons-summary__row{{ ' ons-summary__row--has-values' if item.valueList else "" }}"
-                                            {% if item.id %}id="{{ item.id }}"{% endif %}
```
precious-onyenaucheya-ons, line 49: Since the row id is removed, we can remove it from the macro-options file

precious-onyenaucheya-ons (approved):

Issue #3296, opened 2024-08-15
[TASK] Rename DS table of contents files


rmccar:
**Description**
The table of contents component files in the DS are named "toc" which is inconsistent with the rest of the components where the files are named after the folder. 

**To do:** 
Rename Table component files from "toc" to  "table-of-contents".

rmccar:
One class name could not be updated because it would cause a potential breaking change for services. We have a new issue #3430  to make this change when we do our next major release because this isn't an urgent change.

PR #3424, opened 2024-11-15
Rename table of contents files

rmccar:

Fixes: #3296 

Renames the table of contents files to match the naming conventions of other similar files in the DS.

Also made some standardisation changes on the back to top component to bring them inline with js on other components

### How to review this PR

- Tests pass
- Table of contents and back to top still render and work as expected with no changes to examples

### Checklist


admilne (approved):

src/components/table-of-contents/_table-of-contents.scss
precious-onyenaucheya-ons, line 1: we should modify the class name to .ons-table-of-contents as well

rmccar: Ive made these changes as you are right but this does introduce a breaking change as we have documentation that recommends adding the `ons-js-toc-container` class on a div around the variant here https://service-manual.ons.gov.uk/design-system/components/table-of-contents#example.example-table-of-contents-sticky.nunjucks. So I am going to revert this change to the classes and add another card to make that change nearer the time when we want to make our next major release.

precious-onyenaucheya-ons (approved):

Issue #3393, opened 2024-10-08
Investigate and adjust the line height on details tabs


rmccar:
**As** a user,
 **I want** the line height on details tabs to be consistent with font being used
**so that** it aligns properly with the grid and provides a clean and organised layout.

**Description**
Currently it is 2.25rem. Line high should not be specified in the Component, but driven by the font, if possible. 

**Acceptance Criteria:**

1. Investigate the current line height settings on details tabs.
2. Identify any discrepancies or misalignments with the grid.
3. Adjust the line height settings to ensure that details tabs sit correctly on the grid.
4. Verify that the changes are consistent across different screen sizes and devices.
5. Ensure that the visual appearance of the details tabs is improved and aligns with the overall design guidelines.

**Visualisation**

<img width="733" alt="Screenshot 2024-10-08 at 16 32 06" src="https://github.com/user-attachments/assets/d0474b3c-7f62-4c70-94b4-68e09267fac0">

**Link to the issue on Figma** https://www.figma.com/design/Wr9b6pK35euFoA6xq3G9Ld/Grid-review?node-id=0-1&node-type=canvas&m=dev

PR #3432, opened 2024-11-20
Remove Line-height in tab component

SriHV:

#3393 ...

As per the ticket, replaced line-height with padding and box-shadow in `.ons-tab--row` class
Also adjusted width of the box size of the element to 40px from 40.8px as per Dina's suggestion.
Screenshot 2024-11-20 at 16 38 34

### How to review this PR
Inspect the examples in tab component and check that line-height and is driven by padding instead

### Checklist


src/components/tabs/_tabs.scss
```diff
@@ -43,11 +43,10 @@
     border-radius: 3px 3px 0 0;
     color: var(--ons-color-text);
     display: inline-block;
-    height: 2.55rem;
-    line-height: 2.25rem;
+    height: 2.5rem;
     margin: 0 0.5rem 0 0;
     overflow: visible;
-    padding: 0 1rem;
+    padding: 0.3125rem 1rem 0.375rem;
```
rmccar, line 49: I think this can be set to this and this causes less changes to the tabs component. It also means that we can stay away from specifically sized padding like this and can align the padding here with our padding size scale
```suggestion
    padding: 0.25rem 1rem 0.25rem;
```

admilne (approved):
LGTM

rmccar (approved):

PR #3439, opened 2024-11-27
Add inline description list variant

precious-onyenaucheya-ons:

[ONSDESYS-189](https://jira.ons.gov.uk/browse/ONSDESYS-189)....

Created new Description list variant as per the [Figma](https://www.figma.com/design/Yf4Cpxc8vYdoVrWcJojk2R/New-Website-Platform?node-id=255-18366&node-type=text&m=dev).

Prototype [link](https://nwp-prototype.ons.gov.uk/)

### How to review this PR
Check that new example-inline-description-list in Description List component matches the figma design

### Checklist


rmccar:
Looks like its possible for the colon to wrap onto another line

Screenshot 2024-11-28 at 10 46 19

src/components/description-list/_description-list.scss
```diff
@@ -1,6 +1,14 @@
 .ons-description-list {
     &__items {
         margin: 0 0 2rem;
+
+        &.ons-description-list--horizontal {
+            @media (width >= 980px) {
```
rmccar, line 6: I dont think we should be using custom breakpoints here we should use the large breakpoint instead which is also set to 980px

You would do something like this for example:
```
@include mq(l) {
    margin-bottom: 1rem;
    margin-top: 1rem;
}
```

src/components/description-list/_macro-options.md
```diff
@@ -6,6 +6,7 @@
 | termCol              | number        | true     | The number of grid columns used for the `<dt>` elements above medium breakpoint                                |
 | descriptionCol       | number        | true     | The number of grid columns used for the `<dd>` elements above medium breakpoint                                |
 | itemsList            | array`<Item>` | true     | Settings for the terms and descriptions of the [description list items](#item)                                 |
+| horizontal           | boolean       | false    | Set to "true" to align the description list horizontally                                                       |
```
rmccar, line 9: To match other components we should have a `variant` param instead and if that param is set to "inline" then it renders it in the horizontal format. This would match what we have with the list component

rmccar:
I think we need to check on how this behaves with smaller screens with Joe, the figma still has it behave like this
Screenshot 2024-11-28 at 11 52 27
but the document list that we have at the moment behaves like this
Screenshot 2024-11-28 at 11 53 09
We may need to adjust the breakpoints

admilne:
I think the use of dt and dd tags are the wrong approach. At the moment this is forcing the text to wrap in a way different to the prototype. Look at the 'Business' column in this example.

image

This is the example in the prototype. You can see from the code that they are using a span tag for the 'Business' column and then just test for information about the release - so the text will wrap underneath 'Business'

Screenshot 2024-11-28 123351

src/components/description-list/_macro.njk
```diff
@@ -1,23 +1,34 @@
 {% macro onsDescriptionList(params) %}
     <dl
-        class="ons-description-list ons-description-list__items ons-grid ons-grid--gutterless ons-u-cf{{ " " + params.classes if params.classes else "" }}"
-        {% if params.id %}id="{{ params.id }}"{% endif %}{% if params.descriptionListLabel %}
+        class="ons-description-list ons-description-list__items{{ " " + params.classes if params.classes else "" }}{{ " ons-description-list--horizontal" if params.horizontal else " ons-grid ons-grid--gutterless ons-u-cf" }}"
```
rmccar, line 3: I think the only class that needs to be removed here when you are using horizontal is `ons-u-cf`

src/components/description-list/_macro.spec.js
rmccar, line 1: I think we should leave refactoring the tests out of this. Keeping this separate would help with reviewing. Save what you have done on another branch though so we don't lose the work you have done

src/components/description-list/_macro.njk
```diff
@@ -1,23 +1,34 @@
...
+                            class="ons-description-list__value ons-grid__col ons-col-{{ params.descriptionCol }}@m"
+                        >
+                            {{- descriptionItem.description -}}
+                        </dd>
+                    {% endif %}
+                {% endfor %}
+            {% endset %}
+
+            {% if params.horizontal %}
+                <div>{{ descriptionItems | safe }}</div>
```
rmccar, line 28: The adding of this div here is breaking the css and no longer adding the margin top to the items so they are less spaced out when they are rendered vertically compared to when the div isn't added

src/components/description-list/_description-list.scss
```diff
@@ -1,6 +1,14 @@
 .ons-description-list {
     &__items {
         margin: 0 0 2rem;
+
+        &.ons-description-list--horizontal {
```
rmccar, line 5: This class is set in the html at the same level as the items class. So it should probably be structured where they are are the same level and not inside eachother

src/components/description-list/_macro-options.md
```diff
@@ -1,11 +1,12 @@
...
+| id                   | string          | false    | The HTML `id` for the description list (`<dl>`)                                                                |
+| classes              | string          | false    | Classes to apply to description list (`<dl>`)                                                                  |
+| descriptionListLabel | string          | false    | Sets the HTML `title` and `aria-label` attributes to provide additional information about the description list |
+| termCol              | number          | true     | The number of grid columns used for the `<dt>` elements above medium breakpoint                                |
+| descriptionCol       | number          | true     | The number of grid columns used for the `<dd>` elements above medium breakpoint                                |
+| itemsList            | array`<Item>`   | true     | Settings for the terms and descriptions of the [description list items](#item)                                 |
+| variants             | array or string | false    | An array of values or single value (string) to adjust the component using available variants:“inline”          |
```
rmccar, line 9: We only have one variant at the moment so this should just be called "variant" and should only accept a string

src/components/description-list/_macro.njk
```diff
@@ -1,23 +1,36 @@
 {% macro onsDescriptionList(params) %}
     <dl
-        class="ons-description-list ons-description-list__items ons-grid ons-grid--gutterless ons-u-cf{{ " " + params.classes if params.classes else "" }}"
-        {% if params.id %}id="{{ params.id }}"{% endif %}{% if params.descriptionListLabel %}
+        class="ons-description-list ons-description-list__items ons-grid ons-grid--gutterless{{ " " + params.classes if params.classes else "" }}{{ " ons-description-list--inline" if params.variants == 'inline' else " ons-u-cf" }}"
```
rmccar, line 3: We should be using single quotes here where we are using them inside double quotes

src/components/description-list/_description-list.scss
```diff
@@ -7,26 +7,39 @@
         clear: both;
         float: left;
         font-weight: $font-weight-bold;
-
-        &:not(:first-child) {
-            margin-top: 0.5rem;
-        }
+        word-break: break-word;
```
rmccar, line 10: We can remove this, its deprecated and allows the breaking of words which is what the default does anyway

src/components/description-list/_description-list.scss
```diff
@@ -7,26 +7,39 @@
...
     }
 
     &__value {
         float: right;
         margin-left: 0; /* As normalize adds a 40px left margin */
+        word-break: break-word;
```
rmccar, line 16: Same as my other comment

src/components/description-list/_macro.spec.js
```diff
@@ -117,22 +127,22 @@ describe('macro: description-list', () => {
         const $listElements = $('.ons-description-list__term, .ons-description-list__value');
 
         expect($listElements[0].tagName).toBe('dt');
-        expect($($listElements[0]).text()).toBe('Survey:');
+        expect($($listElements[0]).text().trim()).toBe('Survey:');
```
rmccar, line 130: Were these no longer passing without adding trim?

src/components/description-list/_macro.njk
```diff
@@ -1,23 +1,31 @@
...
-                    <dd
-                        {% if descriptionItem.id %}id="{{ descriptionItem.id }}"{% endif %}class="ons-description-list__value ons-grid__col ons-col-{{ params.descriptionCol }}@m"
+            <div class="ons-description-list__item">
+                {% if item.term | length %}
+                    <dt
+                        class="ons-description-list__term ons-grid__col ons-col-{{ params.termCol }}@{{ 'xs@l' if params.variant == 'inline' else 'm' }}"
                     >
-                        {{- descriptionItem.description -}}
-                    </dd>
+                        {{ item.term }}
```
rmccar, line 15:
```suggestion
                        {{- item.term -}}
```
I think the reason you have had to use trim is because you removed these here

src/components/description-list/_description-list.scss
```diff
@@ -7,25 +7,65 @@
         clear: both;
         float: left;
         font-weight: $font-weight-bold;
-
-        &:not(:first-child) {
-            margin-top: 0.5rem;
-        }
+        word-wrap: break-word;
```
rmccar, line 10: I don't think we need this
```suggestion
```

src/components/description-list/_description-list.scss
```diff
@@ -7,25 +7,65 @@
...
     }
 
     &__value {
         float: right;
         margin-left: 0; /* As normalize adds a 40px left margin */
+        word-wrap: break-word;
```
rmccar, line 16:
```suggestion
```
I don't think we need this

rmccar:
This is the prototype with a screen width of about 950px

Screenshot 2024-12-05 at 15 18 12

This is your branch with a screen width of about 950px

Screenshot 2024-12-05 at 15 19 42

We need to check this is correct because it isn't quite the same at the prototype

rmccar:
This change to columns is now causing this issue

Screenshot 2024-12-06 at 09 23 45

precious-onyenaucheya-ons:
> This change to columns is now causing this issue
> 
> Screenshot 2024-12-06 at 09 23 45

I have added padding to this

rmccar:
I think you need to revert the col change otherwise we are getting this

Screenshot 2024-12-06 at 14 51 08

src/components/description-list/_description-list.scss
```diff
@@ -7,25 +7,62 @@
...
+                    margin-top: 0;
+                }
+                @include mq(xs, l) {
+                    margin-top: 0.5rem;
+                }
+            }
+            .ons-description-list--inline & .ons-description-list__term {
+                @include mq(l) {
+                    margin-top: 0;
+                }
```
rmccar, line 45: This could be reorganised to avoid the repetition of `.ons-description-list--inline` and `@include mq(l)` here

rmccar: Theres still a lot of repetition and redundant code here, all this could be replaced by just this:
```
.ons-description-list--inline & {
    .ons-description-list__value:nth-of-type(1), .ons-description-list__term {
        @include mq(l) {
            margin-top: 0;
        }
    }
}
```

src/components/description-list/_description-list.scss
```diff
@@ -7,25 +7,62 @@
...
 
     &__value {
         float: right;
         margin-left: 0; /* As normalize adds a 40px left margin */
 
-        &:not(:nth-of-type(1)) {
-            @include mq(m) {
+        & + & {
+            @include mq(xs) {
+                margin-top: 0;
```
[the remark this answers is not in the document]
precious-onyenaucheya-ons: Yeah let me have a look

src/components/description-list/_description-list.scss
```diff
@@ -7,25 +7,61 @@
...
+                .ons-description-list__term {
+                    @include mq(l) {
+                        margin-top: 0;
+                    }
+                }
+            }
+        }
+    }
+
+    &--inline {
```
rmccar, line 48:
```suggestion
    &--inline & {
```
This can be changed like this and then you can replace all the instances of ".ons-description-list" below with "&"

src/components/description-list/_description-list.scss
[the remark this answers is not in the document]
precious-onyenaucheya-ons: updated

rmccar (approved):

src/components/description-list/_description-list.scss
```diff
@@ -7,25 +7,56 @@
...
+
+            .ons-description-list--inline & {
+                .ons-description-list__value:nth-of-type(1),
+                .ons-description-list__term {
+                    @include mq(l) {
+                        margin-top: 0;
+                    }
+                    @include mq(xs, l) {
+                        margin-top: 0.5rem;
+                    }
```
admilne, line 37: I don't think this is doing anything and could perhaps be removed?

admilne (approved):

PR #3463, opened 2024-12-20
Remove panel title for warn, announcement and bare panels

rmccar:

Fixes: [ONSDESYS-226](https://jira.ons.gov.uk/browse/ONSDESYS-226)

Removes support for titles with warn, announcement and bare panels. When titles were used with these variants it was breaking the styling

Also updates the params documentation which was out of date with some of the variants

### How to review this PR

Check that all panels still display as expected when the `title` param is provided including the warn, announcement and bare panels

### Checklist


SriHV (approved):

src/components/panel/_macro-options.md
```diff
@@ -1,13 +1,13 @@
...
+| ------------------- | ------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| body                | string  | true     | The contents of the panel. This can contain HTML.                                                                                                                                                                                                                                                       |
+| title               | string  | false    | The title for the error summary panel (not supported for the warn, warn-branded, announcement and bare panels )                                                                                                                                                                                         |
```
admilne, line 4: I don't think it works correctly for 'ghost' either as it ends up with white text on a white background.

Thinking about making it future proof, I wonder if it might be better to switch it round and say 'only supported by x, xyz variants' rather than list the ones it doesn't support?

src/components/panel/_macro.njk
```diff
@@ -50,7 +50,7 @@
                 <span class="ons-panel__assistive-text ons-u-vh">{{ params.assistiveTextPrefix | default("Announcement: ") }}</span>
             {% endif %}
 
-            {% if params.title %}
+            {% if params.title and (params.variant not in ['bare', 'warn', 'warn-branded', 'announcement'] ) %}
```
admilne, line 53: Related to my comment above, I think it might be better to list the variants the title is permitted for rather than list the ones it isn't. Since the title is white it feels safer to deliberately allow the title than to accidentally forget to disallow it.

src/components/panel/_macro-options.md
rmccar: There isn't a ghost variant of the panel component. But I agree with your point on switching it round, I think that makes it a bit simpler

admilne: This is the 'ghost' variant I'm referring to 
image

Yeah, I get what you are saying about not switching them around. I was thinking more in terms of if/when another variant gets created it will be easy to forget to add the new variant to the list of disallowed variants which use the title. It seems more likely that if you want the title you would deliberately add it to the list of allowed variants rather than risk forgetting to add it to the disallowed variants. It's a minor thing which may never come up so I'm not too worried about it but thought I'd raise it just incase you felt it was worth chaning.

rmccar: Yeah I agree with your logic, and if we didn't have the issue that I explained I would switch them round. As for the ghost variant, there is no variant called ghost so this looks like an issue with the documentation that I have now fixed in this PR

admilne (approved):

PR #3465, opened 2024-12-24
Moved description into dl tag in message list component

SriHV:

[204](https://jira.ons.gov.uk/browse/ONSDESYS-204)

Moved description into dl tag in message list component.

To achieve this new param `bodyLabel` was added and  `ariaLabelMsg` param was deprecated.

As per the deprecation implementation process available in confluence
 `(DEPRECATED)` is added to the `ariaLabelMsg` param and `NEW` is added to the `bodyLabel` param in message list documentation.

### How to review this PR
Inspect example-message-list.html in send and reply message patterns and check that the description is inside the dl tag.

### Checklist


rmccar (changes requested):

admilne (approved):

src/components/message-list/_macro.njk
```diff
@@ -15,10 +15,16 @@
                     <dd class="ons-message-item__metadata-value ons-message-item__metadata-value--date ons-u-fs-s">
                         {{ message.dateText }}
                     </dd>
+                    <dt
+                        class="ons-message-item__metadata-term ons-message-item__metadata-term--body ons-u-vh"
+                        aria-label="{{ params.ariaLabelMsg | default("Message text") }}"
```
rmccar, line 20: We shouldn't need this line anymore

src/components/message-list/_macro.spec.js
```diff
@@ -74,18 +74,6 @@ describe('macro: message-list', () => {
...
-        const $ = cheerio.load(renderComponent('message-list', EXAMPLE_MESSAGE_LIST_MINIMAL));
-
-        expect($('.ons-message-item__body:first').attr('aria-label')).toBe('Message text');
-    });
-
-    it('has `aria-label` attribute on `.ons-message-item__body` using the provided value', () => {
-        const $ = cheerio.load(renderComponent('message-list', EXAMPLE_MESSAGE_LIST));
-
-        expect($('.ons-message-item__body:first').attr('aria-label')).toBe('Message preview');
-    });
```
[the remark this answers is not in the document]
rmccar: I think these actually need to just be repurposed to test that the new message params do what we expect because I don't think we need the aria attribute anymore

rmccar:
As this is a deprecation we need to follow the process outlined in the "Deprecation Implementation and Release Process" in confluence we have set out. You have added the "(DEPRECATED)" tag but we need to follow the other parts of the process

src/components/message-list/_macro.spec.js
```diff
@@ -39,7 +41,6 @@ const EXAMPLE_MESSAGE_LIST = {
     ...EXAMPLE_MESSAGE_LIST_MINIMAL,
     ariaLabel: 'Message list for ONS Business Surveys',
     ariaLabelMetaData: 'Message information',
-    ariaLabelMsg: 'Message preview',
```
rmccar, line 42: As we still support this param we still need to test it. We will need to add an example in where this is set instead of `bodyLabel` and a test similar to the new test you have added

src/components/message-list/_macro.spec.js
```diff
@@ -20,6 +20,7 @@ const EXAMPLE_MESSAGE_LIST_MINIMAL = {
             },
             fromText: 'Example Sender 1',
             dateText: 'Tue 4 Jul 2020 at 7:47',
+            bodyLabel: 'Body',
```
rmccar, line 23: This param shouldn't be at the message level, it should be at the top param level like the other label params

rmccar (changes requested):

src/components/message-list/_macro.njk
```diff
@@ -15,10 +15,16 @@
                     <dd class="ons-message-item__metadata-value ons-message-item__metadata-value--date ons-u-fs-s">
                         {{ message.dateText }}
                     </dd>
+                    <dt class="ons-message-item__metadata-term ons-message-item__metadata-term--body ons-u-vh">
+                        {% if params.ariaLabelMsg %}
+                            {{ params.ariaLabelMsg | default("Message text") }}:
+                        {% endif %}
+                        {{ params.bodyLabel | default("Message text") }}:
```
rmccar, line 22: The way this is written allows both to be set and then both to be passed into the HTML, but we want one or the other. Currently when I test it with both params both pieces of text are rendered. We want `bodyLabel` to be the only one used if both are set. Then if one of the two are set that one is used. There is also currently a bit of an odd bug where if only `ariaLabelMsg` is set what it is set to is rendered but also "Message text:" is added on the end

Screenshot 2025-01-14 at 09 09 20

Tests need to be written for all these situations to make sure we have coverage and the test that is there will need to be checked because this should make the test fail

src/components/message-list/_macro.spec.js
```diff
@@ -39,7 +40,11 @@ const EXAMPLE_MESSAGE_LIST = {
     ...EXAMPLE_MESSAGE_LIST_MINIMAL,
     ariaLabel: 'Message list for ONS Business Surveys',
     ariaLabelMetaData: 'Message information',
-    ariaLabelMsg: 'Message preview',
+};
+
+const EXAMPLE_MESSAGE_LIST_WITH_DEPRECATED_ariaLabelMsg_PARAM = {
+    ...EXAMPLE_MESSAGE_LIST_MINIMAL,
+    ariaLabelMsg: 'Message Preview',
 };
```
rmccar, line 48: This doesnt need to be its own example config as its only being used in one test at the moment you can just add the param in that test. Also the name needs updating because it should be all in caps

rmccar (approved):

PR #3502, opened 2025-02-17
Adjusted alignment of button within new grid

SriHV:

[ONSDESYS-137](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-137)

As per Dina's suggestion. increased height of the button to sit in the 8px grid
Removed CSS set for IE11 High Contrast mode which adjusted the button size.

I changed padding to increase the height. Small button height is changed from 32 to 36 and large button's height from 52 to 56.

### How to review this PR
Inspect any of the button examples and check that button's height is rounder
### Checklist


src/components/button/_button.scss
```diff
@@ -42,7 +34,7 @@ $button-shadow-size: 3px;
         color: var(--ons-color-text-inverse);
         display: flex;
         align-items: center;
-        padding: 0.75rem 1rem;
+        padding: 0.875rem 1rem;
```
rmccar, line 37: Why was this change made? If I revert it, the button is still on the grid

rmccar:
Can you update the ticket link and mention the change in height of the buttons in the description

PR #3506, opened 2025-02-24
Added Line chart variant of the Chart component

precious-onyenaucheya-ons:
### What is the context of this PR?

Addressed [ONSDESYS-297](https://jira.ons.gov.uk/browse/ONSDESYS-297)

This PR introduces the Line Chart variant of the Chart component, built based on the Torchbox POC and technical specifications. An example implementation of the chart is also included.

### How to review this PR

1. Verify that the Line Chart component is implemented correctly and adheres to the Torchbox POC and technical specifications.
3. Check that the component is fully responsive across different screen sizes.
5. Confirm that the component has passed all required testing (backstop JS tests, snapshot tests, visual regression tests, and manual testing in the Python environment).
6. Validate that the component meets accessibility criteria.

### Checklist

This needs to be completed by the person raising the PR.

-   [x] I have selected the correct Assignee
-   [x] I have linked the correct Issue

PR #3507, opened 2025-02-24
Split out global chart options from ones specific to the colour theme or chart type

helenb:

Demo for @precious-onyenaucheya-ons to show how we might split the chart options into ones that are global, for all chart types, and ones that need to be updated for each chart.

### How to review this PR

Have a look at the code changes and test them to see if they make sense.
Either merge them to your PR if you would like to, or copy and adapt if you would prefer.

### Checklist

### DIFF
--- a/src/components/charts/common-chart-options.js
+++ b/src/components/charts/common-chart-options.js
@@ -2,10 +2,6 @@
 class CommonChartOptions {
     constructor() {
         this.constants = {
-            primaryTheme: ['#206095', '#27a0cc', '#003c57', '#118c7b', '#a8bd3a', '#871a5b', '#f66068', '#746cb1', '#22d0b6'],
-            // Alternate theme colours from https://service-manual.ons.gov.uk/data-visualisation/colours/using-colours-in-charts
-            alternateTheme: ['#206095', '#27A0CC', '#871A5B', '#A8BD3A', '#F66068'],
-            labelColor: '#414042',
             axisLabelColor: '#707071',
             gridLineColor: '#d9d9d9',
             zeroLineColor: '#b3b3b3',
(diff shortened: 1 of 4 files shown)


src/components/charts/common-chart-options.js
```diff
@@ -1,6 +1,6 @@
 /* this contains global options for all charts */
-class ChartOptions {
-    constructor(theme, title, type) {
+class CommonChartOptions {
```
helenb, line 2: I've renamed the class and the file to reflect the fact that this is now common chart options only, and to differentiate it from the `SpecificChartOptions` class.

src/components/charts/common-chart-options.js
```diff
@@ -1,6 +1,6 @@
...
+    constructor() {
```
helenb, line 3: We no longer need to pass in the type of chart or the theme, as any options that need to check these have moved to `SpecificChartOptions`

src/components/charts/common-chart-options.js
```diff
@@ -16,27 +16,13 @@ class ChartOptions {
...
         this.options = {
-            colors: theme === 'primary' ? this.constants.primaryTheme : this.constants.alternateTheme,
             chart: {
                 backgroundColor: 'transparent',
                 style: {
                     fontFamily: '"OpenSans", "Helvetica Neue", arial, sans-serif',
                     color: '#222222',
                 },
             },
-            legend: {
```
helenb, line 27: I've moved *all* the legend configuration to `SpecificChartOptions`, even though some of it is general, in order to avoid the merging of the options getting too complicated - it's easier to merge in all the top level values for each key in the options object.

helenb:
Closing this MR as @precious-onyenaucheya-ons has now incorporated the changes in https://github.com/ONSdigital/design-system/pull/3506

Issue #141, opened 2019-02-06
Password


jrbarnes9:
Use this issue to discuss the [password component](https://ons-design-system.netlify.com/components/password) in the [ONS Design System](https://ons-design-system.netlify.com)

SteveScorfield:
Hello, I have noticed that there is an issue with the “password/_macro.njk”. When calling "onsPassword", I receive the following error:

  File "/app/app/templates/signin.html", line 29, in top-level template code
    {% set password_input = onsPassword({
  File "/usr/local/lib/python3.6/site-packages/jinja2/runtime.py", line 579, in _invoke
    rv = self._func(*arguments)
  File "/app/app/templates/components/password/_macro.njk", line 7, in template
    {% call onsField({
  File "/usr/local/lib/python3.6/site-packages/jinja2/runtime.py", line 579, in _invoke
    rv = self._func(*arguments)
  File "/app/app/templates/components/field/_macro.njk", line 17, in template
    {{ caller() }}
  File "/usr/local/lib/python3.6/site-packages/jinja2/runtime.py", line 579, in _invoke
    rv = self._func(*arguments)
  File "/app/app/templates/components/password/_macro.njk", line 32, in template
    onsInput({
  File "/usr/local/lib/python3.6/site-packages/jinja2/runtime.py", line 579, in _invoke
    rv = self._func(*arguments)
  File "/app/app/templates/components/input/_macro.njk", line 44, in template
    {{ onsLabel({
  File "/usr/local/lib/python3.6/site-packages/jinja2/environment.py", line 430, in getattr
    return getattr(obj, attribute)
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'label'

We have managed to fix the issue locally by including the “label” attribute within “onsInput”:

```
            onsInput({
                "id": params.id,
                "name": params.name,
                "type": "password",
                "classes": "u-mt-xs js-password-input",
                "dontWrap": true
                "label": {
                    "text": params.showPasswordText
                }
            })
```

 This may need adding to future releases.

glancyons:
If this is used against a panel with background colour, the checkbox is filled with the background colour - I'd expect it to be white:
![image](https://user-images.githubusercontent.com/60258968/75863129-e647ec80-5df7-11ea-9f64-4f1b16b0c5f9.png)
not like this
![image](https://user-images.githubusercontent.com/60258968/75863169-f790f900-5df7-11ea-9330-fc5229dc9832.png)

PR #3509, opened 2025-02-28
Refactor Card component test file to new format

adi-unni:

Fixes: [#141](https://github.com/ONSdigital/design-team/issues/141)

Refactors the card component test file to the new agreed format.

### How to review this PR

Run tests locally and check that the messages make sense and describe what is being tested

Ask yourself....

- Do we need any additional tests?
- Are the tests descriptions clear?
- Does the order of the test document make sense?
- Is it internally consistent?
- Is it consistent with the [guidelines for test refactoring (https://confluence.ons.gov.uk/display/DS/Testing+Refactor+Guidance)?
- run this test command to view the test results:
- yarn test --testNamePattern="FOR: Macro: Card"


src/components/card/_macro.spec.js
```diff
@@ -1,268 +1,246 @@
...
-                },
+            test('THEN: the provided `text` is accessible via the `textId` identifier', () => {
+                expect($('#example-text-id').text().trim()).toBe('Example card text.');
             });
+        });
 
-            expect(listsSpy.occurrences[0]).toEqual({
-                variants: 'dashed',
-                itemsList: [{ text: 'Test item 1' }, { text: 'Test item 2' }],
+        describe('WHEN: heading levels are tested', () => {
```
rmccar, line 37:
```suggestion
        describe('WHEN: headingLevel is provided', () => {
```

src/components/card/_macro.spec.js
```diff
@@ -1,268 +1,246 @@
...
+            it.each([
+                [1, 'h1'],
+                [4, 'h4'],
+            ])('THEN: it has the correct element headingLevel tag %i outputs <%s> tags', (headingLevel, expectedTag) => {
```
rmccar, line 41:
```suggestion
            ])('THEN: it renders with the headings with the correct heading tag (%i -> <%s>)', (headingLevel, expectedTag) => {
```

rmccar (changes requested):
Theres a few pointers here. Ive gone through half of the changes and created suggestions. I think there needs to be some more refactoring here to get it in the right format. Have a look at some of the other tests for structure and wording for these tests. Access code is a good example

janehughes (approved):
Looks good, working as expected 👍  I was wondering if it's worth adding a test to check for duplicate IDs if we have multiple cards on the same page?

rmccar:
> Looks good, working as expected 👍 I was wondering if it's worth adding a test to check for duplicate IDs if we have multiple cards on the same page?

Its a good point but we don't have the functionality build into the component to deal with that at the moment. Also I feel this would be above the component level to check and if we were to look into that we would need to implement it on many components. Maybe something to explore at some point but at the moment we trust the service to not duplicate ids

admilne (approved):

PR #3521, opened 2025-03-12
Fix summary rendering when no valueList is supplied

rmccar:

Fixes: #3541
[ONSDESYS-394](https://jira.ons.gov.uk/browse/ONSDESYS-394)

Have updated the summary macro to add a class that sets the width of the actions to be double width to take up the space that the valueList would take up. This fixes a misalignment when valueList is not provided. Also added tests for this and a missing test for the equivalent in actions.

Have also updated the example so that these scenarios are covered so if this ever happens again it should be picked up and have updated the documentation because either valueList or actions should be set to make the html correct from a semantic standpoint.

### How to review this PR

- Remove the setting of the valueList param from the `example-summary-grouped` file on main
- See that on main it misaligns all the items in the summary
- Compare this with this branch using the same config
- See that the issue is fixed

### Checklist


admilne (commented):
@rmccar this looks good to me. The only thing I would question is if there are no actions then this layout looks odd to me. I think I would expect the values to stay in the column and not span two columns but I'm not sure if there have been discussions about this and how it should look?
image

rmccar:
> @rmccar this looks good to me. The only thing I would question is if there are no actions then this layout looks odd to me. I think I would expect the values to stay in the column and not span two columns but I'm not sure if there have been discussions about this and how it should look? image

Yeah good point, Ive added padding in this situation to resolve this and match what would be happening if there was a action link on that row

rmccar:
> > > @rmccar this looks good to me. The only thing I would question is if there are no actions then this layout looks odd to me. I think I would expect the values to stay in the column and not span two columns but I'm not sure if there have been discussions about this and how it should look? image
> > 
> > 
> > Yeah good point, Ive added padding in this situation to resolve this and match what would be happening if there was a action link on that row
> 
> I'm not sure this solution works because if you have a lot of text in a row that does include actions it looks like this: image

Ive changed the padding size to be 33% of the width. I think this should fix this issue. This then should span 33% if there are actions in other rows and if not should span 66%

admilne (approved):

precious-onyenaucheya-ons (approved):

PR #3529, opened 2025-03-28
Add information panel to hero

rmccar:

Fixes: [ONSDESYS-371](https://jira.ons.gov.uk/browse/ONSDESYS-371)

Adds the ability to add an information panel and link to the grey variant of the Hero. The panel supports 3 different colours.
[Figma](https://www.figma.com/design/Yf4Cpxc8vYdoVrWcJojk2R/New-Website-Platform?node-id=0-1&p=f&m=dev)

### How to review this PR

- Check the information panel works as expected by making changes to the example using the newly introduced params
- Check the panel behaves as expected when the screen size is reduced

### Checklist


src/components/hero/_hero.scss
```diff
@@ -98,6 +108,22 @@
         height: 100%;
     }
 
+    &__information {
+        display: flex;
+        flex-wrap: wrap;
+        column-gap: 1rem;
+
+        & .ons-hero__panel {
+            padding: 0.5rem;
```
janehughes, line 117: Tiny nitpick: based on the design I think this should be (gives a bit more breathing room on mobile especially)

```suggestion
            padding: 1rem 0.5rem;
```

janehughes (approved):
Just one small comment re: padding otherwise this LGTM!

src/components/hero/_hero.scss
rmccar: I think what I already has matches the figma. Its 8px top and bottom. Although I can see there is a red panel image in the ticket that has more space. Ill double check this

Screenshot 2025-04-01 at 12 30 47

rmccar: Oh actually did you mean to suggest `padding: 0.5rem 1rem;`?

SriHV (approved):

src/components/hero/_hero.scss
janehughes: Actually no I think I was looking at the larger red example as well - but you're right, this should be `1rem 0.5rem`.  Looking good now, thank you :)

PR #3526, opened 2025-03-26
Add point annotations

helenb:
### What is the context of this PR?

Ticket reference: https://jira.ons.gov.uk/browse/CCB-26
Adds point annotations to charts, with footnotes at mobile

Builds on https://github.com/ONSdigital/design-system/pull/3512

### How to review this PR

Visit http://localhost:3000/components/chart/example-line-chart-with-annotations.html and compare the styles at desktop and mobile

- Check that the styling matches the ACs
- See datavis guidlines for annotations, although note that we have agreed not to use the arrow styling: https://ons-design.notion.site/Chart-furniture-dad53df69db049aa9e34cb38aed546ce?pvs=25#a8ed2e596aa74bd099b73fcbba9b52bd
- See mockup of a mobile annotations style on the spike ticket for this work: https://jira.ons.gov.uk/browse/CCB-61

### Checklist

This needs to be completed by the person raising the PR.

-   [x] I have selected the correct Assignee
-   [x] I have linked the correct Issue

### Screenshots
<details>
<summary>Click to view screenshots</summary>

Mobile:

Screenshot 2025-03-26 at 10 58 58

Desktop:

Screenshot 2025-03-26 at 10 59 06

</details>

PR #3532, opened 2025-03-28
Add point annotations

helenb:

Squashed version of https://github.com/ONSdigital/design-system/pull/3526.

Ticket reference: https://jira.ons.gov.uk/browse/CCB-26
Adds point annotations to charts, with footnotes at mobile

### How to review this PR

Visit http://localhost:3000/components/chart/example-line-chart-with-annotations.html and compare the styles at desktop and mobile

Check that the styling matches the ACs
See datavis guidlines for annotations, although note that we have agreed not to use the arrow styling: https://ons-design.notion.site/Chart-furniture-dad53df69db049aa9e34cb38aed546ce?pvs=25#a8ed2e596aa74bd099b73fcbba9b52bd
See mockup of a mobile annotations style on the spike ticket for this work: https://jira.ons.gov.uk/browse/CCB-61

### Checklist


src/components/chart/annotations-options.js
```diff
@@ -0,0 +1,76 @@
...
+        let annotations = [
+            {
+                labels: [],
+                labelOptions: {
+                    shape: 'connector',
+                    borderColor: this.constants.labelColor,
+                    padding: 3,
+                    style: {
+                        color: this.constants.labelColor,
+                        fontSize: this.constants.desktopFontSize,
```
janehughes, line 19: This is currently displaying as 14px for me at desktop - based on the ACs it should be 16px

janehughes (commented):
I've just added one small comment re: desktop font size for labels, otherwise this LGTM!

src/components/chart/_chart.scss
```diff
@@ -21,6 +22,55 @@
...
+
+    &__footnotes {
+        list-style: none;
+        padding: 0;
+        margin: 0 0 1rem;
+
+        // We need to use a container query here in order to match the responsive rules for Highcharts
+        // which is based on the width that the chart renders not the viewport
+        // Update this is the Highcharts responsive rules change
+        @container (min-width: 601px) {
```
precious-onyenaucheya-ons, line 62:
```suggestion
        @media (min-width: 601px) {
```

src/components/chart/annotations-options.js
helenb: Thanks @janehughes - this has made me realise we have smaller fonts than the ACs specify throughout the charts, not just here. But changing them to larger looks quite large, and increases the chance of overlapping text. I'll raise this as a question on slack, but hopefully shouldn't block this PR, as we can do it as a separate fix for everything if needed.

src/components/chart/_chart.scss
helenb: Hi @precious-onyenaucheya-ons we definitely need a [container query](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries) here. The reason is because Highcharts responsive rules are not responsive to the viewport, but to the width of the chart. Therefore in hiding and showing the annotations footnotes, we also need to respond to the width of the chart - otherwise there will be a screen size where either we get the numbers in the chart, but no footnotes below, or we get the text in the chart as well as below.

precious-onyenaucheya-ons: I'm happy to resolve this since the minimum width is applied to the chart itself rather than the viewport.

src/components/chart/annotations-options.js
helenb: I've chatted with Erin from datavis about this - the AC is my mistake. We do need a ticket to adjust the mobile font-sizes, but I'll cover that separately. I'm going to do a sweep of tickets and update the ACs regarding font size this afternoon.

src/components/chart/_chart.scss
```diff
@@ -21,6 +22,55 @@
...
+
+    &__footnote-number {
+        display: flex;
+        flex-shrink: 0;
+        flex-grow: 0;
+        align-items: center;
+        justify-content: center;
+        min-width: 1.125rem; // 18px
+        min-height: 1.125rem; // 18px
+        padding: 3px;
```
admilne, line 34: I wonder if this should be a relative size so it can remain responsive

src/components/chart/_chart.scss
```diff
@@ -21,6 +22,55 @@
...
+    }
+
+    &__footnotes {
+        list-style: none;
+        padding: 0;
+        margin: 0 0 1rem;
+
+        // We need to use a container query here in order to match the responsive rules for Highcharts
+        // which is based on the width that the chart renders not the viewport
+        // Update this is the Highcharts responsive rules change
```
admilne, line 61: I don't think we need this last line of the comment

src/components/chart/_chart.scss
helenb: I think this is fine as px in this case - the line will always be 3px away from the text, and the text itself is responsive. The screenshot shows what it looks like if we zoom:

Screenshot 2025-04-02 at 11 37 11

And with the font size setting changed in the browser:

Screenshot 2025-04-02 at 11 38 00

src/components/chart/_chart.scss
helenb: Ah thanks - it was badly worded with a typo. I've improved it now - it's to remind us to update this breakpoint if it changes in the Highcharts config.

src/components/chart/_chart.scss
admilne: I was looking at it from the point of view of if the font size ever changed.
image

helenb: I have amended it to rems (although note that if the font size changed, we'd still need to adjust the padding).

admilne: Ah yes, I see. If we change it to em instead of rem though I think this can get round this issue? An attractive looking `0.2143em` looks like it may do the job if we are sticking to 3px

admilne (approved):

PR #3533, opened 2025-03-28
Set Domain Cookie to Sub-Domains

SriHV:

[ONSDESYS-341](https://jira.ons.gov.uk/browse/ONSDESYS-341)

As per the ticket. I've refactored getDomain() to first attempt setting a cookie at the full subdomain level, rather than immediately simplifying the domain.
This function falls back to higher-level domains only if setting a cookie fails at the full subdomain.
Also added an option to override document.cookie so that we can use this to mock the cookie in the tests.

Added Tests which cover:
Successful cookie setting on a full subdomain.
Fallback when cookie cannot be set on a subdomain.
Removal of www. prefix from domain names.

### How to review this PR
Inspect and open the console tab in  the service-manual website. Copy the `getDomain` function with the new changes and call this function  (`getDomain(document.domain)`). Verify that domain name is `service-manual.gov.uk`. 
Compare this with existing code which returns `ons.gov.uk`

### Checklist


src/js/cookies-functions.js
```diff
@@ -74,8 +74,8 @@ export function getConsentCookie() {
 }
 
 export function setConsentCookie(options) {
-    const domain = getDomain(document.domain);
-
+    const domain = getDomain(window.location.hostname);
+    console.log(domain);
```
precious-onyenaucheya-ons, line 78: Do we need this log here?

SriHV: Removed it 👍

janehughes (approved):

precious-onyenaucheya-ons (approved):

rmccar (changes requested):
Just putting a blocking review so I can have a quick chat this afternoon before this is merged

rmccar:
After testing document.domain and window.location.hostname return the same thing. The issue seems to be with the setting of the domain rather than changing where it comes from

src/js/cookies-functions.js
```diff
@@ -74,8 +74,7 @@ export function getConsentCookie() {
 }
 
 export function setConsentCookie(options) {
-    const domain = getDomain(document.domain);
```
rmccar, line 77: I think we can remove this change now

src/js/cookies-functions.js
```diff
@@ -149,6 +148,7 @@ export function setCookie(name, value, options) {
         if (document.location.protocol === 'https:') {
             cookieString = cookieString + '; Secure';
         }
+        cookieString = cookieString + '; Secure';
```
rmccar, line 151: Was this left in by mistake? It now looks like we are doing this inside and outside of the if statement

src/js/cookies-functions.js
```diff
@@ -169,13 +169,18 @@ export function getCookie(name) {
...
+    if (domain.startsWith('www.')) {
+        domain = domain.substring(4);
+    }
     let i = 0,
         domainName = domain,
         p = domainName.split('.'),
         s = '_gd' + new Date().getTime();
     while (i < p.length - 1 && document.cookie.indexOf(s + '=' + s) == -1) {
-        domainName = p.slice(-1 - ++i).join('.');
+        domainName = p.slice(i, p.length).join('.');
```
rmccar, line 180: We will need to add tests for this

src/js/cookies-functions.spec.js
```diff
@@ -11,27 +11,25 @@ describe('script: getDomain()', () => {
                 .map(([key, value]) => `${key}=${value}`)
                 .join('; '),
         set: (value) => {
-            const url = new URL(value);
-            const allowedHosts = ['service-manual.example.com'];
-            if (allowedHosts.includes(url.host)) return;
+            if (value.includes('new-website.example.com')) return; //Prevent setting cookies at 'new-website.example.com'
```
rmccar, line 14: I don't think we want to add logic into tests like this

src/js/cookies-functions.spec.js
```diff
@@ -11,27 +11,25 @@ describe('script: getDomain()', () => {
...
 
-    test('should remove `www` from the domain name', () => {
+    test('should remove `www` from the domain name www.ons.gov.uk', () => {
         const result = getDomain('www.ons.gov.uk');
         expect(result).toBe('ons.gov.uk');
     });
 
-    test('should return a wider domain name if the cookie is not set at subdomain', () => {
-        const result = getDomain('service-manual.example.com');
+    test('returns `example.com` as the domain name when cookies was not set at subdomain `new-website.example.com`', () => {
```
rmccar, line 31:
```suggestion
    test('returns `example.com` as the domain name when cookies can not be set at subdomain `new-website.example.com`', () => {
```

rmccar:
Can you update the description of this PR with more detail. Just because theres quite a bit more to this change

SriHV:
> Can you update the description of this PR with more detail. Just because theres quite a bit more to this change

Updated it now 👍

PR #3535, opened 2025-03-31
Feature/chart aspect ratio

helenb:

https://jira.ons.gov.uk/browse/CCB-45

Adds options to control the chart aspect ratio at desktop and mobile.
Prevents the legend from adding pagination when it gets to a certain height.

### How to review this PR

Check out a variety of options for `percentageHeightDesktop` and `percentageHeightMobile` and ensure that they consistently set the plot height to be that percentage of the plot width.

### Checklist


janehughes (approved):
Works well! LGTM 👍

src/components/chart/_macro-options.md
```diff
@@ -1,19 +1,21 @@
...
+| series                  | array   | true     | The data [series](#series) to be plotted, including labels and values.                                                                                                                       |
+| useStackedLayout        | boolean | false    | Determines whether the chart should use a stacked layout. It is useful only for bar charts.                                                                                                  |
+| percentageHeightDesktop | integer | false    | Sets the percentage plot height at desktop, relative to the width. If undefined the chart will fall back to the default height of 400px at desktop.                                          |
+| percentageHeightMobile  | integer | false    | Sets the percentage plot height at mobile, relative to the width. If undefined the chart will fall back to the default height of 400px at mobile.                                            |
```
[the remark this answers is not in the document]
helenb: I have updated that now.

rmccar:
Ive noticed that changing the desktop height is reflected on refresh for desktop but not for mobile. For mobile the height change is only reflected when you change to desktop size and then back again

src/components/chart/example-stacked-column-chart.njk
```diff
@@ -60,6 +60,8 @@
         },
         "yAxis": {
             "title": "Percentage of GDP (%)"
-        }
+        },
+        "percentageHeightDesktop": 50,
+        "percentageHeightMobile": 120
```
rmccar, line 65: 120 feels very big for the examples. I think it would be good to reduce this number for the examples

helenb:
> Ive noticed that changing the desktop height is reflected on refresh for desktop but not for mobile. For mobile the height change is only reflected when you change to desktop size and then back again

I haven't managed to repeat that issue

src/components/chart/example-stacked-column-chart.njk
helenb: I've adjusted this now if you could check again.

helenb:
> > Ive noticed that changing the desktop height is reflected on refresh for desktop but not for mobile. For mobile the height change is only reflected when you change to desktop size and then back again
> 
> I haven't managed to repeat that issue - was a it a particular chart type where it happened?

I did see that if I open the inspector and this forces it straight to mobile view, that the chart height is initially very tall. However, testing on my phone this does not happen, and it doesn't happen on resizing, or re-loading the page at that size. I don't think that this is a big concern or that it should block merging. We can raise a separate ticket to investigate if you still think it could be an issue.

rmccar:
> > > Ive noticed that changing the desktop height is reflected on refresh for desktop but not for mobile. For mobile the height change is only reflected when you change to desktop size and then back again
> > 
> > 
> > I haven't managed to repeat that issue - was a it a particular chart type where it happened?
> 
> I did see that if I open the inspector and this forces it straight to mobile view, that the chart height is initially very tall. However, testing on my phone this does not happen, and it doesn't happen on resizing, or re-loading the page at that size. I don't think that this is a big concern or that it should block merging. We can raise a separate ticket to investigate if you still think it could be an issue.

Seems to be working fine now, maybe I was wrong

src/components/chart/example-stacked-column-chart.njk
helenb: I've reduced it a little - but also added a max-width to the chart component as it won't render very wide on the page.

Untitled-1
rmccar, line 1: Added by accident?

PR #3548, opened 2025-04-11
Update Tick Intervals to avoid overlapping

precious-onyenaucheya-ons:

Fixes [ONSDESYS-398](https://jira.ons.gov.uk/browse/ONSDESYS-398)
This PR introduces logic to dynamically calculate and apply sensible default tickInterval values
### How to review this PR

1. Review examples for all chart types (line, column, bar, etc.) on both desktop and mobile breakpoints.
    Confirm that:
- Axis labels do not overlap
- Labels remain horizontal
- Long labels are truncated with ellipses
- TickInterval is adjusted automatically based on screen size and label length
2. Check that all test and visual regression tests pass

### Checklist


helenb:
HI @precious-onyenaucheya-ons let's raise this at stand-up - at the moment the solution is not the one specified on the ticket, which allows the editor to set the tick intervals (both for x and y axes)

PR #3551, opened 2025-04-14
Add tick interval params

precious-onyenaucheya-ons:
Fixes [ONSDESYS-398](https://jira.ons.gov.uk/browse/ONSDESYS-398)
This PR introduces support for the tickInterval parameter, allowing users to define custom spacing for axis tick marks.

### How to review this PR
1. Review examples for all chart types (line, column, bar, etc.) on both desktop and mobile breakpoints.
Confirm that:
- Axis labels do not overlap
- Labels remain horizontal
- Long labels are truncated with ellipses

2. Check that all test and visual regression tests pass

### Checklist


src/components/chart/common-chart-options.js
```diff
@@ -223,6 +224,25 @@ class CommonChartOptions {
         }
     };
 
+    formatAxisLabels = (chart) => {
+        const targetAxis = chart.userOptions.chart.type === 'bar' ? chart.yAxis : chart.xAxis;
+        targetAxis.forEach((axis) => {
+            axis.update({
+                labels: {
+                    useHTML: true,
+                    formatter: function () {
```
helenb, line 233: Query: do we need to do this in a loadEvent? If we set `useHTML` true in the options (https://api.highcharts.com/highcharts/xAxis.labels.useHTML) could we not then use CSS to style the HTML element as we want?

helenb: And a further query: do we definitely need this extra styling? Highcharts will automatically use an ellipsis to hide some of the label if it overlaps - but maybe this adds some improvements to the auto-behaviour?

src/components/chart/_macro-options.md
```diff
@@ -36,19 +36,21 @@
...
+| ------------ | ------ | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| title        | string | false    | The title text displayed on the x-axis                                                                                                                    |
+| labelFormat  | string | false    | A format string for the x-axis label. Examples of string formats can be found in these [docs](https://www.highcharts.com/docs/chart-concepts/templating). |
+| categories   | array  | false    | Labels for each tick mark along the x-axis.                                                                                                               |
+| type         | string | false    | The type of axis. Can be one of `linear`, `logarithmic`, `datetime` or `category`. Defaults to linear.                                                    |
+| tickInterval | number | false    | The interval of the tick marks in axis units. Useful when you want to space out the labels (e.g. show every 2nd or 5th label).                            |
```
helenb, line 53: We need this to be `tickIntervalMobile` and `tickIntervalDesktop` so that the editor can configure it differently for both. You can then update the options in https://github.com/ONSdigital/design-system/blob/main/src/components/chart/common-chart-options.js#L144-L176 to add the mobile tickInterval.

helenb (changes requested):

src/components/chart/common-chart-options.js
precious-onyenaucheya-ons: Yeah i can add the css but with the ellipsis, It wasnt working as expected when set in the options

src/components/chart/_macro-options.md
precious-onyenaucheya-ons: updated

janehughes (commented):
This is working for me locally though the horizontal truncated labels mean it's impossible to read the longer text labels on mobile:

Pasted Graphic 2

Before:
Percentage of GDP

And on the desktop line chart example it feels like worse UX to not see the labels in full:

image

I might be missing some context on this however - but keen to know what the reasoning for this is, it feels like it's going to make most charts with text labels more difficult to read?

helenb:
> This is working for me locally though the horizontal truncated labels mean it's impossible to read the longer text labels on mobile:
> 
> Pasted Graphic 2
> Before: Percentage of GDP
> 
> And on the desktop line chart example it feels like worse UX to not see the labels in full:
> 
> image
> I might be missing some context on this however - but keen to know what the reasoning for this is, it feels like it's going to make most charts with text labels more difficult to read?

Hey @janehughes , thanks for raising these concerns. The idea with this work is to allow an editor in the chart builder in Wagtail to set the tick interval themselves, both for desktop and mobile, to minimise the likelihood of any overlapping text. So they'll have control over how many labels appear. I agree that the ellipsis for the cases where the labels do overlap is less than ideal, but we've told the guidelines definitely don't allow for the angled text. I think it will be a case of trying this out when it is hooked up to the chart builder so we can find out how well it works. It might be that we revert back to the angled text as a fallback as being preferable to cutting off the label.

In the mean time perhaps it might help to choose a slightly bigger tick interval for the examples, @precious-onyenaucheya-ons (note I haven't tested myself yet - I was just responding to @janehughes's comment).

janehughes:
> > This is working for me locally though the horizontal truncated labels mean it's impossible to read the longer text labels on mobile:
> 
> ...
> 
> > And on the desktop line chart example it feels like worse UX to not see the labels in full:
> 
> ...
> 
> > I might be missing some context on this however - but keen to know what the reasoning for this is, it feels like it's going to make most charts with text labels more difficult to read?
> 
> Hey @janehughes , thanks for raising these concerns. The idea with this work is to allow an editor in the chart builder in Wagtail to set the tick interval themselves, both for desktop and mobile, to minimise the likelihood of any overlapping text. So they'll have control over how many labels appear. I agree that the ellipsis for the cases where the labels do overlap is less than ideal, but we've told the guidelines definitely don't allow for the angled text. I think it will be a case of trying this out when it is hooked up to the chart builder so we can find out how well it works. It might be that we revert back to the angled text as a fallback as being preferable to cutting off the label.
> 
> In the mean time perhaps it might help to choose a slightly bigger tick interval for the examples, @precious-onyenaucheya-ons (note I haven't tested myself yet - I was just responding to @janehughes's comment).

Thanks for clarifying! This is looking good besides the cut off text, just need to make sure the visual regression tests are updated 👍

src/components/chart/common-chart-options.js
```diff
@@ -171,10 +175,16 @@ class CommonChartOptions {
                         fontSize: this.constants.mobileFontSize,
                     },
                 },
+                tickInterval: yAxisTickInterval,
             },
         };
     };
 
+    updateDesktopTickIntervals = (config, xAxisTickInterval, yAxisTickInterval) => {
```
helenb, line 183: Suggestion: rather than have this in a separate function, just set it in the options for the x axis and y axis in the `this.option` object above - it's more consistent with the other options that way.

src/components/chart/example-line-chart.njk
```diff
@@ -154,7 +154,9 @@
                 'Aug 2024',
                 'Sep 2024',
                 'Oct 2024'
-            ]
+            ],
+            "tickIntervalDesktop": 7,
```
helenb, line 158: As noted in my response to Jane we might want to use some different examples here to try and avoid the ellipses as much as possible.

helenb (commented):
Thank @precious-onyenaucheya-ons this is looking good now - just a couple of small comments.

helenb (approved):
This is all looking good now and the example tickIntervals on the line chart are much better.

src/components/chart/_chart.scss
```diff
@@ -78,3 +78,8 @@
 .highcharts-axis-title {
     width: 100%;
 }
+
+.highcharts-xaxis-labels span {
+    display: block !important;
+    white-space: nowrap !important;
+}
```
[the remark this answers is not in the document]
precious-onyenaucheya-ons: I've moved this to CSS because Highcharts doesn't support display styles directly. Also, the white-space style was being applied to the x-axis in bar charts—which we don't want. This update ensures the styling is applied to all chart types except bar charts.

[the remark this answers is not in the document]
admilne: I could have sworn I had `display: 'block'` working but jumped on a call with Precious and sure enough it wasn't. Either I'm going mad and it never worked or its flakey, either way I'm happy to approve.

janehughes (approved):
Can see visual regression tests have been updated and new thumbnails created 👍

PR #3545, opened 2025-04-11
Scatter chart, update markers and small fixes

helenb:
### What is the context of this PR?

- Adds a Scatter chart (https://jira.ons.gov.uk/browse/CCB-20)
- Updates the markers so that a different one is used for each series (https://jira.ons.gov.uk/browse/CCB-38)
- Removes the mobile font sizes (https://jira.ons.gov.uk/browse/CCB-77)
- Adds connectNulls (https://jira.ons.gov.uk/browse/CCB-85)
- Fixes zero line width (https://jira.ons.gov.uk/browse/CCB-83)
- Fixes legend hover (https://jira.ons.gov.uk/browse/CCB-83)

### How to review this PR

Test in your local build and review the following new examples:

- http://localhost:3000/components/chart/example-line-chart-with-markers.html
- http://localhost:3000/components/chart/example-scatter-chart.html

Also review all the existing examples as font-sizes have changed, zero line width updated, and legend items should no longer hover.

### Checklist

This needs to be completed by the person raising the PR.

-   [x] I have selected the correct Assignee
-   [x] I have linked the correct Issue

PR #3553, opened 2025-04-15
Add example area chart

helenb:

https://jira.ons.gov.uk/browse/CCB-21
Adds an example area chart
Also adds code to fail gracefully if the chart type is not supported

### How to review this PR

### Checklist

### DIFF
--- a/src/components/chart/_macro.njk
+++ b/src/components/chart/_macro.njk
@@ -1,7 +1,7 @@
 {% from "components/list/_macro.njk" import onsList %}
 
 {% macro onsChart(params) %}
-    {% set supportedChartTypes = ['line', 'bar', 'column'] %}
+    {% set supportedChartTypes = ['line', 'bar', 'column', 'scatter', 'area'] %}
 
     {% if params.headingLevel and params.headingLevel >= 1 and params.headingLevel <= 4 %}
         {% set headingLevel = params.headingLevel %}
@@ -55,6 +55,8 @@
                     <figcaption class="ons-chart__caption">{{ params.caption }}</figcaption>
                 {% endif %}
             </figure>
+        {% else %}
+            <p><b>Chart type "{{ params.chartType }}" is not supported</b></p>
         {% endif %}
 
         {% if params.download.title and params.download.itemsList | length > 0 %}
--- a/src/components/chart/area-chart.js
+++ b/src/components/chart/area-chart.js
@@ -1,0 +1,26 @@
+class AreaChart {
+    getAreaChartOptions = () => {
+        return {
+            legend: {
+                symbolWidth: 12,
+                symbolHeight: 12,
+            },
+            plotOptions: {
+                area: {
+                    fillOpacity: 1,
+                    // Use a circle instead of the default blob plus line icon
+                    // 'rectangle' counterintuitively gives a circle, because the legend icon has a border radius of half it's height by default
+                    legendSymbol: 'rectangle',
+                    stacking: 'normal',
+                },
+                series: {
+                    marker: {
+                        enabled: false,
+                    },
+                },
+            },
+        };
+    };
+}
+
+export default AreaChart;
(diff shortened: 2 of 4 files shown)


janehughes:
While testing I noticed the mouse cursor is changing to a pointer when hovering over the legend items, which I thought had been updated? It's the same for all charts, not just this one. Otherwise, LGTM!

helenb:
> While testing I noticed the mouse cursor is changing to a pointer when hovering over the legend items, which I thought had been updated? It's the same for all charts, not just this one. Otherwise, LGTM!

That fix is in https://github.com/ONSdigital/design-system/pull/3545 which is not yet merged :)

src/components/chart/_macro-options.md
```diff
@@ -1,6 +1,6 @@
 | Name                    | Type    | Required | Description                                                                                                                                                                                  |
 | ----------------------- | ------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| chartType               | string  | true     | The type of chart to render. Supported types include: 'line', 'bar' and 'column'.                                                                                                            |
+| chartType               | string  | true     | The type of chart to render. Supported types include: 'line', 'bar', 'column', 'area' and 'scatter'.                                                                                         |
```
rmccar, line 3:
```suggestion
| chartType               | string  | true     | The type of chart to render. Supported types include: 'line', 'bar', 'column', and 'area'.                                                                                                   |
```
I know its small but to keep changes separate, I think we should only be adding in changes that relate to this PR

src/components/chart/_macro.njk
```diff
@@ -1,7 +1,7 @@
 {% from "components/list/_macro.njk" import onsList %}
 
 {% macro onsChart(params) %}
-    {% set supportedChartTypes = ['line', 'bar', 'column'] %}
+    {% set supportedChartTypes = ['line', 'bar', 'column', 'scatter', 'area'] %}
```
rmccar, line 4:
```suggestion
    {% set supportedChartTypes = ['line', 'bar', 'column', 'area'] %}
```
Same here

src/components/chart/_macro.njk
```diff
@@ -67,6 +67,8 @@
                     <figcaption class="ons-chart__caption">{{ params.caption }}</figcaption>
                 {% endif %}
             </figure>
+        {% else %}
+            <p><b>Chart type "{{ params.chartType }}" is not supported</b></p>
```
rmccar, line 71: We should probably add a test for this

helenb: I've added one. I also changed it so that only the chart node was not rendered if the chart type is invalid - so we're showing all the other chart surroundings still (this is now more consistent as previously the title and subtitle were hidden, but the downloads still showed).

src/components/chart/_macro-options.md
helenb: Now updated.

src/components/chart/_macro.njk
helenb: Now updated.

src/components/chart/_macro.njk
rmccar: Ive just noticed that I dont think we have tests for the title and id params, unless Ive missed them

helenb: I'll check that now...

helenb: The id check was there - I have added a check for the title. I've also remove the config for an invalid chart, and added more thorough testing for the invalid chart.

rmccar: Great! I can see the `data-highcharts-id` attribute being tested but not the `id` attribute

helenb: I don't know why that `id` attribute was added as it isn't needed. I'd prefer to remove it rather than add a test for it. Is that ok as part of this PR or would it be better kept separate?

rmccar: Yeah just remove it in this PR then and mention it in the description

rmccar: Have you pushed? Doesn't look like the id has been removed. I was talking about the one added on line 25 in the macro

helenb: I hadn't pushed - should be there now.

janehughes (approved):

precious-onyenaucheya-ons (approved):

PR #3556, opened 2025-04-22
Update column chart spacing to follow Data Viz Guidelines

precious-onyenaucheya-ons:
Fixes [ONSDESYS-425](https://jira.ons.gov.uk/browse/ONSDESYS-425)

This PR refactors the logic for updating column chart pointPadding, groupPadding, and maxPointWidth for both desktop and mobile views based on the Data Viz guidance outlined in [Notion guide](https://ons-design.notion.site/Chart-types-e916b8337b064aab9d4ee0db0f56eb86?p=2fe3112a5f4347ddb8120e737f500af6&pm=c).

The goal is to ensure consistent spacing and bar widths based on the number of data points and the screen size.

### How to review this PR

- Confirm that column spacing and widths follow the design system guidance.
- Verify that maxPointWidth, pointPadding, and groupPadding are set correctly based on the number of bars and viewport size.
- Test responsiveness by resizing the browser window or toggling device view in dev tools.
- Check that updated visual tests reflect the new spacing rules accurately.

### Checklist


src/components/chart/column-chart.js
```diff
@@ -15,31 +15,33 @@ class ColumnChart {
...
-            if (numberOfCategories > 5) {
-                pointPadding = 0.03;
-            } else {
-                pointPadding = 0.04;
-            }
+            pointPadding = numberOfCategories >= categoryThreshold ? 0.1 : 0.2;
         } else {
             // clustered charts
-            groupPadding = 0.04;
+            groupPadding = numberOfCategories >= categoryThreshold ? 0.1 : 0.2;
```
rmccar, line 36: There is some repetition here can we refactor this so that we don't need to repeat this?

src/components/chart/column-chart.js
```diff
@@ -15,31 +15,33 @@ class ColumnChart {
         };
     };
 
-    // Set the point padding between each bar to be 3% (an overall gap of 6%)
-    // For charts with fewer than 5 categories, we use a wider point padding of 4% (8% gap between bars)
-    // For cluster charts we use 0 for the point padding and a group padding of 4% (8% gap between bars)
-    updatePointPadding = (config, currentChart, stackedLayout, numberOfExtraLines) => {
+    // Set the point padding between each bar
+    // For charts with fewer than 5 categories, we use a wider point padding of 2%
+    // For cluster charts we use 0 for the point padding and a group padding of 2%
```
rmccar, line 20: This comment now isn't that clear, what is the default now? is 2% wider than it?

rmccar:
Are there no tests to update to cover these changes and make sure the chart spacing calculations are correct?

src/components/chart/column-chart.js
```diff
@@ -15,31 +15,33 @@ class ColumnChart {
         };
     };
 
-    // Set the point padding between each bar to be 3% (an overall gap of 6%)
-    // For charts with fewer than 5 categories, we use a wider point padding of 4% (8% gap between bars)
-    // For cluster charts we use 0 for the point padding and a group padding of 4% (8% gap between bars)
-    updatePointPadding = (config, currentChart, stackedLayout, numberOfExtraLines) => {
+    // Set the point padding between each bar
+    // For charts with fewer than 5 categories, we use a wider point padding of 2%
```
helenb, line 19: These comments need clarifying as they don't reflect the new logic.

src/components/chart/column-chart.js
```diff
@@ -15,31 +15,33 @@ class ColumnChart {
...
         let groupPadding = 0;
+
         // non-clustered charts or stacked charts
         if (numberOfSeries === 1 || stackedLayout === true) {
-            if (numberOfCategories > 5) {
-                pointPadding = 0.03;
-            } else {
-                pointPadding = 0.04;
-            }
+            pointPadding = numberOfCategories >= categoryThreshold ? 0.1 : 0.2;
```
helenb, line 33: I don't think this is quite correct - we're ending up with a spacing of about 40% at desktop. If you look at my original comments, they clarify how the the pointPadding relates to the final overall gap - it's like padding in CSS - if you had 2 divs next to each other, each with a 20px left and right padding, you'd end up with an overall gap of 40px.

src/components/chart/column-chart.js
```diff
@@ -15,31 +15,33 @@ class ColumnChart {
...
-    // For cluster charts we use 0 for the point padding and a group padding of 4% (8% gap between bars)
-    updatePointPadding = (config, currentChart, stackedLayout, numberOfExtraLines) => {
+    // Set the point padding between each bar
+    // For charts with fewer than 5 categories, we use a wider point padding of 2%
+    // For cluster charts we use 0 for the point padding and a group padding of 2%
+    updatePointPadding = (config, currentChart, stackedLayout, numberOfExtraLines, isMobile = false) => {
         const numberOfCategories = config.xAxis.categories.length;
         const numberOfSeries = currentChart.series.length - numberOfExtraLines; // Get number of column series
+        // Guidelines
+        const categoryThreshold = isMobile ? 10 : 20;
```
helenb, line 25: The logic to test for `isMobile` isn't working here. It's true the entire time that you call this from the window resize event. Rather than call this on resize and load, I think we might need to refactor things a bit, so that we can set the config initially for both desktop and mobile before we call `Highcharts.chart`. It will take a bit of refactoring though. I'll send some example code.

helenb (changes requested):
There are a few issues here - I'll try and send some suggestions for the refactoring.

src/components/chart/column-chart.js
helenb: See my demo (which needs some tidying up) of how this could be refactored: https://github.com/ONSdigital/design-system/pull/3557

helenb:
Thanks @precious-onyenaucheya-ons this is now looking good and working to set the spacing as per the guidelines.

I tested a column chart with 20 bars, and in that scenario there are still issues with the columns joining together at mobile:

Screenshot 2025-04-24 at 09 12 47

The ACs say we should adapt the guidelines if needed, so I think the solution is to make the spacing a bit wider than the guidelines say at mobile.

GIVEN I view a column chart in the design system service manual    
WHEN I look at the example  
THEN I can see that the column gaps appear even at all breakpoints and the gap never disappears
AND (IF POSSIBLE) the gaps follow the updated guidelines (screenshot below)
AND IF NECESSARY the guideline spaces are adapted slightly in order to achieve the first condition

src/components/chart/column-chart.js
```diff
@@ -39,7 +39,7 @@ class ColumnChart {
 
         let pointPadding = 0;
         let groupPadding = 0;
-        let spacing = numberOfCategories >= categoryThreshold ? 0.05 : 0.1;
+        let spacing = numberOfCategories >= categoryThreshold ? 0.1 : 0.2;
```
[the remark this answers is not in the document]
precious-onyenaucheya-ons: Sure will do

janehughes (changes requested):
This is looking good, however It looks like the scatter chart isn't rendering correctly, it isn't loading for me locally and I can see the regression tests have rendered an empty chart as well.

precious-onyenaucheya-ons:
> This is looking good, however It looks like the scatter chart isn't rendering correctly, it isn't loading for me locally and I can see the regression tests have rendered an empty chart as well.

Nice spot @janehughes 

Seems like a minor bug was introduced, but its all fixed now
I will update the VR tests as well

janehughes (approved):

Issue #3559, opened 2025-04-24
[Feature]: Heading link should be optional for Headline Figures card variant


janehughes:
### What feature would you like to add to the ONS Design System?

We need to make sure the card link is optional for the Headline Figures variant. This is required to allow for an unlinked title in the Headline Figures cards:

https://www.figma.com/design/Yf4Cpxc8vYdoVrWcJojk2R/New-Website-Platform?node-id=5135-22287&t=ePjxFpxOHdyBa6AS-0

### Why should this new feature be added to the Design System?

The Headline Figures cards on the statistical article page may not always have a link provided by the user, so we need to make sure it's an optional parameter in the component.

PR #3560, opened 2025-04-24
Bug: Update card title URL parameter to 'optional'

janehughes:

This fix comes off the back of a bug raised after creating the Headline Figures card variant (relates to https://github.com/ONSdigital/design-system/issues/3559). 

Ticket: https://jira.ons.gov.uk/browse/ONSDESYS-430

This flagged an existing bug in the card component that it would always add a link even if a URL isn't provided, which meant the URL might be empty.

This PR makes the URL optional and updates the macro to remove the link if a URL isn't provided.

### How to review this PR

- Test adding & removing a URL to the 'title' object in the card component, and see that the card title displays a link if a URL is provided, otherwise it displays static text (e.g. http://localhost:3000/components/card/example-card-set-with-images.html)

### Checklist


MaciekBaron (approved):

backstop_data/bitmaps_reference/ds-vr-test__components_card_example-card-set-with-headline-figures_0_document_0_desktop.png
SriHV, line 1: I think we can keep the url in the example. We might add this in service manual in future and it can be helpful for the users to know how it looks when url is present. But this might just be my personal preference. Ignore this if there is discussion on removing it already :)

janehughes: We still have the url included in all other card examples, I've just removed it from the headline figures one so we do still have an example of each :)

PR #3566, opened 2025-04-28
Remove functionality to add  bar with line chart

precious-onyenaucheya-ons:

This PR addresses [CCB-89](https://jira.ons.gov.uk/browse/CCB-89).

It removes all functionality related to adding a bar with a line chart combination.

### How to review this PR

1. Verify that attempting to add a bar with a line chart now defaults to a standard bar chart.
2. Confirm that there are no remaining examples of a bar with a line chart.
3. Ensure all unit and visual tests pass successfully.
4. Confirm that the changes made are strictly limited to what is described in the ticket.

### Checklist


janehughes (commented):
This is working well for me locally & tests are passing locally except for the visual regressions, looks like they might need re-running

src/components/summary/_macro.njk
```diff
@@ -104,7 +104,10 @@
                                             </dd>
                                         {% endif %}
                                         {% if item.actions %}
-                                            <dd class="ons-summary__actions{{ ' ons-summary__column-size--2' if not item.valueList }}" {% if item.id %}id="{{ item.id }}"{% endif %}>
+                                            <dd
```
helenb, line 107: Nitpick: this one change doesn't look related to the ticket.

helenb (commented):
Thanks @precious-onyenaucheya-ons this all seems to work as expected. There is another bit of clean-up we can do.

https://github.com/ONSdigital/design-system/blob/main/src/components/chart/bar-chart.js#L130

We no longer need to account for the `numberOfExtraLines` when calculating the number of series in a bar chart - and can also remove the parameter for it from the `updateBarChartHeight` function.

I also spotted when testing by using other chart types that it is possible to display dataLabels on chart types other than bar charts. We probably need to consider limiting that (but one for another PR)

helenb:
> We no longer need to account for the `numberOfExtraLines` when calculating the number of series in a bar chart - and can also remove the parameter for it from the `updateBarChartHeight` function.

Thanks - I can see that is now removed.

src/components/summary/_macro.njk
precious-onyenaucheya-ons: Removed

helenb (approved):

janehughes (approved):

PR #3567, opened 2025-04-29
Fix accessibility console error in Document List

SriHV:
[ONSDESYS-249](https://jira.ons.gov.uk/browse/ONSDESYS-249)

This PR fixes accessibility issue caused due to available focus on `<a>` link inside a `<div>` that has `aria-hidden="true"` . 
Changing aria-hidden to inert fixed the issue.

### How to review this PR

Right click and inspect on the placeholder image in `example-document-list-articles`. See that there is no warning or error message now
you will then see this error in the console

### Checklist


src/components/document-list/_macro.njk
```diff
@@ -9,10 +9,7 @@
             {% set closingTag = "</h" ~ titleTag ~ ">" %}
             {% set documentItem %}
                 {% if document.thumbnail %}
-                    <div
-                        class="ons-document-list__item-image{{ ' ons-document-list__item-image--file' if document.metadata.file }}"
-                        aria-hidden="true"
-                    >
+                    <div class="ons-document-list__item-image{{ ' ons-document-list__item-image--file' if document.metadata.file }}" inert>
```
janehughes, line 12: This `inert` attribute is disabling any kind of interactivity from the child elements (for all users, not just accessibility software) which makes the image link entirely redundant here anyway. 

Why were we using `aria-hidden` originally? I assume it's because the link doesn't have any alt text? If so, rather than trying to force accessibility software to ignore the link we could add an alternative label so they can still access it. We could use `aria-labelledby` and point it to an id on the card title, so that the link will read the card title instead - this means the image will still be treated as decorative (and ignored) by assistive tech and the meaningful title text would be read instead. 

Alternatively, we could remove the link from the image entirely if it's not used or needed anymore?

janehughes (changes requested):

src/components/document-list/_macro.njk
SriHV: We didn't have id attribute for each title so, I've used `aria-label`. Can you tell me if this is okay?

src/components/document-list/_macro.njk
```diff
@@ -11,7 +11,7 @@
                 {% if document.thumbnail %}
                     <div
                         class="ons-document-list__item-image{{ ' ons-document-list__item-image--file' if document.metadata.file }}"
-                        aria-hidden="true"
+                        aria-label="{{ document.title.text }}"
```
janehughes, line 14: We need to move this to the `<a>` rather than the `<div>`, as `aria-label`s are only valid on focusable elements not static ones like `<div>`s so this would be ignored by SRs.

It's not ideal having two links with the same label here but maybe this is the most pragmatic solution if we need to keep this link in.

Can we also remove the `tabindex="-1"` on the image link, so users can access it via the keyboard?

src/components/document-list/_macro.njk
janehughes: Thanks Sri, have updated with another comment

rmccar:
After these changes, if you use a screenreader the title is now being read out twice, because of the aria label added to the image. This might be why `aria-hidden` was added to it. Janes suggestion of using `aria-labelledby` might work better

SriHV:
I've added new `id` to the `title` param. Could you tell me if this is okay @rmccar  and @janehughes. Cause when I used screenreader, title is still read twice. I am not completely sure about this.

janehughes:
> I've added new `id` to the `title` param and using `aria-labelled-by` now. Could you tell me if this is okay @rmccar and @janehughes. Cause when I used screenreader, title is still read twice. I am not completely sure about this.

Yeah that's expected as we still have 2 links so it will read out both.

Ideally we'd combine them into one link (wrapping the entire card in a single `<a>` tag) and point the `aria-labelledby` to the title's `id` so that the link only reads the title aloud and ignores the rest of the content - does that make sense? This way the entire block is clickable (good for UX), but screenreaders don't read out the entire content of the card. 

Alternatively, we just remove the second image link and keep one link on the title, but I'm guessing the image link has been included for improved UX so users can click the image too. 

@rmccar do you have thoughts on the above? It feels like a good opportunity to improve the UX - removing the `aria-hidden` is a quick fix to remove the error, but doesn't resolve the accessibility/usability issues entirely.

rmccar:
> > I've added new `id` to the `title` param and using `aria-labelled-by` now. Could you tell me if this is okay @rmccar and @janehughes. Cause when I used screenreader, title is still read twice. I am not completely sure about this.
> 
> Yeah that's expected as we still have 2 links so it will read out both.
> 
> Ideally we'd combine them into one link (wrapping the entire card in a single `<a>` tag) and point the `aria-labelledby` to the title's `id` so that the link only reads the title aloud and ignores the rest of the content - does that make sense? This way the entire block is clickable (good for UX), but screenreaders don't read out the entire content of the card.
> 
> Alternatively, we just remove the second image link and keep one link on the title, but I'm guessing the image link has been included for improved UX so users can click the image too.
> 
> @rmccar do you have thoughts on the above? It feels like a good opportunity to improve the UX - removing the `aria-hidden` is a quick fix to remove the error, but doesn't resolve the accessibility/usability issues entirely.

Yeah I think the first option sounds good but I'd like to run it past the designers. I think this is better than removing the link from the image because I wouldn't want to remove that functionality, I think that would make the UX worse. Would you be able to have a chat with Dina/Joe about this Sri. If they are happy I think have a look at going with the first option

SriHV:
> > > I've added new `id` to the `title` param and using `aria-labelled-by` now. Could you tell me if this is okay @rmccar and @janehughes. Cause when I used screenreader, title is still read twice. I am not completely sure about this.
> > 
> > 
> > Yeah that's expected as we still have 2 links so it will read out both.
> > Ideally we'd combine them into one link (wrapping the entire card in a single `<a>` tag) and point the `aria-labelledby` to the title's `id` so that the link only reads the title aloud and ignores the rest of the content - does that make sense? This way the entire block is clickable (good for UX), but screenreaders don't read out the entire content of the card.
> > Alternatively, we just remove the second image link and keep one link on the title, but I'm guessing the image link has been included for improved UX so users can click the image too.
> > @rmccar do you have thoughts on the above? It feels like a good opportunity to improve the UX - removing the `aria-hidden` is a quick fix to remove the error, but doesn't resolve the accessibility/usability issues entirely.
> 
> Yeah I think the first option sounds good but I'd like to run it past the designers. I think this is better than removing the link from the image because I wouldn't want to remove that functionality, I think that would make the UX worse. Would you be able to have a chat with Dina/Joe about this Sri. If they are happy I think have a look at going with the first option

I spoke to Joe, he suggested to remove the link for the image as its just decorative rather than wrapping the card with a link.

janehughes (approved):

PR #3577, opened 2025-05-08
Fix legend parameter and add better tests

helenb:

The 'legend' parameter was not previously functioning as expected. It was being displayed even when the legend parameter was set to be false. This PR addresses the issue and adds better tests.

https://jira.ons.gov.uk/browse/CCB-90

### How to review this PR

Check the tests run OK and experiment with different values for the `legend` parameter, as well as checking that it defaults to displaying the legend if the parameter is unset.

### Checklist


janehughes (approved):
Setting to true and false works as expected 👍 If I set legend to an empty string i.e. `"legend": "",` that also counts as false and the legend doesn't show, which I assume is fine as that's what I'd expect?

nimasmi:
> Setting to true and false works as expected 👍 If I set legend to an empty string i.e. `"legend": "",` that also counts as false and the legend doesn't show, which I assume is fine as that's what I'd expect?

I think this is an unfortunate side effect, but necessary. Really the test should be for whether the legend is a boolean (as required by the [macro spec](https://github.com/ONSdigital/design-system/blob/main/src/components/chart/_macro-options.md)), and should use the default ("show legend") when it's badly set. But we can't use triple equals `===` to achieve this, because it then breaks in Jinja2.

Since `"legend": "",` is not a supported case, perhaps it's fine for it to behave either way. :shrug:

helenb:
> > Setting to true and false works as expected 👍 If I set legend to an empty string i.e. `"legend": "",` that also counts as false and the legend doesn't show, which I assume is fine as that's what I'd expect?
> 
> I think this is an unfortunate side effect, but necessary. Really the test should be for whether the legend is a boolean (as required by the [macro spec](https://github.com/ONSdigital/design-system/blob/main/src/components/chart/_macro-options.md)), and should use the default ("show legend") when it's badly set. But we can't use triple equals `===` to achieve this, because it then breaks in Jinja2.
> 
> Since `"legend": "",` is not a supported case, perhaps it's fine for it to behave either way. 🤷
> 
> Addendum: the hammer-nut option is to implement a custom filter for this. I don't expect we need that.

Yes, this isn't ideal, and comes down to it being hard to distinguish between an explicitly set boolean value for a particular variable, and whether that variable is defined at all. But I think it is 'good enough' and certainly fixes the previous broken behaviour.

src/components/chart/_macro.spec.js
```diff
@@ -240,6 +243,36 @@ describe('Macro: Chart', () => {
                 });
             });
         });
+
+        describe('GIVEN: Params: Legend is set to true', () => {
```
rmccar, line 247:
```suggestion
        describe('GIVEN: Params: Legend', () => {
```

src/components/chart/_macro.spec.js
```diff
@@ -240,6 +243,36 @@ describe('Macro: Chart', () => {
...
+            describe('WHEN: legend parameter is provided and set to true', () => {
+                const $ = cheerio.load(renderComponent('chart', EXAMPLE_LINE_CHART_WITH_LEGEND_TRUE_PARAMS));
+                test('THEN: it renders the legend', () => {
+                    const configScript = $(`script[data-highcharts-config--line-chart-legend-true-123]`).html();
+                    expect(configScript).toContain('"legend":{"enabled":true}');
+                });
+            });
+        });
+
+        describe('GIVEN: Params: Legend is set to false', () => {
```
rmccar, line 257:
```suggestion
        describe('GIVEN: Params: Legend', () => {
```

src/components/chart/_macro.spec.js
```diff
@@ -240,6 +243,36 @@ describe('Macro: Chart', () => {
...
+            describe('WHEN: legend parameter is provided and set to false', () => {
+                const $ = cheerio.load(renderComponent('chart', EXAMPLE_LINE_CHART_WITH_LEGEND_FALSE_PARAMS));
+                test('THEN: it does not render the legend', () => {
+                    const configScript = $(`script[data-highcharts-config--line-chart-legend-false-123]`).html();
+                    expect(configScript).toContain('"legend":{"enabled":false}');
+                });
+            });
+        });
+
+        describe('GIVEN: Params: Legend is not provided', () => {
```
rmccar, line 267:
```suggestion
        describe('GIVEN: Params: Legend', () => {
```

src/components/chart/_test-examples.js
```diff
@@ -46,6 +46,47 @@ export const EXAMPLE_LINE_CHART_WITH_CONFIG_PARAMS = {
...
+    title: 'Example Line Chart with legend unset',
+    subtitle: 'A sample subtitle',
+    id: 'line-chart-legend-unset-123',
+    xAxis: { title: 'X Axis Title', categories: ['Jan', 'Feb', 'Mar'] },
+    yAxis: { title: 'Y Axis Title' },
+    series: [
+        { name: 'Series 1', data: [10, 20, 30] },
+        { name: 'Series 2', data: [15, 25, 35] },
+    ],
+};
```
rmccar, line 88: Theres a lot of repitition here, you could just have one example with the same config as "EXAMPLE_LINE_CHART_WITH_LEGEND_UNSET_PARAMS" then in the test just use the same example each time and add the legend param setting in the test itself. Its a pattern we have used elsewhere

src/components/chart/_macro.njk
```diff
@@ -104,7 +104,7 @@
         {%
             set config = {
                 "legend": {
-                    "enabled" : params.legend if params.legend else true
+                    "enabled" : false if params.legend == false else true
```
rmccar, line 107: This logic would be true if anything is passed into params.legend as a string even "false" or "fghdakgfhdsgfhd". I think this logic should be a bit more explicit to stop this. Also in a few of the examples what is passed into this is a string and some a bool. They should all be a bool.

rmccar (changes requested):

src/components/chart/_macro.spec.js
helenb: Query: in this case we wouldn't be passing a legend parameter - is there a better 'given' we could use here?

src/components/chart/_macro.njk
helenb: Update: it looks like I can check if the parameter is defined or not with `{% if params.legend is defined %}`, and testing suggests this works in Wagtail too. cc @nimasmi .

nimasmi: Good spot. I wasn't aware of this, but it is part of Jinja2 (see [Test syntax docs](https://jinja.palletsprojects.com/en/stable/templates/#tests) and [`jinja_tests.defined()`](https://jinja.palletsprojects.com/en/stable/templates/#jinja-tests.defined)).

src/components/chart/_test-examples.js
helenb: This should now be addressed.

src/components/chart/_macro.spec.js
helenb: Updated as suggested.

src/components/chart/_macro.spec.js
helenb: Updated as suggested.

src/components/chart/_macro.spec.js
helenb: (I have updated as suggested for now)

src/components/chart/_macro.njk
helenb: This should now be more robust in its checking, and should still work OK in Wagtail. @nimasmi @janehughes

helenb: I've realised the `===` has crept back in which won't work in Wagtail, so I need to address this.

helenb: Sorted.

src/components/chart/_macro.spec.js
```diff
@@ -240,6 +241,61 @@ describe('Macro: Chart', () => {
...
+                        ...EXAMPLE_LINE_CHART_WITH_LEGEND_UNSET_PARAMS,
+                        legend: 'false',
+                    }),
+                );
+                test('THEN: it renders the legend', () => {
+                    const configScript = $(`script[data-highcharts-config--line-chart-legend-tests-123]`).html();
+                    expect(configScript).toContain('"legend":{"enabled":true}');
+                });
+            });
+        });
```
rmccar, line 298: Wouldnt we want this to not render the legend? Its a bit confusing to be able to provide any string even "false" and the legend be rendered

nimasmi: This is a parameter that takes a boolean, and has a default value. The options for when a non-boolean value are passed in are:

1. Use the default value (which it does now).
2. Refuse to render (do we have any other parameters that are strict like this?).
3. Add a special case for some flavours of wrongly-typed values whose string representation look a bit like those of values we would allow.

Option three sounds undesirable from a logical and code complexity point of view. Would this be more palatable if we were to render no chart, or change the default to `false`? What was the decision process for defaulting to `true`? What would it take to change that?

src/components/chart/_macro.spec.js
```diff
@@ -240,6 +241,61 @@ describe('Macro: Chart', () => {
                 });
             });
         });
+
+        describe('GIVEN: Params: Legend', () => {
```
rmccar, line 245: Here we only need one given for these 4 tests

src/components/chart/_macro.spec.js
rmccar: I think this test is concerning the legend param though and thats the style we have adopted when writing our tests

src/components/chart/_macro.spec.js
rmccar: Yeah happy to go with the first option

src/components/chart/_macro.spec.js
helenb: Sorted.

rmccar (approved):

PR #3578, opened 2025-05-08
Add Word Counter

SriHV:
[ONSDESYS-320](https://jira.ons.gov.uk/browse/ONSDESYS-320) Added word counter to the textarea which counts number of words and updates the allowed limit in realtime. Once the reaches the allowed limit , user is informed they are over the word limit rather than prevent them from typing anything else
Added new test cases in text area to test this 

### How to review this PR
Check that new example `example-textarea-with-word-limit` follows the acceptance criteria mentioned in [ONSDESYS-320](https://jira.ons.gov.uk/browse/ONSDESYS-320)

### Checklist


src/components/textarea/textarea.dom.js
```diff
@@ -2,11 +2,18 @@ import domready from '../../js/domready';
 
 async function initialise() {
     const limitedInputs = [...document.querySelectorAll('.ons-js-char-limit-input')];
+    const wordlimitedInputs = [...document.querySelectorAll('.ons-js-word-limit-input')];
+    console.log(wordlimitedInputs);
```
MaciekBaron, line 6: NP: Leftover `console.log()`
```suggestion
```

src/components/word-limit/word-limit.js
```diff
@@ -0,0 +1,82 @@
...
+    updateLimitReadout(event, firstRun) {
+        const wordCount = this.countWords(this.input.value);
+        const remaining = this.maxWords - wordCount;
+        const message = remaining === 1 ? this.singularMessage : this.pluralMessage;
+
+        // Prevent aria live announcement when component initialises
+        if (!firstRun && event.inputType) {
+            this.limitElement.setAttribute('aria-live', 'polite');
+            this.limitElement.setAttribute('aria-live', [remaining > 0 ? 'polite' : 'assertive']);
+            console.log(this.limitElement);
```
MaciekBaron, line 51: NP: Leftover `console.log()`

```suggestion
```

MaciekBaron (changes requested):

janehughes (commented):
Added some suggested wording improvements to the tests, just to make them a bit easier to understand with consistent naming ('readout' doesn't feel very clear to me, and isn't consistent with the rest of the markup naming)

src/components/word-limit/word-limit.js
```diff
@@ -0,0 +1,82 @@
...
+
+    updateLimitReadout(event, firstRun) {
+        const wordCount = this.countWords(this.input.value);
+        const remaining = this.maxWords - wordCount;
+        const message = remaining === 1 ? this.singularMessage : this.pluralMessage;
+
+        // Prevent aria live announcement when component initialises
+        if (!firstRun && event.inputType) {
+            this.limitElement.setAttribute('aria-live', 'polite');
+            this.limitElement.setAttribute('aria-live', [remaining > 0 ? 'polite' : 'assertive']);
```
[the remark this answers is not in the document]
SriHV: You are right, I've removed the first line now 👍 . We are setting `polite` using this condition `remaining > 0 ? 'polite' : 'assertive'` anyway in the next line.

admilne (changes requested):
When you reach the maximum length of words allowed, the component essentially prevents you from entering a space to create a new word. Personally I don't like the way this works and I think it would be better to just go in to negative number of words remaining but I'd like to have a chat with @rmccar to discuss something similar for the character count ticket I'm working on.

admilne (changes requested):
If you paste in text which has too many words in it to begin with you get an error in the console.
image

SriHV:
> If you paste in text which has too many words in it to begin with you get an error in the console. image

This is fixed now 👍

src/components/char-check-limit/character-check.js
```diff
@@ -37,7 +50,8 @@ export default class CharCheck {
 
     updateCheckReadout(event, firstRun) {
         const value = this.input.value;
-        const remaining = this.checkVal - this.getCharLength(value);
+        const getLength = this.input.getAttribute(attrCharCheckVal) ? this.getCharLength(value) : this.getWordLength(value);
```
admilne, line 53: Not keen on `getLength` as a variable name as it sounds like a function. Perhaps change to `currentLength`?

src/components/char-check-limit/character-check.js
```diff
@@ -86,6 +100,14 @@ export default class CharCheck {
                 'ons-u-d-no',
             );
         }
+        if (this.tagName.toLowerCase() === 'textarea') {
```
admilne, line 103: This whole `if/else` statement looks like duplicated code so I think can be removed

src/components/textarea/_macro-options.md
```diff
@@ -18,6 +18,7 @@
 | dontWrap              | boolean                                                       | false                                  | Set to “true” to prevent the textarea from being wrapped in a [field component](/components/field)                                                                |
 | legendIsQuestionTitle | boolean                                                       | false                                  | Creates an `h1` inside the legend. Use when the mutually exclusive textarea is [the only fieldset on the page](/components/fieldset#legend-as-pagequestion-title) |
 | error                 | `Error` [_(ref)_](/components/error)                          | false                                  | Settings for validation errors                                                                                                                                    |
+| wordLimit             | Object`<wordLimit>`                                           | false                                  | Settings for the textarea [word limit counter](#wordlimit)                                                                                                        |
```
admilne, line 21: Perhaps we should add something here to say that wordLimit and charCheckLimit should not be used at the same time - it's one or the other

src/components/word-limit/_macro-options.md
admilne, line 1: Given this isn't a new component, but some new properties added to an existing component I wonder if this should be inside `src/components/char-check-limit`

src/components/word-limit/_macro.njk
admilne, line 1: Given this isn't a new component, but some new properties added to an existing component I wonder if this should be inside `src/components/char-check-limit`

src/components/word-limit/_macro.spec.js
admilne, line 1: Given this isn't a new component, but some new properties added to an existing component I wonder if this should be inside `src/components/char-check-limit`

src/components/word-limit/_test-examples.js
admilne, line 1: Given this isn't a new component, but some new properties added to an existing component I wonder if this should be inside `src/components/char-check-limit`

src/components/char-check-limit/character-check.js
```diff
@@ -18,13 +21,23 @@ export default class CharCheck {
...
+        this.pluralMessage =
+            this.checkElement.getAttribute('data-charcount-plural') || this.checkElement.getAttribute('data-wordcount-plural') || null;
+        this.charLimitSingularMessage =
+            this.checkElement.getAttribute('data-charcount-limit-singular') ||
+            this.checkElement.getAttribute('data-wordcount-limit-singular') ||
+            null;
+        this.charLimitPluralMessage =
+            this.checkElement.getAttribute('data-charcount-limit-plural') ||
+            this.checkElement.getAttribute('data-wordcount-limit-plural') ||
+            null;
```
admilne, line 40: I don't love all the duplication here. Perhaps we can change the data attributes to be more generic? `limit-message-plural` and `limit-message-singular`. 

We just need the user to fill in different properties into the component (unless we give them generic names but this would be a breaking change so should be avoided for now in my opinion) but they can still populate the same data attributes.

src/components/char-check-limit/character-check.js
```diff
@@ -18,13 +19,13 @@ export default class CharCheck {
...
-        this.pluralMessage = this.checkElement.getAttribute('data-charcount-plural') || null;
-        this.charLimitSingularMessage = this.checkElement.getAttribute('data-charcount-limit-singular') || null;
-        this.charLimitPluralMessage = this.checkElement.getAttribute('data-charcount-limit-plural') || null;
+        this.checkElement = document.getElementById(this.input.getAttribute(attrMessageCheckRef));
+        this.checkVal = this.input.getAttribute(attrMessageCheckVal);
+        this.countdown = this.input.getAttribute(attrMessageCheckCountdown) || false;
+        this.singularMessage = this.checkElement.getAttribute('data-message-count-singular') || null;
+        this.pluralMessage = this.checkElement.getAttribute('data-message-count-plural') || null;
+        this.LimitSingularMessage = this.checkElement.getAttribute('data-message-count-limit-singular') || null;
+        this.LimitPluralMessage = this.checkElement.getAttribute('data-message-count-limit-plural') || null;
```
admilne, line 28: This block looks much neater now. Can we rename the data attributes so its explicit that one is the _**over**_ limit message to save confusion? Also, perhaps we can drop the word `count` from the data attribute to make it a little shorter?
```
this.singularMessage = this.checkElement.getAttribute('data-message-singular') || null;
this.pluralMessage = this.checkElement.getAttribute('data-message-plural') || null;
this.overLimitSingularMessage = this.checkElement.getAttribute('data-message-over-limit-singular') || null;
this.overLimitPluralMessage = this.checkElement.getAttribute('data-message-over-limit-plural') || null;
```

src/components/textarea/_macro.njk
```diff
@@ -19,14 +19,18 @@
...
+            class="ons-input ons-input--textarea{{ ' ons-input--error' if params.error }}{{ ' ons-js-char-check-input' if (params.charCheckLimit and params.charCheckLimit.limit) or (params.wordLimit and params.wordLimit.limit) }}{{ textareaExclusiveClass }}{{ ' ' + params.classes if params.classes else '' }}{% if params.width %}{{ ' ' }}ons-input--w-{{ params.width }}{% endif %}"
             {% if params.name %}
                 name="{{ params.name }}"
             {% endif %}
             rows="{{ params.rows | default(8) }}"
             {% if params.charCheckLimit %}
-                data-char-check-ref="{{ params.id }}-check" data-char-check-num="{{ params.charCheckLimit.limit }}"
-                aria-describedby="{{ params.id }}-check" data-char-check-countdown=true
+                data-message-check-ref="{{ params.id }}-check" data-message-check-num="{{ params.charCheckLimit.limit }}"
+                aria-describedby="{{ params.id }}-check" data-messae-check-countdown=true
```
admilne, line 29: Minor typo
```suggestion
                aria-describedby="{{ params.id }}-check" data-message-check-countdown=true
```

src/components/textarea/example-textarea-with-word-limit.njk
```diff
@@ -0,0 +1,20 @@
...
+        "width": "30",
+        "label": {
+            "text": "Please provide some feedback",
+            "description": "For example, describe any difficulties you experienced in the use of this service"
+        },
+        "wordLimit": {
+            "limit": 5,
+            "wordCountSingular": "You have {x} word remaining",
+            "wordCountPlural": "You have {x} words remaining",
+            "wordCountOverLimitSingular": "You are {x} word over the limit",
```
janehughes, line 16: Suggestion: I think we should include the limit number in these 'over limit' warnings i.e. 'You are {x} words over the {limit} word limit' so users have a quick reference of max words allowed - this could be especially useful for people copying & pasting text in.

Happy for this to be considered as a separate enhancement though so we can progress this PR.

janehughes (approved):
Working as expected! I've added a suggestion about including the word limit in the 'over limit' warning as well, but I'm happy to raise this as an issue separately so we can progress this one 👍

src/components/char-check-limit/character-check.js
```diff
@@ -56,9 +58,9 @@ export default class CharCheck {
         if (this.countdown && remaining === 1) {
```
MaciekBaron, line 58: **Question**: I know this is pre-existing code, but I'm trying to understand why `this.countdown` is only checked if `remaining === 1`?

If `this.countdown` is false and `remaining === 1`, then `message = this.pluralMessage`.

I'm looking at `setShowMessage()`. From what I understand, the message is shown:

- When using a `textarea`, or
- When `remaining` characters/words is less than `checkVal` and `this.countdown` is true and `remaining` is greater than 0, or
- When `remaining` is less than 0

So if I'm looking at this correctly, if `this.countdown` is false and `remaining === 1` and it's a `textarea`, it will use `pluralMessage`, which seems wrong to me? If this is the case, I think it would be good to fix this and also write a test to pick this up, so that the textarea uses `this.singularMessage` when `remaining === 1`. If I'm not mistaken, we can just remove `this.countdown` from that condition and it will work as expected.

---

There are some other pre-existing issues, such as using `classList['remove']()` as opposed to `classList.remove()`, and not using `classList.toggle()` when appropriate, but this could be improved in a separate PR.

src/components/textarea/_macro.njk
```diff
@@ -46,6 +50,20 @@
...
+                    "variant": "word",
+                    "limit": params.wordLimit.limit,
+                    "wordCountSingular": params.wordLimit.wordCountSingular,
+                    "wordCountPlural": params.wordLimit.wordCountPlural,
+                    "wordCountOverLimitSingular": params.wordLimit.wordCountOverLimitSingular,
+                    "wordCountOverLimitPlural": params.wordLimit.wordCountOverLimitPlural
+                })
+            %}
+            {% endcall %}
+        {% endif %}
```
rmccar, line 66: Does this also want to be mutually excluisive?

src/components/char-check-limit/_macro-options.md
```diff
@@ -1,9 +1,13 @@
...
+| variant                    | string  | false                                | Set to “check” to add the character limit checker to the input. Set to "word" to count words instead of characters                                                                                 |
+| id                         | string  | true                                 | The HTML `id` of the component                                                                                                                                                                     |
+| limit                      | integer | true                                 | The maximum number of characters allowed in the input                                                                                                                                              |
+| charCountPlural            | string  | true (unless `variant` is set)       | The string displayed when multiple characters can be entered before the limit is reached. Set `{x}` in the string to be replaced with the number, for example “You have {x} characters remaining”. |
```
rmccar, line 6: What if variant is set to "char"?

src/components/char-check-limit/_macro.spec.js
```diff
@@ -58,4 +58,28 @@ describe('FOR: Macro: CharCheckLimit', () => {
...
+        const $ = cheerio.load(renderComponent('char-check-limit', EXAMPLE_WORD_LIMIT));
+
+        test('THEN: passes jest-axe checks', async () => {
+            const results = await axe($.html());
+
+            expect(results).toHaveNoViolations();
+        });
+        test('THEN: has the provided id attribute', () => {
+            expect($('.ons-input__limit').attr('id')).toBe('example-word-limit');
+        });
```
rmccar, line 71: This test seems to be in the wrong place, the id attribute isn't set because the variant is set to word which is what this seems to suggest

src/components/char-check-limit/_macro.spec.js
```diff
@@ -58,4 +58,28 @@ describe('FOR: Macro: CharCheckLimit', () => {
...
+        test('THEN: has the data attribute which defines wordCountSingular', () => {
+            expect($('.ons-input__limit').attr('data-message-singular')).toBe('You have {x} word remaining');
+        });
+        test('THEN: has the data attribute which defines wordCountOverLimitSingular', () => {
+            expect($('.ons-input__limit').attr('data-message-over-limit-singular')).toBe('You have {x} word too many');
+        });
+        test('THEN: has the data attribute which defines wordCountOverLimitPlural', () => {
+            expect($('.ons-input__limit').attr('data-message-over-limit-plural')).toBe('You have {x} words too many');
+        });
```
rmccar, line 83: Same with these, maybe this needs restructuring

src/components/char-check-limit/character-check.js
SriHV: Good point, removed countdown form the file completely we wouldn't require that. Confirmed this with @admilne 

Regarding 
`There are some other pre-existing issues, such as using classList['remove']() as opposed to classList.remove(), and not using classList.toggle() when appropriate, but this could be improved in a separate PR`

Yep we can raise a ticket about this

MaciekBaron: Thanks! I can see that `data-message-check-countdown` is still being referenced in tests, should that be removed from there as well?

MaciekBaron (approved):

rmccar:
Can you tag the github issue in the description

src/components/textarea/_macro-options.md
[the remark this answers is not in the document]
rmccar: This hasnt been moved below `charCheckLimit`

src/components/char-check-limit/_macro.spec.js
```diff
@@ -21,22 +21,22 @@ describe('FOR: Macro: CharCheckLimit', () => {
...
             test('THEN: has the data attribute which defines charCountOverLimitPlural', () => {
-                expect($('.ons-input__limit').attr('data-charcount-limit-plural')).toBe('{x} characters too many');
+                expect($('.ons-input__limit').attr('data-message-over-limit-plural')).toBe('{x} characters too many');
             });
         });
     });
 
     describe('GIVEN: Params: variant', () => {
-        describe('WHEN: variant is provided', () => {
+        describe('WHEN: variant is set to check', () => {
```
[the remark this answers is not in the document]
rmccar: I dont think these tests have been added

src/components/char-check-limit/_macro.spec.js
```diff
@@ -21,40 +21,70 @@ describe('FOR: Macro: CharCheckLimit', () => {
...
 
-            test('THEN: renders the passed content', () => {
-                expect($('.ons-js-char-check-input').text()).toBe('Test content.');
+            test('THEN: sets word as the count type to count number of words', () => {
+                expect($('.ons-input__limit').attr('data-count-type')).toBe('word');
+            });
+        });
+    });
+
+    describe('GIVEN: Params: wordCountPlural', () => {
```
rmccar, line 56: These tests could be restructured to have less repetition they could all come under the same section as variant and then you would only need line 43 once

src/components/textarea/_macro-options.md
```diff
@@ -38,3 +39,14 @@
 | for         | string | true     | Set with the HTML `id` of the textarea the label is for |
 | description | string | false    | Hint text to help users fill in the input               |
 | classes     | string | false    | Classes to add to the label                             |
+
+## WordLimit
```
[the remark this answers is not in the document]
rmccar: This object is now missing

src/components/char-check-limit/_macro.spec.js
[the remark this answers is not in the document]
SriHV: I've added tests when we have default and word. Because this variant type is a string, I think we can set one variant only. We may not require covering both scenarios

src/components/textarea/example-textarea-with-word-limit.njk
```diff
@@ -0,0 +1,20 @@
...
+    onsTextarea({
+        "id": "textarea-word-limit",
+        "name": "feedback-limited",
+        "width": "30",
+        "label": {
+            "text": "Please provide some feedback",
+            "description": "For example, describe any difficulties you experienced in the use of this service"
+        },
+        "wordCheckLimit": {
+            "limit": 5,
```
rmccar, line 13: It would make sense for this number to be higher to make it more similar to how it would be used

src/components/textarea/_macro-options.md
```diff
@@ -30,6 +31,17 @@
...
 | charCountOverLimitSingular | string  | false    | The string displayed when one more character has been entered after the limit is reached. Set `{x}` in the string to be replaced with the number, for example “You are {x} character over the limit". Defaults to "You have exceeded the character limit by {x} character"     |
 
+## WordCheckLimit
+
+| Name                       | Type    | Required | Description                                                                                                                                                                              |
+| -------------------------- | ------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| id                         | string  | true     | The HTML `id` of the component                                                                                                                                                           |
```
rmccar, line 38:
```suggestion
| id                         | string  | true     | The HTML `id` of the textarea                                                                                                                                                           |
```
Think this should be updated here and in the char one too. Makes it more explicit

src/components/char-check-limit/_macro.spec.js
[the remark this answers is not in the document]
rmccar: Ive added another comment about this, I think you missed my rewording suggestion

src/components/textarea/example-textarea-with-word-limit.njk
SriHV: Increased this to 150 👍

src/components/char-check-limit/_macro.spec.js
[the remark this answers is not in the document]
SriHV: Reworded this similar to other tests for variants 👍

src/components/textarea/_macro-options.md
```diff
@@ -23,13 +24,24 @@
...
 | charCountOverLimitSingular | string  | false    | The string displayed when one more character has been entered after the limit is reached. Set `{x}` in the string to be replaced with the number, for example “You are {x} character over the limit". Defaults to "You have exceeded the character limit by {x} character"     |
 
+## WordCheckLimit
+
+| Name                       | Type    | Required | Description                                                                                                                                                                              |
+| -------------------------- | ------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| id                         | string  | true     | The HTML `id` of the textarea                                                                                                                                                            |
```
rmccar, line 38: The id isn't passed into this object so this can be removed

src/components/textarea/_macro-options.md
```diff
@@ -23,13 +24,24 @@
...
 | -------------------------- | ------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| id                         | string  | true     | The HTML `id` of the component                                                                                                                                                                                                                                                 |
+| id                         | string  | true     | The HTML `id` of the textarea                                                                                                                                                                                                                                                  |
```
rmccar, line 27: The id isn't passed into this object so this can be removed

src/components/input/_macro.spec.js
rmccar, line 1: Line 598 can be removed

src/components/textarea/_macro.spec.js
rmccar, line 1: Line 248 can be removed

src/components/input/_macro.spec.js
SriHV: Screenshot 2025-06-27 at 12 52 19
I am getting this error if I remove the Id

rmccar (approved):

rmccar (approved):

admilne (approved):
LGTM

PR #3585, opened 2025-05-12
Add bar and column with confidence levels

precious-onyenaucheya-ons:

### How to review this PR

### Checklist


janehughes (approved):
Tested locally and displaying as expected 👍

src/components/chart/_macro.njk
```diff
@@ -95,16 +102,20 @@
...
         {% endfor %}
         {%
             set config = {
+                "chart": {
+                    "type": params.chartType,
+                    "inverted": params.isChartInverted if params.isChartInverted else false
+                },
                 "legend": {
-                    "enabled" : params.legend if params.legend else true
+                    "enabled" : params.legend if params.legend else false
```
helenb, line 118: This has a more rigourous update in https://github.com/ONSdigital/design-system/pull/3577

src/components/chart/_macro.njk
```diff
@@ -95,16 +102,20 @@
...
                 }
             %}
             {# Use `set` tag to avoid printing the return value of extend #}
             {% set _ = extend(series, seriesItem) %}
         {% endfor %}
         {%
             set config = {
+                "chart": {
+                    "type": params.chartType,
+                    "inverted": params.isChartInverted if params.isChartInverted else false
```
helenb, line 115: if `isChartInverted` is supplied and set to `false`, does this give the expected value? I found that for the `legend` parameter it unexpectedly would resolve to `true` in that case.

src/components/chart/_macro.spec.js
```diff
@@ -879,6 +881,186 @@ describe('Macro: Chart', () => {
...
+                });
+
+                test('THEN: it includes the Highcharts JSON config', () => {
+                    expect(configScript).toContain('"text":"Years"');
+                    expect(configScript).toContain('"text":"Percentage of GDP"');
+                });
+            });
+        });
+
+        describe('GIVEN: Params: Legend', () => {
```
helenb, line 916: The [other MR](https://github.com/ONSdigital/design-system/pull/3577/files) also adds some more robust testing of the legend, so I don't think this will be needed.

src/components/chart/_macro.njk
helenb: It would also be worth adding tests around this similar to the ones I've added for the legend.

src/components/chart/_macro.spec.js
```diff
@@ -879,6 +881,186 @@ describe('Macro: Chart', () => {
...
+                const $ = cheerio.load(renderComponent('chart', { ...EXAMPLE_BOXPLOT_CHART_PARAMS, legend: false }));
+
+                test('THEN: it disables the legend in config', () => {
+                    const configScript = $(`script[data-highcharts-config--uuid]`).html();
+                    expect(configScript).toContain('"enabled":false');
+                });
+            });
+        });
+
+        describe('GIVEN: Params: Caption', () => {
```
helenb, line 927: I have been wondering if we need to repeat the tests for caption, description etc for each chart type, given that these are the same for every chart. What do you think?

src/components/chart/boxplot.js
```diff
@@ -0,0 +1,66 @@
+import ChartConstants from './chart-constants';
+
+class Boxplot {
+    constructor() {
+        this.constants = ChartConstants.constants();
+    }
+
+    getBoxplotOptions = () => {
+        return {
+            plotOptions: {
```
helenb, line 10: Are we missing setting the options for the column spacing here? It should match the column chart.

src/components/chart/chart.js
```diff
@@ -39,13 +42,18 @@ class HighchartsBaseChart {
         this.yAxisTickIntervalMobile = parseInt(this.node.dataset.highchartsYAxisTickIntervalMobile);
         this.yAxisTickIntervalDesktop = parseInt(this.node.dataset.highchartsYAxisTickIntervalDesktop);
         this.commonChartOptions = new CommonChartOptions(this.xAxisTickIntervalDesktop, this.yAxisTickIntervalDesktop);
+        this.estimateLineLabel = this.node.dataset.highchartsEstimateLineLabel;
```
helenb, line 45: These are optional parameters, so we need to check if these are defined before assigning them

src/components/chart/columnrange-chart.js
```diff
@@ -0,0 +1,94 @@
...
+                    fontSize: this.constants.defaultFontSize,
+                    color: this.constants.legendLabelColor,
+                },
+            },
+        };
+    };
+
+    // This updates the height of the vertical axis and overall chart to fit the number of categories
+    // Note that the vertical axis on a bar chart is the x axis
+    updateColumnRangeChartHeight = (config, currentChart) => {
```
helenb, line 72: Nitpick: this is very similar to the function used on the bar chart. If there is time we could consider moving it to a shared function used by both (but I wouldn't say this is essential now - it could be logged in the tidy up ticket to be done as part of a future clean-up).

helenb (changes requested):
Really nice work @precious-onyenaucheya-ons this is looking great. There are a few things to check, but overall working very well.

src/components/chart/chart.js
```diff
@@ -52,13 +55,18 @@ class HighchartsBaseChart {
             ? parseInt(this.node.dataset.highchartsYAxisTickIntervalDesktop)
             : undefined;
         this.commonChartOptions = new CommonChartOptions(this.xAxisTickIntervalDesktop, this.yAxisTickIntervalDesktop);
+        this.estimateLineLabel = this.node.dataset.highchartsEstimateLineLabel || undefined;
```
helenb, line 58: Nitpick: `this.estimateLineLabel` and `this.uncertaintyRangeLabel` are now inconsistent in how they are declared. I guess if they default to `undefined` anyway we don't actually need the `|| undefined` for the first one.

src/components/chart/chart.js
```diff
@@ -137,6 +153,10 @@ class HighchartsBaseChart {
             // Merge the bar chart options with the existing config
             this.config = this.mergeConfigs(this.config, barChartOptions);
         }
+        if (this.chartType === 'columnrange') {
+            // Merge the bar chart options with the existing config
+            this.config = this.mergeConfigs(this.config, columnRangeChartOptions);
```
helenb, line 158: Heads up that in my reference line pr I moved `mergeConfigs` to a utilities file - we'll just need to make sure this line, and the similar ones below, change to `mergeConfigs` (without the `this.`) when we merge them.

src/components/chart/chart.js
```diff
@@ -149,13 +169,23 @@ class HighchartsBaseChart {
...
+        }
 
         if (this.extraLines > 0) {
             this.config = this.mergeConfigs(this.config, this.lineChart.getLineChartOptions());
             if (this.chartType === 'column') {
                 this.config = this.mergeConfigs(this.config, columnChartOptions);
             }
         }
+        if (this.extraScatter > 0) {
+            //this.config = this.mergeConfigs(this.config, this.scatterChart.getScatterChartOptions());
```
helenb, line 184: Nitpick: remove commented code.

src/components/chart/boxplot.js
```diff
@@ -0,0 +1,90 @@
...
+                symbolRadius: 0,
+                itemStyle: {
+                    fontSize: this.constants.defaultFontSize,
+                    color: this.constants.legendLabelColor,
+                },
+            },
+        };
+    };
+
+    getPointPadding = (config, stackedLayout, numberOfExtraLines, isMobile) => {
```
helenb, line 33: I think that currently we are calling this same function from both  `column-chart.js` and `boxplot.js` - perhaps it was already being set in `column-chart.js` when I added my previous comment about this - in which case sorry for confusion.

src/components/chart/boxplot.js
```diff
@@ -0,0 +1,90 @@
...
+
+        let pointPadding = 0;
+        let groupPadding = 0;
+        let spacing = numberOfCategories >= categoryThreshold ? 0.1 : 0.2;
+
+        // non-clustered charts or stacked charts
+        if (numberOfSeries === 1 || stackedLayout === true) {
+            pointPadding = spacing;
+        }
+        // clustered charts
```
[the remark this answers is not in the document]
helenb: Update to say that this comment may be irrelevant if we remove this and just use the version in `column-chart.js`

src/components/chart/chart.js
```diff
@@ -178,7 +208,13 @@ class HighchartsBaseChart {
                 this.useStackedLayout,
                 this.extraLines,
             );
-            mobileChartOptions = this.mergeConfigs(mobileChartOptions, mobileColumnChartOptions);
+            const mobileBoxplotChartOptions = this.columnChart.getColumnChartMobileOptions(
```
helenb, line 211: fyi This is where the `getPointPadding` frpm the column chart is effectively being pulled in.

helenb (commented):
Thanks @precious-onyenaucheya-ons - a few more bits and pieces to check but nearly there.

src/components/chart/_macro.njk
helenb: If we assume the default for `isChartInverted` is `false` then this works as expected - the issue with the legend parameter was it defaulted to `true` which made the logic needed to test was more complex. I assume it is ok for it to default to `false` - can you add it to the `macro-options.md` file?

src/components/chart/_macro.spec.js
helenb: It looks like you've removed this now so I'll resolve.

src/components/chart/boxplot.js
```diff
@@ -0,0 +1,37 @@
...
+                    lineWidth: 0,
+                    medianColor: this.constants.estimateLineColor,
+                    medianDashStyle: 'Solid',
+                    medianWidth: 3,
+                    stemWidth: 0,
+                    whiskerWidth: 0,
+                },
+            },
+            legend: {
+                enabled: false,
```
helenb, line 26: Maybe add a comment here to explain why it is set to false by default

helenb:
I noticed that we can only pass the custom labels to the legend for the column (boxplot) confidence interval chart at the moment. I think we might also need to be able to do that with the bar (column-range) version as well. Would you mind checking with Erin about that?

src/components/chart/boxplot.js
helenb: The `column-chart.js` `getPointPadding` function is still getting called twice currently. It's called where you've added it in `boxplot.js` on line 14, but also called where I flagged it in my previous comment - line 211 of `chart.js`

src/components/chart/chart.js
helenb: Discussed with Precious and this one is for getting the mobile point padding so needs to remain. It gets called at desktop too, but we're getting the mobile values. Probably would need to refactor how this works in order to change it, so leaving for now.

helenb:
> Could you also update the `type` parameter for `series` in the macro-options file, as we now support `scatter` too.

Just fyi I also updated it in https://github.com/ONSdigital/design-system/pull/3601/files so we'll need to merge those carefully.

helenb (approved):
Discussed the fact that we aren't allowing for custom legend labels for the bar chart with confidence intervals, as this can be handled via the series name.

PR #3218, opened 2024-05-24
Add formatting for Nunjucks files

rmccar:
### What is the context of this PR?

Fixes: ONSdigital/design-team#58 

Adds the Jinja plugin (`prettier-plugin-jinja-template`) to prettier and have then run prettier on all our Nunjucks files to format them correctly. I have also added them to the lint staging step so that they are formatted on the Husky pre commit hook. I have also fixed all errors and warnings from running prettier which has required some refactoring.

Some other fixes have gone in too:
- Removed some component imports in some examples because those components weren't being used.
- Removed the setting of the `variants` param on the related content component in some examples. This was having no effect because it's hardcoded to "bare" in the macro.

### How to review this PR

- Tests pass
- Check through all files checking no unexpected changes have happened, this is the list of components that needed refactoring:
	- address-input
	- autosuggest
	- button
	- card
	- details
	- document-list
	- feedback
	- field
	- footer
	- header
	- input
	- label
	- language-selector
	- list
	- panel
	- section-navigation
	- share-page
	- summary
	- table
	- tabs
- Check no breaking changes have been introduced

### Checklist

This needs to be completed by the person raising the PR.

- [x] I have selected the correct Assignee
- [x] I have linked the correct Issue

PR #3594, opened 2025-05-16
Enhancement/add new table component

SriHV:

[ONSDESYS-367](https://jira.ons.gov.uk/browse/ONSDESYS-367).

Added new variant `merge-cells` which can merge rows and columns using `colspan` and `rowspan`,
Added an option to add multiple headers, option to make any row a header
Added tests to support this

### How to review this PR

Check that new example `example-table-merge-cells` follows the acceptance criteria mentioned in [ONSDESYS-367](https://jira.ons.gov.uk/browse/ONSDESYS-367).

### Checklist


backstop_data/bitmaps_reference/ds-vr-test__components_table_example-table-merge-cells_0_document_2_mobile.png
rmccar, line 1: The spacing between the first and second column here doesnt look right, the dividing lines look too close to the text in the first column

rmccar:
I think we need to update the title and description to be a bit more descriptive, this isnt adding a new vairant of the table component this is adding the ability to merge columns and rows using colspan and rowspan

src/components/table/_macro.spec.js
```diff
@@ -95,6 +95,36 @@ describe('macro: table', () => {
...
+                            value: 'Column 1',
+                            colspan: 2,
+                        },
+                    ],
+                }),
+            );
+
+            expect($('.ons-table__header').attr('colspan')).toBe('2');
+        });
+        it('adds additional colspan attribute to column header', () => {
```
rmccar, line 113:
```suggestion
        it('adds additional rowspan attribute to column header', () => {
```

src/components/table/_macro.spec.js
```diff
@@ -291,6 +343,44 @@ describe('macro: table', () => {
...
+                                },
+                            ],
+                        },
+                    ],
+                }),
+            );
+
+            expect($('.ons-table__cell').attr('colspan')).toBe('2');
+        });
+        it('adds additional colspan attribute to column header', () => {
```
rmccar, line 365:
```suggestion
        it('adds additional rowspan attribute to row header', () => {
```

src/components/table/_macro.spec.js
```diff
@@ -291,6 +343,44 @@ describe('macro: table', () => {
 
             expect($('.ons-table__cell').hasClass('ons-table__cell--numeric')).toBe(true);
         });
+        it('adds additional colspan attribute to column header', () => {
```
rmccar, line 346:
```suggestion
        it('adds additional colspan attribute to row header', () => {
```

src/components/table/example-table-merge-cells.njk
```diff
@@ -0,0 +1,155 @@
+{% from "components/table/_macro.njk" import onsTable %}
+{{
+    onsTable({
+        "caption": "Table with cells merged, multiple row headers and column headers",
+        "variants": "merge-cells",
```
rmccar, line 5: I also wouldn't see this as a variant. This PR enhances the table component to add the ability to merge cells. A variant would be a table that is different to the default table in some way, such as being compact so that it is styled so that the cells are closer together. I think it could be confusing to have to set this as well as the colspan/rowspan params for the user

src/components/table/_table.scss
```diff
@@ -13,10 +13,15 @@
         text-align: left;
     }
 
+    .ons-table__row:not(.ons-table__row--after-rowspan) {
```
rmccar, line 16:
```suggestion
    &__row:not(.ons-table__row--after-rowspan) {
```

src/components/table/_macro-options.md
```diff
@@ -6,12 +6,13 @@
...
 | ariaAsc      | string          | false (unless “sortable” variant set) | Sets the `data-aria-asc` attribute for the table. Used to update `aria-sort` attribute to announce to screen readers how a table is sorted by a column, for example, "Sort by Date, **ascending**".      |
 | ariaDesc     | string          | false (unless “sortable” variant set) | Sets the `data-aria-desc` attribute for the table. Used to update `aria-sort` attribute to announce to screen readers how a table is sorted by a column, for example, "Sort by Date, **descending**".    |
 | tfoot        | Array`<tfoot>`  | false                                 | An array of `td` elements for a `tfoot` [footer element](#tfoot)                                                                                                                                         |
+| thGroups     | Array`<ths>`    | true (unless "ths" is set)            | An array of `ths` [multiple header cell elements](#ths) for table                                                                                                                                        |
```
rmccar, line 15: Do we have this param anymore?

src/components/table/_macro-options.md
```diff
@@ -6,12 +6,13 @@
...
 | hideCaption  | boolean         | false                                 | Set to “true” to visually hide the caption                                                                                                                                                               |
 | ariaLabel    | string          | false                                 | The ARIA label to be added if ”scrollable” variant set, to inform screen reader users that the table can be scrolled. Defaults to “Scrollable table“.                                                    |
-| ths          | Array`<th>`     | true                                  | An array of `th` [header cell elements](#th) for table                                                                                                                                                   |
+| ths          | Array`<th>`     | true (unless "thGroups" is set)       | An array of `th` [header cell elements](#th) for table                                                                                                                                                   |
```
rmccar, line 9: Do we have `thGroups` anymore?

src/components/table/_macro.njk
```diff
@@ -17,55 +18,111 @@
...
+                        class="ons-table__row{{ ' ons-table__row--highlight' if tr.highlight }}{{ ' ons-table__row--after-rowspan' if previousRowHasRowspan }}"
+                        {% if tr.id %}id="{{ tr.id }}"{% endif %}
+                    >
                         {% for td in tr.tds %}
-                            <td
-                                class="ons-table__cell{{ ' ' + td.tdClasses if td.tdClasses else '' }}{{ ' ons-table__cell--numeric' if td.numeric }}"
-                                {% if td.id %}id="{{ td.id }}"{% endif %}
-                                {% if td.data %}{{ ' ' }}data-th="{{ td.data }}"{% endif %}
-                                {% if td.dataSort %}{{ ' ' }}data-sort-value="{{ td.dataSort }}"{% endif %}
+                            {% if td.heading %}
```
rmccar, line 91:
```suggestion
                            {% if td.heading == true %}
```
At the moment this defaults to true if something other than a boolean is passed in. I think it should probably default to false. This change will mean it does that

src/components/table/_macro.njk
```diff
@@ -17,55 +18,111 @@
...
+                                {% endif %}
+                            </th>
+                        {% endfor %}
+                    </tr>
+                {% endif %}
             </thead>
             <tbody class="ons-table__body">
                 {% for tr in params.trs %}
-                    <tr class="ons-table__row{{ ' ons-table__row--highlight' if tr.highlight }}" {% if tr.id %}id="{{ tr.id }}"{% endif %}>
+                    {% set trIndex = loop.index0 %}
```
rmccar, line 75: This value doesn't seem to be used

src/components/table/_macro.njk
```diff
@@ -2,6 +2,7 @@
     {% from "components/button/_macro.njk" import onsButton %}
     {% from "components/icon/_macro.njk" import onsIcon %}
     {% set variants = params.variants if params.variants %}
+    {% set rowspanRemaining = 0 %}
```
rmccar, line 5: It would make sense for this to be set nearer to where it is used

src/components/table/_macro.njk
```diff
@@ -17,55 +18,111 @@
...
+                                    <span class="ons-table__header-text">{{- th.value | safe -}}</span>
+                                    {% if 'sortable' in variants %}
+                                        {{-
+                                            onsIcon({
+                                                "iconType": "sort-sprite",
+                                                "id": th.value | replace(' ', '-'),
+                                                "classes": 'ons-u-d-no'
+                                            })
+                                        -}}
+                                    {% endif %}
```
rmccar, line 42: This seems to be duplicated, you could restructure this logic so that it only needs to be used once

src/components/table/_macro.njk
```diff
@@ -17,55 +18,111 @@
...
+                    {% if rowspanRemaining > 0 %}
+                        {% set previousRowHasRowspan = true %}
+                        {% set rowspanRemaining = rowspanRemaining - 1 %}
+                    {% else %}
+                        {% set previousRowHasRowspan = false %}
+                        {% set currentRowspan = tr.tds[0].rowspan | default(1) %}
+                        {% if currentRowspan > 1 %}
+                            {% set rowspanRemaining = currentRowspan - 1 %}
+                        {% endif %}
+                    {% endif %}
```
rmccar, line 85: This logic is a bit confusing and convoluted just to be able to set `ons-table__row--after-rowspan` you should be able to do this just using css and not need to set that class

rmccar (changes requested):

src/components/table/example-table-merge-cells.njk
SriHV: The main reason I added is we set `border:0` for the last item in our tables. All the examples linked have border for the last item. 
image

You are right name is not right. Shall call it `add-borders` so we can add border to our tables?

rmccar: No I don't think adding a variant is the right idea, the user should be able to just add colspan/rowspan params and the rest is done for them. This isn't a variant of the table I dont think its just adding some new functionality. We should be able to do this another way, probably with just css

src/components/table/_macro.njk
SriHV: removed it

src/components/table/_macro.njk
SriHV: I am sure how to do this in css. I have tired using parent and sibling classes it didn't work.

We want first cell in every row to set with padding left 0 and rest of them with padding.

What I tired  to do is to detect when a first cell of table row has a rowspan, and remember how many rows that span covers. While we’re inside that span, we mark each following row with a class so we know it comes after a rowspan. We are removing padding left:0 to these classes because visually it is shown as second cell.

Open to suggestions here :)

src/components/table/example-table-merge-cells.njk
```diff
@@ -0,0 +1,154 @@
...
+                    },
+                    {
+                        "value": "1.2"
+                    }
+                ]
+            },
+            {
+                "tds": [
+                    {
+                        "heading":true,
```
rmccar, line 82:
```suggestion
                        "heading": true,
```

src/components/table/example-table-merge-cells.njk
```diff
@@ -0,0 +1,154 @@
...
+                        "value": "%Change"
+                    },
+                    {
+                        "value": "3.5"
+                    },
+                    {
+                        "value": "-1.8"
+                    },
+                    {
+                        "value": ""
```
rmccar, line 148: I realise this might be tricky to solve but it would be good if we didn't have to set empty values here

src/components/table/_macro.njk
```diff
@@ -16,60 +32,97 @@
...
-                            {% if 'sortable' in variants %}aria-sort="{{- th.ariaSort | default('none') -}}"{% endif %}{% if th.widthPercentage %}width="{{ th.widthPercentage }}%"{% endif %}
-                        >
-                            <span class="ons-table__header-text">{{- th.value -}}</span>
-                            {% if 'sortable' in variants %}
-                                {{-
-                                    onsIcon({
+                {% for header_cols in thGroups %}
+                        <tr class="ons-table__row">
+                            {% for th in header_cols.ths %}
+                                {% set thHtml %}
```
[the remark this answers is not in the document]
SriHV: Removed this

src/components/table/_macro.njk
SriHV: I noticed that Flexbox is generally used with <div>s rather than <table> elements. Do you think we should switch our structure to use Flex and its properties?

src/components/table/_macro.njk
```diff
@@ -2,11 +2,27 @@
     {% from "components/button/_macro.njk" import onsButton %}
     {% from "components/icon/_macro.njk" import onsIcon %}
     {% set variants = params.variants if params.variants %}
+    {% set hasrowSpan = false %}
```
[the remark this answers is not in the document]
rmccar: It should be `hasRowSpan`

src/components/table/_macro.njk
```diff
@@ -16,60 +32,96 @@
...
+                                </form>
+                            {% endif %}
+                            {% if td.value %}
+                                {{ td.value | safe }}
+                            {% endif %}
+                            {{ closingTag | safe }}
                         {% endfor %}
                     </tr>
                 {% endfor %}
+
```
rmccar, line 122:
```suggestion
```

src/components/table/_macro.njk
```diff
@@ -16,60 +32,96 @@
                     {{ params.caption }}
                 </caption>
             {% endif %}
+            <!-- prettier-ignore-start -->
+            {# Uses thList if multiple-row headers, otherwise wrap single ths array into a default group #}
+            {% set thGroups = params.thList if params.thList else [ { ths: params.ths } ] %}
```
rmccar, line 37:
```suggestion
            {% set thGroups = params.thList if params.thList else [{ ths: params.ths }] %}
```

src/components/table/_macro.njk
```diff
@@ -16,60 +32,96 @@
...
                                         "iconType": "sort-sprite",
                                         "id": th.value | replace(' ', '-'),
                                         "classes": 'ons-u-d-no'
-                                    })
-                                -}}
-                            {% endif %}
-                        </th>
-                    {% endfor %}
-                </tr>
+                                        })
```
rmccar, line 57: This needs formatting

src/components/table/_macro.njk
```diff
@@ -16,60 +32,97 @@
                     {{ params.caption }}
                 </caption>
             {% endif %}
+            <!-- prettier-ignore-start -->
```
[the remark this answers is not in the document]
rmccar: This should be fixed rather than ignored and may need a bit of a refactor to get it to pass. I dont think its that line it has issue with. I removed that line and it still errored. It will be something to do with the way its written and just need to move a few things around. In this [PR](https://github.com/ONSdigital/design-system/pull/3218/files) here I did a lot of that to get around these issues and avoid adding ignores

src/components/table/_macro.njk
SriHV: To keep things simple, I set padding-left: 0 for all the cells and checked this with Joe - we don't need this code anymore.

src/components/table/_macro.njk
```diff
@@ -17,55 +33,77 @@
...
             <tbody class="ons-table__body">
                 {% for tr in params.trs %}
                     <tr class="ons-table__row{{ ' ons-table__row--highlight' if tr.highlight }}" {% if tr.id %}id="{{ tr.id }}"{% endif %}>
                         {% for td in tr.tds %}
-                            <td
-                                class="ons-table__cell{{ ' ' + td.tdClasses if td.tdClasses else '' }}{{ ' ons-table__cell--numeric' if td.numeric }}"
-                                {% if td.id %}id="{{ td.id }}"{% endif %}
-                                {% if td.data %}{{ ' ' }}data-th="{{ td.data }}"{% endif %}
-                                {% if td.dataSort %}{{ ' ' }}data-sort-value="{{ td.dataSort }}"{% endif %}
+                            {% if td.heading == true %}
```
janehughes, line 72: We need to add scope="row" for any th cells

src/components/table/example-table-merge-cells.njk
```diff
@@ -0,0 +1,154 @@
...
+                    {
+                        "value": "Heading B",
+                        "colspan": 4
+                    }
+                ]
+            },
+            {
+                "ths": [
+                    {
+                        "value": ""
```
janehughes, line 21: We had feedback in the DAC report that table headings should not be left empty, as it's confusing for screen reader users without context. They should have some visually hidden text that describes their purpose (a relevant sub heading under "Heading A" and "Heading B"). It would help to use real heading examples here (instead of Heading A/B) so that it's clear what is expected when using them

janehughes: Yes perfect thanks @SriHV. We need to make sure we provide editorial guidance on this to make sure no empty heading cells get added.

rmccar:
Screenshot 2025-06-23 at 11 34 07
For smaller screens there is no horizontal scroll so the data moves off the page and is no longer readable. This might be a requirement

Other than this looks good

rmccar (approved):

Issue #3581, opened 2025-05-09
[Bug]: Accessibility - "title" attribute is not valid on <svg> element


janehughes:
### Describe the bug

Flagged as a validation issue via https://validator.w3.org/ after testing the ONS Beta site.

Several SVG elements are using an inline `title` attribute which is invalid HTML and will not be read by assistive tech.

### Expected behaviour / outcome

The title attribute should be replaced with `<title>` inside the <svg> parent element

### List the steps necessary to reproduce the bug

Inspect any SVGs in the design system codebase and see that they have a `title=""` attribute

### What version of the ONS Design System are you using?

72.9.1

### What device(s) are you using?

Desktop

### What operating system(s) are you using?

macOS

### What browser(s) are you using?

Chrome

MagdalenaLarge:
Moved to Jira: https://jira.ons.gov.uk/browse/ONSDESYS-465

PR #3596, opened 2025-05-16
[Fix] Remove title attributes from SVGs

janehughes:

Ticket: https://jira.ons.gov.uk/browse/ONSDESYS-465
Related issue: https://github.com/ONSdigital/design-system/issues/3581

This PR removes title attributes from all SVG icons as they are invalid HTML and ignored by the browser. SVGs must use the `<title>` child element instead.

### How to review this PR

- Review SVG icons and see that all redundant title attributes have been removed.
- Check SVG includes on components and inspect element to check the attribute has been removed successfully.

### Checklist


janehughes:
> Do we not want to replace these with a child title element?

No they're all decorative icons so don't need a title and will have `aria-hidden` applied as well through the changes in https://github.com/ONSdigital/design-system/pull/3592

Any non-decorative svgs that do need a title are already using `<title>` 👍

rmccar (approved):

SriHV (approved):

PR #3597, opened 2025-05-16
Fix/onsdesys 413/textarea character limit calculated incorrectly

admilne:

Ticket: [ONSDESYS-413](https://jira.ons.gov.uk/browse/ONSDESYS-413)

- Sets up default text for exceeding the character limit to prevent a breaking change.
- Move away from using `input.maxLength` to using `input.getAttribute('data-maxlength')` so the browser won't stop a user over typing.
- Update documentation to inform user of ability to override default texts for going over text limits.
- Updates tests to take into account that new lines now count as two characters, not just one.

### How to review this PR

- Check the textarea examples works without error.
- Check the textarea example with character limit updates the character count correctly for single and plural examples. This should accurately reflect when you have characters remaining and when you are over the limit.

### Checklist


admilne:
Closing this as the approach was wrong. Re-done in this PR - https://github.com/ONSdigital/design-system/pull/3604

PR #3599, opened 2025-05-19
Feature/reference line annotations

helenb:

- Implements reference line annotations - see https://jira.ons.gov.uk/browse/CCB-78
- Introduces a `utilities` file in order to avoid the `chart.js` file becoming too bloated

### How to review this PR

Check out the merge request and review the new examples.

### Checklist


src/components/chart/chart.js
```diff
@@ -83,40 +91,6 @@ class HighchartsBaseChart {
         Highcharts.setOptions(chartOptions);
     };
 
-    // Utility function to merge two configs together
-    mergeConfigs = (baseConfig, newConfig) => {
```
helenb, line 87: This has been moved to a utilities file now. We just need to be a bit careful when this is merged to ensure any other code that is using `this.mergeConfigs` gets updated.

src/components/chart/_macro-options.md
```diff
@@ -1,25 +1,24 @@
...
-| xAxis                   | object  | true     | Defines the horizontal axis [x-axis](#x_axis) configuration parameters                                                                                                                       |
-| series                  | array   | true     | The data [series](#series) to be plotted, including labels and values.                                                                                                                       |
-| useStackedLayout        | boolean | false    | Determines whether the chart should use a stacked layout. It is useful only for bar and column charts.                                                                                       |
-| annotations             | array   | false    | An array of annotations.                                                                                                                                                                     |
```
helenb, line 17: This update removes a duplicate entry for `annotations` which probably happened in a merge at some point.

src/components/chart/_macro-options.md
rmccar, line 1: I think the way the objects have been documented here needs updating to be more inline with the style of the rest of the components in the DS. When an object is documented in its own table, there is a reference to it after the type value in the main table. Have a look at some of the other components. Happy to have a chat about it if any of that doesn't make sense

rmccar: This could be broken out into a ticket in itself though, other than this looks good

helenb: Re the object documentation, I've added a note to https://jira.ons.gov.uk/browse/CCB-87

PR #3604, opened 2025-05-23
Fix/onsdesys 413/textarea character limit calculated incorrectly

admilne:

Update character count for textareas to allow for browsers creating two characters for new lines, not just one. To prevent issues with javascript and to allow users to be responsible for trimming their own text down we've changed from using a character limit to a character check - informing the user when they are over the character limit rather than prevent them from typing anything else. 

Ticket: [ONSDESYS-413](https://jira.ons.gov.uk/browse/ONSDESYS-413). This [replaces this PR](https://github.com/ONSdigital/design-system/pull/3597) which took the wrong approach.

- Update visual regression tests as examples are now called character-check rather than character-limit.
- Sets up default text for exceeding the character limit to prevent a breaking change.
- Updates the CharCheck class so it can handle being passed an input as well as a containing div.
- Creates the `getCharLength` method so we can calculate the length of text knowing that a new line should count as two characters.
- Updates the `setShowMessage` method so textareas will always show the character count. This is so it maintains the functionality it had when it used limit characters.
- Removes `textarea.dom.js` and associated tests as these are no longer required.
- Removes `character-limit.js` as it's no longer used in the codebase

### How to review this PR

- Check the textarea examples works without error.
- Check the textarea example with character limit updates the character count correctly for single and plural examples. This should accurately reflect when you have characters remaining and when you are over the limit.

### Checklist


src/components/char-check-limit/character-check.spec.js
```diff
@@ -190,11 +190,6 @@ describe('script: character-check', () => {
                 await page.focus('#feedback-checkbox');
                 await page.keyboard.press('Space');
             });
-
-            it('then aria-live attribute should removed', async () => {
-                const hasAriaLiveAttribute = await page.$eval('#feedback-lim', (element) => element.hasAttribute('aria-live'));
-                expect(hasAriaLiveAttribute).toBe(false);
-            });
```
rmccar, line 197: Just wondering why this was removed, I cant see any other changes with the aria-live attribute? I saw your explanation in the description but couldn't get why this doesn't always fail before and after this change?

src/components/char-check-limit/_macro.njk
```diff
@@ -8,8 +8,8 @@
         class="ons-input__limit ons-u-fs-s--b ons-u-d-no ons-u-mt-2xs"
         data-charcount-singular="{{ params.charCountSingular }}"
         data-charcount-plural="{{ params.charCountPlural }}"
-        data-charcount-limit-singular="{{ params.charCountOverLimitSingular }}"
-        data-charcount-limit-plural="{{ params.charCountOverLimitPlural }}"
+        data-charcount-limit-singular="{{ params.charCountOverLimitSingular | default("You have exceeded the character limit by {x} character") }}"
+        data-charcount-limit-plural="{{ params.charCountOverLimitPlural | default("You have exceeded the character limit by {x} characters") }}"
```
rmccar, line 12: Although adding this would prevent a breaking change, it does mean it could be breaking in that the language used on the page could be not English, which would mean then that this wouldn't fix with the page language

src/components/char-check-limit/character-check.js
```diff
@@ -6,9 +6,18 @@ const attrCharCheckVal = 'data-char-check-num';
...
+        // Handle either an input directly or a container with an input inside
+        if (this.tagName.toLowerCase() === 'input' || this.tagName.toLowerCase() === 'textarea') {
+            this.input = context;
+        } else {
+            this.input = context.querySelector('input');
+        }
+
+        // Find the button: if input is passed directly, look at its parent
+        let parent = this.input.parentNode;
+        this.button = parent ? parent.querySelector('button') : null;
```
rmccar, line 20: Can see youve added this so that it will deal with inputs that have containers, why was this needed? For textareas which havent been used with char check before?

rmccar (commented):
Looks good just had a few questions. Also the description on this PR was great, had lots of great information about the changes and why on there!

src/components/char-check-limit/character-check.spec.js
admilne: @rmccar when we were using **limit** before, it did remove the attribute but switching it to **check** from **limit** that attribute is no longer removed. I couldn't actually find the piece of code that removed the attribute for limit but since it doesn't seem to make a difference to screen readers I question whether we even need this test.

src/components/char-check-limit/_macro.njk
admilne: @rmccar given there wasn't a default before, it meant that you would have had to supply your own value and since you are still able to provide your own value I don't think this is a breaking change?

src/components/char-check-limit/character-check.js
admilne: @rmccar this was a bit tricky to wrap my head around but it seems an "input" will get wrapped in a span with a class of `ons-js-char-check-input` whereas the textarea gets rendered without the span wrapping it so the class is applied directly to it. I've attached screenshots to explain further.

Character limit and character check just worked slightly differently in what they were expecting to be passed through.

Screenshot 2025-05-23 at 17 19 53

Screenshot 2025-05-23 at 17 21 45

rmccar: I cant think of a reason off the top of my head why they work slightly differently but yeah ok thats fine

src/components/char-check-limit/character-check.spec.js
rmccar: This is used for mutually exclusive options where when a checkbox like "prefer not to say" is checked it will clear the checkbox and the attribute is removed.

This is the line where it gets removed https://github.com/ONSdigital/design-system/blob/5362fda111f22c2c3641c926ca4d0d020526821a/src/components/char-check-limit/character-limit.js#L30

I think we should leave this test in.

admilne: @rmccar I figured out what was different about this compared with the character limit. There was some duplicate code which ended up overriding the removal of the attribute so I've fixed that and added this test back in.

admilne:
> LGTM. Just a small thing - should we remove `character-limit.js`, I think it's not used anywhere now

@rmccar thoughts on this? I don't think it is being used anymore after my changes.

rmccar:
> > LGTM. Just a small thing - should we remove `character-limit.js`, I think it's not used anywhere now
> 
> @rmccar any thoughts against this? I don't think it is being used anymore after my changes so I think I'd be inclined to agree.

Yeah happy to remove if its not being used

PR #3612, opened 2025-05-30
Don't set a width on the range annotation label when it is set to be 'inside'

helenb:

See https://jira.ons.gov.uk/browse/CCB-130
We should not set an explicit width on range annotation labels when they are set to be 'inside'. Instead the text should flow inside the grey area.

### How to review this PR

Test on your local build with wider and narrower versions of the range annotation used on http://localhost:3000/components/chart/example-line-chart-with-range-annotations-inside.html

The screenshots below show that chart with the width of the range extended to a value of 100.

<details>
<summary>Expand for screenshots</summary>

Before (fixed width label):

Screenshot 2025-05-30 at 11 10 19

After (width not set):

Screenshot 2025-05-30 at 11 11 00

</details>

### Checklist


src/components/chart/range-annotations-options.js
```diff
@@ -24,7 +28,8 @@ class RangeAnnotationsOptions {
                     style: {
                         color: this.constants.labelColor,
                         fontSize: this.constants.defaultFontSize,
-                        width: rangeAnnotation.labelWidth ? rangeAnnotation.labelWidth : 150,
+                        // this property is ignored if rangeAnnotation.labelInside is undefined
```
nimasmi, line 31: Should this read "this property is undefined if…"?

helenb: No, it's ignored by Highcharts, and not set as an inline style, if undefined. I can make that clearer.

precious-onyenaucheya-ons (approved):

PR #3615, opened 2025-06-02
custom reference line feature

helenb:

See https://jira.ons.gov.uk/browse/CCB-128

This feature adds the ability to customise the value for the darker reference line that appears at zero (on the y axis) by default.

- The default value remains zero
- Adding a customReferenceLineValue allows you to change the y axis value for the darker reference line
- But only for column and line charts - for other charts the customReferenceLineValue will be ignored, and the line will remain at zero.

### How to review this PR

- Test the new examples in your local build
- Test adding a customReferenceLine on various chart types - ensure it only gets applied for column charts (normal, clustered or stacked) and line charts

### Checklist


janehughes (approved):
Working as expected 👍

PR #3618, opened 2025-06-04
Add flex grid gap modifier classes

janehughes:

Ticket: https://jira.ons.gov.uk/browse/ONSDESYS-525

This adds some custom grid gap modifier classes to allow us to match the grid gaps provided in the new website Figma designs, so we can maintain consistent spacing between column and grid elements in the page templates without requiring any custom styles.

I've also added one tiny styling update which was spotted while testing - adding a transparent outline to featured cards so they are more visually distinctive in high contrast mode

### How to review this PR

- Test the headline figures example card component to see the new grid modifiers in action - the columns should have consistent sizing and display correctly
- Also check in high contrast mode to see the border is showing on the headline figures cards

### Checklist


src/scss/utilities/_grid.scss
rmccar, line 1: Do we want to add a `&--m` option in here for 1.25rem? Feels weird to not offer it.

src/components/card/example-card-set-with-headline-figures.njk
```diff
@@ -1,8 +1,8 @@
 {% from "components/card/_macro.njk" import onsCard %}
 
 <div class="ons-container">
-    <div class="ons-grid ons-grid-flex ons-grid-flex--vertical-top">
-        <div class="ons-grid__col ons-grid__col--flex-col ons-grid__col--stretch ons-u-flex-grow ons-col-4@m">
+    <div class="ons-grid ons-grid-flex-gap ons-grid-flex-gap--24 ons-grid-flex--vertical-top">
```
rmccar, line 4: Wouldn't we use one of the sizes here? Like `ons-grid-flex-gap--l`

src/scss/utilities/_grid.scss
janehughes: Yeah agreed, have added now 👍

SriHV (approved):

PR #3619, opened 2025-06-06
Fix to reference line annotations to ensure they display for x axis only

helenb:

See https://jira.ons.gov.uk/browse/CCB-137.

This PR addresses a couple of bugs reported to me by @nimasmi  when working on the integration of reference line annotations into the chart builder.

1) The macro spec specifies `axisValue`, but actually the code, and the examples, all use value instead.
2) The vertical X axis line annotation appears only if both annotations are there:
- X and Y – works fine
- Y alone – works fine
- X alone – no display

This was due to a bug in the code where we combine reference line annotations, and the zero line / custom reference line configs, as Highcharts uses plotLines for both under the hood.

### How to review this PR

- Check the macro spec is now correct
- Check the example line chart with range annotations and adjust the values to test with annotations on both axes, annotations on just the y axis, and annotations on just the x axis. Also try adding and removing a custom reference line value on the y axis. Make sure the reference line annotations, and the reference line values, always display.

### Checklist

Ticking the box as otherwise the tests fail, but there is no explicit ticket for this issue.


src/components/chart/utilities.js
```diff
@@ -81,8 +81,10 @@ export const preparePlotLinesAndBands = (
         desktopAllPlotLines.yAxis.plotLines = desktopMergedPlotLines;
         mobileAllPlotLines.yAxis.plotLines = mobileMergedPlotLines;
     } else {
-        desktopAllPlotLines = { ...plotLineOptions };
-        mobileAllPlotLines = { ...plotLineOptions };
+        desktopAllPlotLines = { ...desktopReferenceLineAnnotations };
+        mobileAllPlotLines = { ...mobileReferenceLineAnnotations };
+        desktopAllPlotLines.yAxis = { ...plotLineOptions.yAxis };
+        mobileAllPlotLines.yAxis = { ...plotLineOptions.yAxis };
```
nimasmi, line 87: **Suggestion**: There's duplication of

```javascript
        desktopAllPlotLines = { ...desktopReferenceLineAnnotations };
        mobileAllPlotLines = { ...mobileReferenceLineAnnotations };
```

inside the two branches of the if statement. Is it safe to put those beforehand?

helenb: That's now updated.

helenb:
> Minun mielestäni se näyttää hyvältä.

Kiitos

Issue #3606, opened 2025-05-23
Header menu styling & alignment


janehughes:
Following integration & testing on the new wesbite, we've found a few small styling updates required for the header menu:

- Need a little more padding at the bottom 
- Spacing of the columns needs adjusting
- Language switcher needs moving after the menu toggle button

Screenshot from new website:

<img width="1077" alt="Image" src="https://github.com/user-attachments/assets/5750fc37-9a0b-4ecb-90af-c4fe85cbffc1" />

PR #3620, opened 2025-06-06
Basic header styling fixes

rmccar:

Fixes: #3606 [ONSDESYS-520](https://jira.ons.gov.uk/browse/ONSDESYS-520)

This PR adds a few different fixes to the basic header variant:
- adds padding to the bottom of the menu to match the top
- moves the language toggle to the right of the menu button
- updates the basic header example to add language links

Also adds:
- removes dividing line for non JS variant when only the menu or search is used but not both
- fixes some JS console errors
- fixes the focus state styling on the search button when mouse button is down to make it behave like the menu button

### How to review this PR

- check that the menu now looks as expected as outlined in the ticket (menu and language toggles in the right place and has the right padding)
- test the menu on the main branch with the console open to see there are errors, then test on this branch to see that the errors are gone
- test the main branch with an example with only the menu or search, turn JS off and see the dividing line appear, do the same on this branch and see that it is now gone. Also check for when using both it does appear. 
- compare the main branch with this one using the `example-header-with-search-button` example and check that on the main branch the search button doesn't add the focus styling on mouse down but on this branch it does.

### Checklist


janehughes (approved):
This is looking good :)

admilne (approved):
@rmccar this still looks good to me

PR #3623, opened 2025-06-09
Addresses an issue on resizing charts, a bug with the legend and moves some functions to the correct options file

helenb:

See https://jira.ons.gov.uk/browse/CCB-133

This PR addresses 2 bugs:

- An issue when charts are rendered in the 'preview' pane in Wagtail whereby the chart aspect ratio does not update to reflect the changed screen size. This was happening due to calling the the `adjustChartHeight` function from a resize event rather than the Highcharts render event.
- An issue with the custom legend symbol - when the line chart is resized down to mobile and then up again, the symbol was reverting to a line. This change ensures it runs on the render event instead of the load event, so it will always update (and any duplicate symbols are removed).

It also addresses any potential bugs with the fact that a few chart specific functions were being called from `common-chart-options` (which runs once for all charts on a page, and sets global options) instead of `specific-chart-options` (which runs once for each chart on the page). These have been moved across.

This PR may also end up resolving some of the VR test issues with height variation between charts, partly due to moving the `adjustChartHeight` function, but also due to adding rounding to the resize calculation.

The rounding is also resulting in VR test failures for quite a few chart examples, but this would be expected. The VR test differences also show around the legend for line charts - because I have updated the way the symbol is generated slightly - it is close to it's previous position but not completely dead on.

### How to review this PR

These changes should not change anything visually with the resulting charts. As noted above there are minor differences coming out of the VR testing due to the rounding of the chart height calculations.

We can't test that it definitely fixes the issue in wagtail preview until it is merged, and a new DS version released.

### Checklist


src/components/chart/specific-chart-options.js
```diff
@@ -18,6 +18,49 @@ class SpecificChartOptions {
...
+
+    adjustChartHeight = (currentChart, percentageHeightDesktop, percentageHeightMobile) => {
+        // get height and width of the plot area
+        const plotHeight = currentChart.plotHeight;
+        const plotWidth = currentChart.plotWidth;
+        // calculate the new plot height based on the percentage height
+        // default to the current height
+        let newPlotHeight = plotHeight;
+        if (plotWidth > 400) {
+            newPlotHeight = Math.round(plotWidth * (percentageHeightDesktop / 100));
```
helenb, line 52: Math.round added here to potentially help with VR issues.

src/components/chart/specific-chart-options.js
```diff
@@ -38,6 +81,80 @@ class SpecificChartOptions {
...
+                    // and if so, it updates the legend to display as a symbol rather than as a line
+                    // We only to this for chart types that are explicitly line charts - i.e. not column with line
+                    if (!symbol && label && label.element && chart.userOptions.chart.type === 'line') {
+                        // Hide the line in the legend
+                        if (line) {
+                            line.hide();
+                        }
+
+                        // Remove any existing custom symbols for this legend item
+                        const existingSymbols = label.parentGroup.element.querySelectorAll('[data-custom-legend-symbol]');
```
helenb, line 116: Additional code here to remove any existing symbols (I found when I moved this function to be called in the render event that the custom symbols were duplicated)

helenb:
> Looks good to me. I did notice that the adjustChartHeight method is triggered multiple times on the page, just wondering if there's a specific reason for that?

Thanks Precious, I double checked this wasn't an oddity with how I have the code running - it does indeed seem to be a feature of the render event that it runs multiple times - which means we need to be a bit careful with performance - but it does seem to feel quite snappy despite that.

From https://www.highcharts.com/blog/tutorials/introduction-to-highcharts-events/:

"The [render event](https://api.highcharts.com/highcharts/chart.events.render) is a very universal and powerful event. It is fired after the load phase and after a few other events. Basically, whenever anything is changed in the chart – this event is triggered.
This function has to be treated with caution because you might accidentally create an infinite loop with this event.
Here are examples of how to use this event:"

precious-onyenaucheya-ons (approved):

PR #3624, opened 2025-06-09
Update details panel to use HTML component

janehughes:

Ticket: https://jira.ons.gov.uk/browse/ONSDESYS-444

This PR updates the existing details panel to use the native `<details>` element and progressively enhance with JS.

- When JS is enabled: the `open` attribute is removed on page load, which closes the `<details>` elements by default
- When JS is disabled: the `open` attribute remains on `<details>` so the `<summary>` content is always shown
- When printing the page: `open` attribute gets added when the print dialogue appears, then removed again when the dialog is closed, to ensure the `<summary>` content is shown on printed pages.  The JS also checks for any panels which were open before printing and keeps them open after the print dialogue is closed.

- We've also added in the relevant print styles:
```
details::details-content {
    display: contents;
}
```
which automatically displays the content in a printed page without relying on JS - this isn't fully supported by all browsers yet but adding it in provides an extra fallback for now which is supported by JS. We can look at removing the javascript once this feature is fully supported!

By using the `<details>` element we allow users to be able to search the page content using CTRL + F which isn't possible with the existing JS solution.

### How to review this PR

- Review the new details-panel component locally and see that it displays and functions as expected
- Add a few extra details panels to the example to test multiple components on the same page to see that they function correctly
- Test printing the page and see that the panels open & close to display the content as expected

### Checklist

### DIFF
--- a/src/components/card/_card.scss
+++ b/src/components/card/_card.scss
@@ -5,6 +5,7 @@
     &--feature {
         background-color: var(--ons-color-banner-bg);
         padding: 2rem;
+        outline: 1px solid transparent;
     }
 
     &__link {
--- a/src/components/card/example-card-set-with-headline-figures.njk
+++ b/src/components/card/example-card-set-with-headline-figures.njk
@@ -1,8 +1,8 @@
 {% from "components/card/_macro.njk" import onsCard %}
 
 <div class="ons-container">
-    <div class="ons-grid ons-grid-flex ons-grid-flex--vertical-top">
-        <div class="ons-grid__col ons-grid__col--flex-col ons-grid__col--stretch ons-u-flex-grow ons-col-4@m">
+    <div class="ons-grid ons-grid-flex-gap ons-grid-flex-gap--s ons-grid-flex--vertical-top">
+        <div class="ons-grid__col ons-grid__col--flex-col ons-grid__col--stretch">
             {{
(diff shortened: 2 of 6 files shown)


src/components/details-panel/_macro.njk
```diff
@@ -1,18 +1,18 @@
...
-                    {% set titleTag = params.headingLevel | default(2) %}
-                    {% set openingTag = "<h" ~ titleTag %}
-                    {% set closingTag = "</h" ~ titleTag ~ ">" %}
-                    {{ openingTag | safe }}
-                    class="ons-details-panel__banner-title ons-u-fs-m ons-u-mb-2xs">{{ params.title }}{{ closingTag | safe }}
-                    <div class="ons-details-panel__banner-detail ons-js-details-heading">
-                        <span class="ons-details-panel__banner-detail-title ons-js-corrections-details-title ons-u-mr-3xs"
-                            >Show detail</span
-                        >
+                    <h3 class="ons-details-panel__banner-title ons-u-mb-2xs">{{ params.title }}</h3>
```
rmccar, line 10: Is this always going to be a h3?

rmccar:
Ive noticed with this that when you turn JS off the details panel always starts open on load of the page because the JS isn't there to close it

PR #3627, opened 2025-06-11
Refactor header macro to fix taborder

janehughes:

Ticket: https://jira.ons.gov.uk/browse/ONSDESYS-464

This PR refactors the header macro to move the menu navigation markup directly after the menu toggle button which has the following impact:

- the taborder now follows a logical reading order in the page, which fixes this accessibility issue by tabbing through the links in the order they appear in the DOM
- the menu dropdown is now absolutely positioned, to ensure that it still displays in line with the designs
- now that the menu uses `position: absolute` it behaves slightly differently when open, overlapping the content beneath the header rather than pushing it down the page
- the search toggle still functions the same way because it is the last item in the nav menu, which means the search dropdown naturally follows a logical order in the DOM so we don't need to update the markup for this

### How to review this PR

- review the header with search button example and test the menu and search dropdowns
- see that the menu dropdown still displays in line with the designs
- review across mobile & desktop to check the dropdowns still function correctly

### Checklist


janehughes:
Have raised this with Joe to check the updated non-JS version, as we've had to change the way it displays since updating the positioning of the menus. Will update on here once he's provided any feedback!

SriHV (approved):

PR #3632, opened 2025-06-13
Fixes to aspect ratio and tick intervals

helenb:

see https://jira.ons.gov.uk/browse/CCB-143

This responds to two issues raised by @nimasmi  testing the charts in the CMS

- There was one more case where we were calling functionality from `common-chart-options.js` when it should have been in `specific-chart-options.js`, which was the tick interval settings - which meant that if there was more than one chart on a page, these were being set incorrectly for the second chart
- I have moved the `hideDataLabels` function as well as a precaution, although there is nothing chart specific in this function definition
- There was a bug with the chart aspect ratio when only one value out of `percentageHeightDesktop` and `percentageHeightMobile` was set, which should now be addressed.

### How to review this PR

On a local build, try changing the `percentageHeightDesktop` and percentageHeightMobile` values in a test chart so there is only one, and ensure that resizing up and down from desktop to mobile works as expected - and that the default height of 400px is used for the chart in the case where no value is set.

Try adding more than one chart to a new example file, and ensure that if you set the tick intervals for the first chart, it doesn't then set them for the second chart.

Check that `hideDataLabels` still functions as expected.

### Checklist

### DIFF
--- a/src/components/chart/specific-chart-options.js
+++ b/src/components/chart/specific-chart-options.js
@@ -2,7 +2,7 @@ import ChartConstants from './chart-constants';
 
 // Options that rely on the chart config but are not specific to the chart type
 class SpecificChartOptions {
-    constructor(theme, type, config) {
+    constructor(theme, type, config, xAxisTickInterval, yAxisTickInterval) {
         this.constants = ChartConstants.constants();
         this.theme = theme;
         this.config = config;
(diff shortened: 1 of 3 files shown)


src/components/chart/specific-chart-options.js
```diff
@@ -42,18 +58,25 @@ class SpecificChartOptions {
...
     adjustChartHeight = (currentChart, percentageHeightDesktop, percentageHeightMobile) => {
-        // get height and width of the plot area
-        const plotHeight = currentChart.plotHeight;
+        // get current width of the plot area
         const plotWidth = currentChart.plotWidth;
-        // calculate the new plot height based on the percentage height
-        // default to the current height
-        let newPlotHeight = plotHeight;
+        let newPlotHeight = undefined;
+        let totalHeight = 400; // Highcharts default height - needed if one of the percentage heights is undefined
```
helenb, line 64: Using the Highcharts default height rather than the current chart height as the default - this means that if there is an undefined value for `percentageHeightmobile`, it won't stay on the desktop height when you resize down (and vice versa),

PR #3635, opened 2025-06-17
Fix/resolve failing tests introduced in recent commits

admilne:

Ticket: [ONSDESYS-528](https://jira.ons.gov.uk/browse/ONSDESYS-528)

[PR-3510](https://github.com/ONSdigital/design-system/pull/3510) introduced a change which resulted in our pipelines passing, even if there were failing tests. This has led to some failing tests being added to our main branch. This PR fixes those consistently failing tests.

### How to review this PR

- Pull this branch down and check the tests pass (please note there seems to be a couple of flakey tests around the timeout-model and video components. This PR doesn't fix those so if either of those tests fail, please try running the tests again.

### Checklist

### DIFF
--- a/src/components/language-selector/_macro.spec.js
+++ b/src/components/language-selector/_macro.spec.js
@@ -102,7 +102,7 @@ describe('macro: language-selector', () => {
         it('has the `abbrText` rendered', () => {
             const $ = cheerio.load(renderComponent('language-selector', EXAMPLE_WITH_TWO_LANGUAGES));
 
-            expect($('.ons-language-links__item a span:first-child').text()).toBe('CY');
+            expect($('.ons-language-links__item a span:nth-child(2)').text()).toBe('CY');
         });
     });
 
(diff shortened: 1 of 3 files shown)


src/components/language-selector/_macro.spec.js
```diff
@@ -124,8 +124,9 @@ describe('macro: language-selector', () => {
         it('does not show the current language', () => {
             const $ = cheerio.load(renderComponent('language-selector', EXAMPLE_WITH_THREE_LANGUAGES));
 
-            expect($('.ons-language-links__item:first-child a').text()).toBe('English');
-            expect($('.ons-language-links__item:last-child a').text()).toBe('Polski');
+            // .replace(/\s+/g, ' ') will replace any sequence of whitespace characters (spaces, tabs, newlines) with a single space
+            expect($('.ons-language-links__item:first-child a').text().replace(/\s+/g, ' ').trim()).toBe('Change language to English');
```
rmccar, line 128: So are we getting incorrect spacing in the link text?

admilne: Yeah, because an additional span has been added before the language specified which reads as `<span class="ons-u-vh">Change language to{{ ' ' }}</span>`, we then have the original span with the language as we had it before. It reads as `Change language to English` but there is some odd spacing before, after and in between.

Also, "Change language to" is fixed, so we might need to look at that further down the line so we can get that to work for various other languages but I figured that's a ticket for another day.

rmccar, line 128: I think updating the macro and changing
`<span class="ons-u-vh">Change language to{ ' ' }</span>`
to
`<span class="ons-u-vh">Change language to </span>`
should fix this issue

PR #3644, opened 2025-06-19
Update heading level style  for Chart Component to match DS standard

precious-onyenaucheya-ons:

This PR updates the font sizes for text elements surrounding charts (title, subtitle, caption, and download title) to align with the standard Design System (DS) font scale. 

The headings used are still configurable via the headingLevel parameter, but now they will correctly reflect the expected hierarchy and scale:

When headingLevel is set to 2:

- Title uses `<h2>` styling (matching DS heading 2)
- Subtitle uses `<h3>` styling (matching DS heading 3)

When headingLevel is set to 3:

- Title uses `<h3>` styling (matching DS heading 3)
- Subtitle uses `<h4>` styling (matching DS heading 4)

When headingLevel is set to 4:

- Title uses `<h4>` styling
- Subtitle uses `<h5>` styling

This change ensures consistency across all chart examples and real-world usage within the design system and publications.

Related to: [ONSDESYS-34](https://jira.ons.gov.uk/browse/CCB-34) 

### How to review this PR
1. Pull down the branch and build the project.
2. Open any example chart from the Design System preview or a page using the chart component.
3. Verify that:
- Chart title uses the correct heading level and font size based on headingLevel input.
- Subtitle uses the correct subsequent heading level and font size.

### Checklist


src/components/chart/_chart.scss
```diff
@@ -8,19 +8,7 @@
         @extend .ons-u-fs-r--b;
     }
 
-    &__title {
```
helenb, line 11: Suggestion: these classes could also be removed from the macro template if they are no longer styled

PR #3645, opened 2025-06-20
Update language switcher location & create new basic header variants

janehughes:

This is based on https://github.com/ONSdigital/design-system/pull/3620

- Updates the position of the language switcher when JS is disabled
- Creates a new header basic variant examples with language switcher included, and renames existing basic header variant examples for clarity

### How to review this PR

- Check the layout of the menu buttons/links when JS is enabled and disabled

### Checklist


src/components/footer/example-footer-cymraeg.njk
```diff
@@ -1,7 +1,3 @@
----
-'fullWidth': true
----
-
```
rmccar, line 4: Was this change inherited from my PR? Just because I thought Id leave this change for now

src/components/header/example-header-basic-with-search-button.njk
rmccar, line 1: If we are going to rename these I think  removing the language from the `example-header-basic` example would make sense

janehughes: Yeah that's a fair point, I've removed it 👍

src/components/footer/example-footer-cymraeg.njk
rmccar: Yeah I think revert it

rmccar: Other than this it looks good

SriHV (approved):

PR #3641, opened 2025-06-18
Include iframe option in chart component

helenb:
### What is the context of this PR?

https://jira.ons.gov.uk/browse/CCB-99

Adds an option to embed a separately published visualisation or chart inside an iframe.

### How to review this PR

On you local build, check that the new example works and that the tests pass.
Also test with JavaScript disabled, to see how the fallback image (inside the iframe) displays.

### Checklist

This needs to be completed by the person raising the PR.

-   [x] I have selected the correct Assignee
-   [x] I have linked the correct Issue

PR #3646, opened 2025-06-23
Charts fallback image for when there is no JavaScript

helenb:

See https://jira.ons.gov.uk/browse/CCB-138

Adds functionality to display a fallback image screenshot if JavaScript is disabled
For iframed charts, if the fallback image is provided, then the iframe will be hidden when JavaScript is disabled. If the fallback image is not provided, the iframe will be shown, as it is assumed in that case that the iframed content will display a fallback image instead.

In order that the fallback image always has a description, this also implements https://jira.ons.gov.uk/browse/CCB-138

### How to review this PR
Fallback images have been added for the line chart and the area chart - view either of these with JavaScript disabled to see the fallback image.

### Checklist

### DIFF
--- a/src/components/chart/chart.dom.js
+++ b/src/components/chart/chart.dom.js
@@ -1,8 +1,10 @@
 import HighchartsBaseChart from './chart';
 import domready from '../../js/domready';
 
-domready(async () => {
-    [HighchartsBaseChart].forEach((Component) => {
-        document.querySelectorAll(Component.selector()).forEach((el) => new Component(el));
+document.fonts.ready.then(() => {
+    domready(async () => {
+        [HighchartsBaseChart].forEach((Component) => {
+            document.querySelectorAll(Component.selector()).forEach((el) => new Component(el));
+        });
     });
 });


janehughes (commented):
This LGTM. Though currently the description param is optional and I'm wondering if we should make this required to make sure it's definitely there as a fallback. I can imagine this would be raised by DAC if there wasn't a text alternative to describe the chart

helenb:
> This LGTM. Though currently the description param is optional and I'm wondering if we should make this required to make sure it's definitely there as a fallback. I can imagine this would be raised by DAC if there wasn't a text alternative to describe the chart

This is a good point. I will raise a separate ticket for this as part of the accessibility work needed.

src/components/chart/_macro-options.md
```diff
@@ -23,6 +23,7 @@
 | estimateLineLabel        | string  | false                  | Label for the estimate line in the legend. Used in column charts with confidence intervals. Will be ignored if an iframe url is set.                                                                                       |
 | uncertaintyRangeLabel    | string  | false                  | Label for the confidence interval in the legend. Used in column charts showing uncertainty ranges. Will be ignored if an iframe url is set.                                                                                |
 | iframeURL                | string  | false                  | A url for a visualisation to display in an iframe instead of displaying a Highcharts chart.                                                                                                                                |
+| fallbackImageURL         | string  | false                  | A url for a fallback image to display instead of the chart or iframed visualisation if JavaScript is disabled. Will be ignored if an iframe url is set, as the iframed content will provide the fallback image.            |
```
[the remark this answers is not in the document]
helenb: At the moment it's using camel case for the words `fallback image URL` becoming `fallbackImageURL` - so I think it is correct? Or would the norm in the DS be to use `fallbackImageUrl`?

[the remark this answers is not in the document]
helenb: Now updated.

src/components/chart/_macro.njk
```diff
@@ -81,6 +81,13 @@
                     <p data-invalid-chart-type><b>Chart type "{{ params.chartType }}" is not supported</b></p>
                 {% endif %}
             {% endif %}
+            {# If the chart is displayed inside an iframe, then the fallback image is not displayed here, as it will be provided in the iframe content instead #}
```
rmccar, line 84: This comment may need to be updated with what we decide on the other PR

src/components/chart/_macro.spec.js
```diff
@@ -1466,6 +1481,19 @@ describe('Macro: Chart', () => {
...
+                    renderComponent('chart', {
+                        ...EXAMPLE_IFRAME_CHART_PARAMS,
+                        fallbackImageUrl: '/img/small/line-chart-screenshot.png',
+                    }),
+                );
+                const noScript = $(`noscript[data-fallback-image]`);
+                test('THEN: it does  not render the fallback image', () => {
+                    expect(noScript.length).toBe(0);
+                });
+            });
```
rmccar, line 1496: This will also probably need to be updated

src/components/chart/_macro-options.md
```diff
@@ -23,6 +23,7 @@
 | estimateLineLabel        | string  | false                  | Label for the estimate line in the legend. Used in column charts with confidence intervals. Will be ignored if an iframe url is set.                                                                                       |
 | uncertaintyRangeLabel    | string  | false                  | Label for the confidence interval in the legend. Used in column charts showing uncertainty ranges. Will be ignored if an iframe url is set.                                                                                |
 | iframeUrl                | string  | false                  | A url for a visualisation to display in an iframe instead of displaying a Highcharts chart.                                                                                                                                |
+| fallbackImageUrl         | string  | false                  | A url for a fallback image to display instead of the chart or iframed visualisation if JavaScript is disabled. Will be ignored if an iframe url is set, as the iframed content will provide the fallback image.            |
```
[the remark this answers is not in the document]
helenb: Hi @rmccar I believe we need it to be optional because of how the wider charts ecosystem is using. We are planning a separate service which will load the DS component in order to automatically take a screenshot of the chart, which will then be used by the chart builder to supply the fallback image. The chart builder will enforce that a fallback image is always used, but we can't make it required here otherwise it won't be possible to load the component to take the screenshot. cc @MaciekBaron who is working on the fall back image screenshots, to confirm.

src/components/chart/_macro.njk
```diff
@@ -81,6 +81,13 @@
...
+            {% if params.fallbackImageUrl and not params.iframeUrl %}
+                <noscript data-fallback-image>
+                    {# alt text is empty as the description is provided above #}
+                    <img src="{{ params.fallbackImageUrl }}" alt="" class="ons-chart__fallback-image" />
```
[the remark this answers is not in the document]
helenb: If the alt text is empty, the screen reader won't announce that there is an image at all. What would be better though on consideration is to use `aria-described-by` and point to an id on the description, so I will use that.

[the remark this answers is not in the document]
helenb: The above is now implemented.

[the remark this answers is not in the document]
rmccar: Would be good to run this by Jane but I would think we would want alt text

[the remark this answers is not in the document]
helenb: I've tested with the screen reader and the `aria-describedby` approach means it reads the description twice. It works nicely to leave the alt text empty, and let the screen reader read the description and skip the image though. I will get a second opinion from @janehughes when she is back in.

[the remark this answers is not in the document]
helenb: HI @rmccar the `aria-describedby` can be used for a longer description, whereas the alt text should be something much shorter.  There's an example of using it with an image here: https://www.w3.org/WAI/GL/wiki/Using_aria-describedby_to_provide_descriptions_of_images. However, you're right that it's probably not correct to leave the image as though it were decorative, so I have added brief `alt` text to explain that it is the fallback image. Also, when I discussed it with Jane she pointed out that because we are using a `figure` element the description makes more sense to be associated with that - so I have moved the `aria-describedby` to the `figure` element instead.

src/components/chart/_macro.njk
```diff
@@ -87,11 +87,10 @@
             {# If the chart is displayed inside an iframe, then the fallback image is not displayed here, as it will be provided in the iframe content instead #}
             {% if params.fallbackImageUrl and not params.iframeUrl %}
                 <noscript data-fallback-image>
-                    {# alt text is empty as the description is provided above #}
+                    {# Note that a more detailed description of the chart is provided in the audio description above #}
                     <img
                         src="{{ params.fallbackImageUrl }}"
-                        aria-describedby="{{ audioDescriptionId }}"
-                        alt=""
+                        alt="Fallback image for the chart as JavaScript is disabled"
```
rmccar, line 93: This text should probably be configurable and have a default instead of being hardcoded. It would mean that other languages will be able to be used for example

src/components/chart/_macro-options.md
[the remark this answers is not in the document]
helenb: It also needs to be optional so that we have the option to use a fallback image within the iframe - as per the discussion on https://github.com/ONSdigital/design-system/pull/3641

src/components/chart/_macro.njk
helenb: That's now configurable with a default value if not set.

src/components/chart/_macro.njk
```diff
@@ -86,6 +93,17 @@
                     </p>
                 {% endif %}
             {% endif %}
+
+            {% if params.fallbackImageUrl %}
+                <noscript data-fallback-image>
```
rmccar, line 98: Rather than use data attributes to identify elements we've tended to use classes

src/components/chart/_macro.njk
```diff
@@ -48,17 +49,23 @@
...
-            {% if params.description %}
-                <p class="ons-u-vh">{{ params.description }}</p>
-            {% endif %}
+
+            <p class="ons-u-vh" id="{{ audioDescriptionId }}">{{ params.description }}</p>
+
             {% if params.iframeUrl %}
-                <div data-chart-iframe id="{{ params.id }}" data-url="{{ params.iframeUrl }}" data-title="{{ params.title }}">
+                <div
+                    data-chart-iframe
```
rmccar, line 63: Same here, the pattern we use is to use a class, so this would be something like "ons-chart__iframe"

src/components/chart/_macro.spec.js
```diff
@@ -1648,6 +1663,19 @@ describe('Macro: Chart', () => {
                     expect($('[data-invalid-chart-type]').length).toBe(0);
                 });
             });
+
+            describe('GIVEN: Params: Fallback Image URL with an iframe chart', () => {
```
rmccar, line 1667:
```suggestion
            describe('GIVEN: Params: fallbackImageUrl', () => {
```
This should be this, and then the a WHEN should mention the use with an iframe chart

rmccar:
I think the description on this PR needs updating now

src/components/chart/_macro.njk
helenb: This one is being used as a hook for the pym.js resizing - is a class really a good idea for that, given it might be changed for styling reasons? I could use an id if that is preferred.

src/components/chart/_macro.spec.js
```diff
@@ -193,6 +193,21 @@ describe('Macro: Chart', () => {
             });
         });
 
+        describe('GIVEN: Params: Fallback Image URL', () => {
+            describe('WHEN: fallback image URL is provided', () => {
```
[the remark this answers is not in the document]
helenb: Updated

src/components/chart/_macro.njk
helenb: I've used an id for this one like we did for the footnotes

src/components/chart/_macro.spec.js
```diff
@@ -193,6 +193,21 @@ describe('Macro: Chart', () => {
...
+                const $ = cheerio.load(
+                    renderComponent('chart', {
+                        ...EXAMPLE_LINE_CHART_WITH_CONFIG_PARAMS,
+                        fallbackImageUrl: '/img/small/line-chart-screenshot.png',
+                    }),
+                );
+                const noScript = $(`noscript[data-fallback-image]`).html();
```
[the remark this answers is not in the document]
helenb: Updated

src/components/chart/_macro-options.md
```diff
@@ -1,30 +1,32 @@
...
+| footnotes                | `Object<Footnotes>` | false                  | Footnotes to appear below the chart in a 'details' element                                                                                                                                                                                                                                                                                            |
+| fallbackImageUrl         | string              | false                  | A url for a fallback image to display instead of the chart or iframed visualisation if JavaScript is disabled. If a fallback image is passed with an iframeUrl parameter, the iframe will be hidden for non-JavaScript uesrs. If it is not passed, it is assumed the iframed content will provide a fallback image to be displayed inside the iframe. |
```
rmccar, line 28:
```suggestion
| fallbackImageUrl         | string              | false                  | A url for a fallback image to display instead of the chart or iframe visualisation if JavaScript is disabled. If a fallback image is passed with an iframeUrl parameter, the iframe will be hidden for non-JavaScript users. If it is not passed, it is assumed the iframe content will provide a fallback image to be displayed inside the iframe. |
```
Couple of typos

src/components/chart/_macro.njk
rmccar: Its just the pattern we have gone with in the DS, we have done it in other components and I think it helps describe the structure of the elements. It is something we could review but I wouldn't see it changing. For this PR Id rather conform here to the way other components are written and then review that as part of another piece of work on the whole of the DS

src/components/chart/_macro.spec.js
```diff
@@ -193,6 +193,23 @@ describe('Macro: Chart', () => {
...
+                    renderComponent('chart', {
+                        ...EXAMPLE_LINE_CHART_WITH_CONFIG_PARAMS,
+                        fallbackImageUrl: '/img/small/line-chart-screenshot.png',
+                        fallbackImageAlt: 'A description of the fallback image for screen readers',
+                    }),
+                );
+                const noScriptFallbackImage = $(`#fallback-image--chart-456`).html();
+                test('THEN: it renders the fallback image', () => {
+                    expect(noScriptFallbackImage).toContain('/img/small/line-chart-screenshot.png');
+                    expect(noScriptFallbackImage).toContain('alt="A description of the fallback image for screen readers"');
```
rmccar, line 208: I think this should be split out into another test as its testing a different param. We've tried to only have one expect per test if we can. I think we also need to test the default works when the param is not provided

src/components/chart/_macro.spec.js
```diff
@@ -1648,6 +1665,19 @@ describe('Macro: Chart', () => {
                     expect($('[data-invalid-chart-type]').length).toBe(0);
```
rmccar, line 1665: I think this needs to be updated as I have said in a previous comment about using classes

src/components/chart/_macro.njk
```diff
@@ -48,17 +49,23 @@
...
+                    id="{{ params.id }}"
+                    data-url="{{ params.iframeUrl }}"
+                    data-title="{{ params.title }}"
+                    class="{% if params.fallbackImageUrl %}ons-chart__non-js-hide{% endif %}"
+                >
                     <iframe src="{{ params.iframeUrl }}" title="{{ params.title }}" class="ons-chart__iframe"> </iframe>
```
rmccar, line 69:
```suggestion
                    <iframe src="{{ params.iframeUrl }}" title="{{ params.title }}" class="ons-chart__iframe"></iframe>
```

src/components/chart/_macro-options.md
helenb: Thanks. I was using 'iframed' in the sense of 'within the iframe' but happy to update.

src/components/chart/_macro.njk
helenb: Are you happy with an id here or would you prefer a class?

rmccar: Id is ok but class would be better just because thats what we've used elsewhere

src/components/chart/_macro.spec.js
helenb: Can we move this to the code tidy up ticket? There's a lot of feedback coming in here that isn't relevant to the PR and this is blocking other work.

src/components/chart/_macro-options.md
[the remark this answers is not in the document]
helenb: Updated.

src/components/chart/_macro.spec.js
helenb: Done.

src/components/chart/_macro.njk
```diff
@@ -48,17 +49,22 @@
...
-            {% if params.description %}
-                <p class="ons-u-vh">{{ params.description }}</p>
-            {% endif %}
+
+            <p class="ons-u-vh" id="{{ audioDescriptionId }}">{{ params.description }}</p>
+
             {% if params.iframeUrl %}
-                <div data-chart-iframe id="{{ params.id }}" data-url="{{ params.iframeUrl }}" data-title="{{ params.title }}">
+                <div
+                    class="ons-chart__iframe-wrapper {% if params.fallbackImageUrl %}ons-chart__non-js-hide{% endif %}"
```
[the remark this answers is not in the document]
rmccar: Yeah that makes sense, I don't know if I'm used to seeing it like this but I don't find it less legible. Its a pattern we use elsewhere

src/components/chart/_macro.spec.js
rmccar: Yeah thats fair

src/components/chart/_macro.spec.js
```diff
@@ -193,6 +193,37 @@ describe('Macro: Chart', () => {
...
+            describe('WHEN: fallbackImageAlt is provided', () => {
+                const $ = cheerio.load(
+                    renderComponent('chart', {
+                        ...EXAMPLE_LINE_CHART_WITH_CONFIG_PARAMS,
+                        fallbackImageUrl: '/img/small/line-chart-screenshot.png',
+                        fallbackImageAlt: 'A description of the fallback image for screen readers',
+                    }),
+                );
+                const noScriptFallbackImage = $(`#fallback-image--chart-456`).html();
+                test('THEN: it renders the fallback image', () => {
```
rmccar, line 221:
```suggestion
                test('THEN: the fallback image has the provided alt text', () => {
```

src/components/chart/_macro.njk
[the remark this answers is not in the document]
helenb: I've updated it to be consistent with the rest of the DS,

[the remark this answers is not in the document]
rmccar: We still need the space after `{% if params.fallbackImageUrl %}` otherwise the two class names will run into eachother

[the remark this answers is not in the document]
helenb: It was because the `git commit` failed as it thought it was an empty commit - I've now overridden that and the space should be in place.

precious-onyenaucheya-ons (approved):

PR #3648, opened 2025-06-24
Add footnotes below the chart

helenb:

See https://jira.ons.gov.uk/browse/CCB-125
This PR builds on https://github.com/ONSdigital/design-system/pull/3641

- Adds an option to display footnotes below the chart using the `Details` component.
- Renames the classes on the annotations footnotes to avoid confusion
- Tweaks the annotations footnotes line spacing based on previous feedback.

### How to review this PR

- Test on your local build. Footnotes have been added to line chart, bar chart and scatter chart examples.
- Ensure the annotations footnotes display correctly.

### Checklist


src/components/chart/_macro.njk
```diff
@@ -125,6 +126,24 @@
...
+            {{
+                onsDetails({
+                    "id": footnotesId,
+                    "title": params.footnotes.title,
+                    "content": params.footnotes.content,
+                    "headingLevel": headingLevel + 2,
+                    "classes": "ons-u-mt-xl",
+                    "attributes": {
+                        "data-footnotes": footnotesId
+                    }
```
rmccar, line 142: Just wondering whats this attribute used for? If its just for the css selector in the tests we could use something else to select the element

src/components/chart/_macro.spec.js
```diff
@@ -954,6 +954,26 @@ describe('Macro: Chart', () => {
...
+                            title: 'Footnotes',
+                            content:
+                                '<ol><li>Non-store retailing refers to retailers that do not have a store presence. While the majority is made up of online retailers, it also includes other retailers, such as stalls and markets.</li><li>More data are available in our Retail Sales Index datasets.</li></ol>',
+                        },
+                    }),
+                );
+                test('THEN: it renders the footnotes', () => {
+                    expect($('[data-footnotes]').length).toBe(1);
+                    expect($('[data-footnotes]').attr('data-footnotes')).toBe('footnotes--area-chart-123');
+                    expect($('[data-footnotes]').find('ol').length).toBe(1);
```
rmccar, line 973: This is just testing that the element is there and there is one of them, would be good to test the content is correct instead

src/components/chart/_macro.njk
helenb: Yep it's used for the tests. In general I always avoid using class names to target DOM elements, as they could change when styling is updated and unexpectedly impact the JS behaviour. But let me know if you use a different convention in the DS.

rmccar: Could you not use the id?

helenb: I've updated that now.

src/components/chart/_macro.spec.js
rmccar: Yeah thats fair enough, what you've done is fine

rmccar (approved):

precious-onyenaucheya-ons (approved):

Issue #3643, opened 2025-06-19
New header params need to be passed in pageConfig


rmccar:
We have added new params recently to the header component to add the menu and search functionality needed for the new website. But we haven't updated `pageConfig` in the page template to use and pass these new params so they cannot be used through the page template at the moment and only can be used through calling the component directly. We should add these new params to the `pageConfig` so that they can be used in this way.

Joe-Nuttall:
Link to JIRA - https://jira.ons.gov.uk/browse/ONSDESYS-551

PR #3657, opened 2025-07-02
Added new Header params in pageConfig

SriHV:

[ONSDESYS-551](https://jira.ons.gov.uk/browse/ONSDESYS-551)
#3643 
As per the user story, updated `pageConfig` with new header params - `menuLinks`, `searchLinks`, `mastheadLogoAltText` in the page template.

### How to review this PR
- Pass these params into a pageConfig (see HEADER_BASIC_EXAMPLE in the `_template.spec.js`) and check that header component is rendered right
- Check that new params are added in the page template and verify that the new snapshot generated has the right structure of the header component with `basic` variant

### Checklist


src/layout/_template.spec.js
```diff
@@ -108,6 +108,186 @@ const FULL_EXAMPLE = `
                 "text": 'Menu',
                 "ariaLabel": 'Toggle main menu'
             },
+            "menuLinks": {
```
rmccar, line 111: Just adding these here are not going to change the html in the snapshot and therefore isn't testing them. Setting these only change the header when we use the basic variant. So we will have to add another snapshot test

rmccar, line 111: Have you tested this configuration to see what it renders? Its highlighted a problem with the spacing on the button that looks like an issue that needs to be raised

SriHV: Can you attach a screenshot please if its possible. When I tested I see this and its looking fine to me
Screenshot 2025-07-09 at 08 31 27

rmccar: This issue was caused when variant wasn't set to basic, which was the case with the last configuration

rmccar:
PR description needs updating

src/components/header/example-header-basic.njk
rmccar, line 1: I think you accidentally pushed your test code

src/layout/_template.spec.js
```diff
@@ -348,6 +348,197 @@ const NO_FAVICONS_EXAMPLE = `
...
+                                {
+                                    "text": "Tourism industry"
+                                }
+                            ]
+                        },
+                        {
+                            "heading": "Employment and labour market",
+                            "url": "#0",
+                            "groupItems":
+                            [
```
rmccar, line 458:
```suggestion
                            "groupItems": [
```

src/layout/_template.spec.js
```diff
@@ -348,6 +348,197 @@ const NO_FAVICONS_EXAMPLE = `
...
+                                {
+                                    "text": "People in work"
+                                },
+                                {
+                                    "text": "People not in work"
+                                }
+                            ]
+
```
rmccar, line 466:
```suggestion
```

precious-onyenaucheya-ons (approved):

PR #3676, opened 2025-07-10
Improve focus states for chart

precious-onyenaucheya-ons:

Addresses [ONSDESYS-586](https://jira.ons.gov.uk/browse/ONSDESYS-586)
This PR enhances the keyboard focus behavior for charts rendered with Highcharts, improving accessibility and aligning with ONS Design System standards.

Key changes include:

- Enabled Highcharts keyboard navigation to allow arrow key interaction
- Wrapped the chart in a custom, focusable container (tabindex="0") to control how and when focus enters the chart
- Applied Design System focus styling (black border with orange accent) to the chart container for visual clarity
- Added a visually hidden description for screen readers to guide keyboard users on how to navigate between data points.

These changes ensure users can:

- Tab to the chart container (without immediately focusing a data point).
- Press Tab again to move focus into the chart itself .
- Use arrow keys to navigate between data points, triggering tooltip updates without immediate jumps or visual highlight outlines, overriding Highcharts’ default behavior.

### How to review this PR
Navigate to the chart example page
Use the Tab key to move through the page

- Focus should land first on the chart container, not directly on a data point
- The container should display a black and orange outline on focus

Interact with the chart using arrow keys

While the container is focused, press tab again and then press any arrow key (←, →, ↑, or ↓)

- Highcharts should begin navigating between data points
- Tooltips should be displayed with each movement but there should be no visual highlights  

### Checklist


admilne (commented):
I'm a little confused about how the focus should work. Your PR description suggests that you should be able to tab to the chart container which displays the black and orange outline (so far so good).

It then says, while the container is focused, you should be able to navigate between the data points with the arrow keys.

For me, I have to focus on the chart, then press tab which makes the chart lose focus but I'm able to navigate the datapoints. Is this incorrect?

I tested this on the example-bar-chart

precious-onyenaucheya-ons:
> I'm a little confused about how the focus should work. Your PR description suggests that you should be able to tab to the chart container which displays the black and orange outline (so far so good).
> 
> It then says, while the container is focused, you should be able to navigate between the data points with the arrow keys.
> 
> For me, I have to focus on the chart, then press tab which makes the chart lose focus but I'm able to navigate the datapoints. Is this incorrect?
> 
> I tested this on the example-bar-chart
> 
> This also doesn't appear to work at all for the iFrame chart
Thanks for pointing that out , sorry if the PR description was a bit unclear.

To clarify:
You first tab into the chart container, which shows the black and orange focus outline. Then, you press Tab again to move focus into the chart itself, where the first data point receives focus. From there, you can use the arrow keys to navigate between data points.

Also,this behaviour is specific to the chart examples that use Highcharts directly. It doesn’t currently work for charts embedded via an iFrame.

I will update the PR description to make this clearer.

rmccar:
This all looks ok, just had one question on the design, wouldn't we want some sort of visible focus state other than the tooltip? Its not obvious to me what has focus as the tooltip isn't consistent with other focus states

precious-onyenaucheya-ons:
> This all looks ok, just had one question on the design, wouldn't we want some sort of visible focus state other than the tooltip? Its not obvious to me what has focus as the tooltip isn't consistent with other focus states

Regarding the visible focus state, I agree it’s not as visually obvious as a typical focus indicator. However, this approach was based on guidance from the Data Viz team, who specifically requested that only the chart container show the standard focus styling (black with orange border), and that individual data points should not have a visible focus ring. The tooltip is used as the primary indicator of which point is currently active.

rmccar:
> > This all looks ok, just had one question on the design, wouldn't we want some sort of visible focus state other than the tooltip? Its not obvious to me what has focus as the tooltip isn't consistent with other focus states
> 
> Regarding the visible focus state, I agree it’s not as visually obvious as a typical focus indicator. However, this approach was based on guidance from the Data Viz team, who specifically requested that only the chart container show the standard focus styling (black with orange border), and that individual data points should not have a visible focus ring. The tooltip is used as the primary indicator of which point is currently active.

I would question if that is the right call for users. Has this been run by any designers or anyone that has knowledge of accessibility? Tool tips aren't the most accessible thing as it is but to then not have any visible focus on data points doesn't sound right to me

rmccar:
Due to the way the user needs to navigate the points on the chart we may need to explain this via instructions visibly hidden text like we have done on the autosuggest component. This seems like a similar situation to that.

precious-onyenaucheya-ons:
> Due to the way the user needs to navigate the points on the chart we may need to explain this via instructions visibly hidden text like we have done on the autosuggest component. This seems like a similar situation to that.

Updated

src/components/chart/_macro.njk
```diff
@@ -91,7 +92,12 @@
                         tabindex="0"
                         role="region"
                         aria-label="chart container"
+                        aria-describedby="{{ instructionsId }}"
                     >
+                        <div id="{{ instructionsId }}" class="ons-u-vh">
+                            Use the Tab key to move focus into the chart. Once inside, use the arrow keys to navigate between data points.
+                            As you move, tooltips will be announced to describe each point.
+                        </div>
```
rmccar, line 100: This shouldn't be hardcoded. Have a look at the autosuggest to see how its been done

precious-onyenaucheya-ons: updated it

rmccar (changes requested):
Just a few minor things

src/components/chart/_macro.njk
```diff
@@ -94,10 +96,7 @@
                         aria-label="chart container"
                         aria-describedby="{{ instructionsId }}"
                     >
-                        <div id="{{ instructionsId }}" class="ons-u-vh">
-                            Use the Tab key to move focus into the chart. Once inside, use the arrow keys to navigate between data points.
-                            As you move, tooltips will be announced to describe each point.
-                        </div>
+                        <div id="{{ instructionsId }}" class="ons-u-vh">{{ instructions | safe }}</div>
```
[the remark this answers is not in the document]
precious-onyenaucheya-ons: updated

src/components/chart/_chart.scss
```diff
@@ -7,6 +7,12 @@
         overflow: visible !important;
     }
 
+    &__container:focus {
+        box-shadow: 0 0 0 3px black !important;
+        outline: 4px solid var(--ons-color-focus) !important;
+        outline-offset: 2px !important;
```
rmccar, line 13: The importants dont seem to be needed here

src/components/chart/_chart.scss
```diff
@@ -7,6 +7,12 @@
         overflow: visible !important;
     }
 
+    &__container:focus {
+        box-shadow: 0 0 0 3px black !important;
```
rmccar, line 11:
```suggestion
        box-shadow: 0 0 0 3px var(--ons-color-black) !important;
```
We can use the var here

src/components/chart/_macro.spec.js
```diff
@@ -1170,7 +1200,7 @@ describe('Macro: Chart', () => {
                 );
 
                 test('THEN: it renders the description for accessibility', () => {
-                    expect($('.ons-u-vh').text()).toBe(accessibleDescription);
+                    expect($('#chart-audio-description-uuid').text()).toBe(accessibleDescription);
```
rmccar, line 1203: Why is the id for this one "uuid" but the rest "invalid-chart-123" for example? Would be good to change this to match the others

rmccar (approved):

PR #3700, opened 2025-07-29
Added option to use external breadcrumbs in hero component 

SriHV:

[ONSDESYS-543](https://officefornationalstatistics.atlassian.net/jira/software/c/projects/ONSDESYS/boards/2876?selectedIssue=ONSDESYS-543)

- As per the ticket we need to move the breadcrumbs out of the hero component to address the DAC issue. This change shouldn't affect visually i.e it should give an impression that both hero and breadcrumbs are one component.

- To address this,  I've created option to use variants in `Breadcrumbs` component. These new variants `"grey", "dark", "pale-blue", "navy-blue"` will match the color of hero variants. 
- I've wrapped the breadcrumbs in a `div` element and used the `ons-container` to bring the breadcrumbs to center of the page to match hero's container and then used negative margin bottom to fit them inside the hero component. 

- However, doing this required moving "bubbles" in navy-blue and pale-blue hero variants. I've reviewed these visual changes with Joe and shifted the position of bubbles to center in navy-blue and removed the top "semi-circular" bubble in pale-blue variants.

I've added the examples for each hero variant with external breadcrumbs

### How to review this PR

- Go to the new examples and tab from the top of the page to reach `skip to main link`.
- Keyboard press the link and check that focus does not land the link within the breadcrumbs.
- Ensure there are no visual inconsistency for these new examples.

### Checklist


admilne (changes requested):
When making the breadcrumbs dark I think the text should be white to match the style of the hero and make the text more readable
image

src/components/breadcrumbs/_macro-options.md
```diff
@@ -1,9 +1,10 @@
...
+| --------- | ------------------------ | -------- | ----------------------------------------------------------------------------------------------------------------------------------- |
+| classes   | string                   | false    | Custom classes to add to the breadcrumbs `nav` element                                                                              |
+| ariaLabel | string                   | false    | The `aria-label` added to the `nav` element. Defaults to “Breadcrumbs”.                                                             |
+| id        | string                   | false    | The `id` added to the `nav` element                                                                                                 |
+| itemsList | `Array<BreadcrumbsItem>` | true     | An array of [breadcrumb items](#breadcrumbsitem)                                                                                    |
+| variant   | string                   | true     | Set variant to "hero-grey", "hero-dark", "hero-default", "hero-pale-blue", "hero-navy-blue" for matching color of the hero variants |
```
admilne, line 7: I don't think we want this variant option to be required. If you look at the breadcrumbs example component it's shown with a white background and I'm assuming we want to keep that as an option.

admilne (changes requested):

rmccar (changes requested):

src/components/breadcrumbs/_macro-options.md
```diff
@@ -1,9 +1,10 @@
...
+| --------- | ------------------------ | -------- | ----------------------------------------------------------------------------------------------------------------------------------- |
+| classes   | string                   | false    | Custom classes to add to the breadcrumbs `nav` element                                                                              |
+| ariaLabel | string                   | false    | The `aria-label` added to the `nav` element. Defaults to “Breadcrumbs”.                                                             |
+| id        | string                   | false    | The `id` added to the `nav` element                                                                                                 |
+| itemsList | `Array<BreadcrumbsItem>` | true     | An array of [breadcrumb items](#breadcrumbsitem)                                                                                    |
+| variant   | string                   | false    | Set variant to "hero-grey", "hero-dark", "hero-default", "hero-pale-blue", "hero-navy-blue" for matching color of the hero variants |
```
rmccar, line 7: `hero-default` and `hero-pale-blue` seem to style the breadcrumbs in exactly the same way. Do we need both?

src/components/breadcrumbs/_macro.spec.js
```diff
@@ -132,4 +132,30 @@ describe('FOR: Macro: Breadcrumbs', () => {
...
+                renderComponent('breadcrumbs', {
+                    ...EXAMPLE_BREADCRUMBS_REQUIRED_PARAMS,
+                    variant: 'test-variant',
+                }),
+            );
+            test('THEN: does not render hero-wrapper', () => {
+                expect($('.ons-hero-wrapper').length).toBe(0);
+            });
+        });
+    });
```
rmccar, line 160: When the logic is sorted out this test will need updating

src/components/hero/example-hero-default-with-external-breadcrumbs.njk
```diff
@@ -169,7 +169,7 @@ layout: ~
     {{
         onsBreadcrumbs({
             "ariaLabel": 'Breadcrumbs',
-            "variant": "default",
+            "variant": "pale-blue",
```
rmccar, line 172: This can just not be set for this example

SriHV: If we remove the variant from here. The container class used to bring the breadcrumbs to center of the page and match hero's container class will not be set . 
image

rmccar:
> Screenshot 2025-08-19 at 11 46 49
> This link in the hero using the dark variant when focused doesn't have black text

This is still an issue

src/components/breadcrumbs/_breadcrumbs.scss
```diff
@@ -75,9 +107,35 @@ $breadcrumb-chevron-height: 0.65rem;
         color: var(--ons-color-text);
         text-decoration: underline;
 
+        .ons-breadcrumbs-wrapper--dark &,
+        .ons-breadcrumbs-wrapper--navy-blue & {
```
rmccar, line 111: You could probably refactor this css so you don't have to keep repeating this

src/components/breadcrumbs/_breadcrumbs.scss
```diff
@@ -1,5 +1,33 @@
...
+    &--dark {
+        background-color: var(--ons-color-hero-bg-dark);
+    }
+
+    &--navy-blue {
+        background-color: var(--ons-color-navy-blue-light);
+        margin-bottom: -1.75rem;
+        @include mq(l) {
+            margin-bottom: -3.25rem;
+        }
```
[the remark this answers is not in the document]
rmccar: You would use it like this `@extend .ons-u-mb-no;` but in this case Ive realised you cant actually use them because you are using minus padding

precious-onyenaucheya-ons (approved):

PR #3709, opened 2025-08-21
Introduce aspect ratio param for iframe charts

precious-onyenaucheya-ons:

Addresses [ONSDESYS-608](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-608) and [ONSDESYS-569](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-569)

This PR updates the iframe chart macro and associated SCSS to allow users to set the aspect ratio for iframe charts. A default aspect ratio of 16/9 is applied if no value is provided or if the user-provided value is not in the list of accepted ratios. A SCSS mixin has been added to generate classes for the allowed ratios.

Refactored the tests as part of this as well

### How to review this PR

1. Check that the default iframe renders correctly with the default aspect ratio when no aspect ratio is provided and when the provided value is not in the allowed list.
2. Test overriding the default aspect ratio using aspectRatio='4/3' or aspectRatio='21/9'.
3. Ensure that the chart or fallback image scales correctly according to the aspect ratio.
4. Verify that the documentation reflects how to use the aspectRatio parameter.

### Checklist


src/components/chart/_macro-options.md
```diff
@@ -28,6 +28,7 @@
...
 | fallbackImageAlt         | string              | false                  | Alt text for the fallback image - a short description only, as the description field describes the chart for screen readers. Defaults to "Fallback image for the chart as JavaScript is disabled"                                                                                                                                                   |
+| iframeAspectRatio        | string              | false                  | Sets the aspect ratio for iframe charts. Only applicable when using an iframe chart. Acceptable values are 16-9, 4-3, 21-9, 1-1, 3-2, and 9-16. The value must be provided in the x-y format (e.g. 16-9). If an unsupported value is provided, the aspect ratio will default to 16-9. Defaults to 16-9 when not specified.                          |
```
rmccar, line 31: I think this could be reworded to something like "Sets the aspect ratio for iframe charts. Only applicable when using an iframe chart. Acceptable values are 16-9, 4-3, 21-9, 1-1, 3-2, and 9-16. The value must be provided in the x-y format (e.g. 16-9). If an unsupported value is provided or a value is not provided, the aspect ratio will default to 16-9."
This avoids the need to say it defaults to 16-9 twice

src/components/chart/_macro.spec.js
```diff
@@ -1568,9 +1568,16 @@ describe('Macro: Chart', () => {
                     expect($('.ons-chart__iframe-wrapper').length).toBe(1);
                 });
 
-                test('THEN: it includes the iframe title as a data attribute', () => {
-                    const iframe = $('.ons-chart__iframe-wrapper');
-                    expect(iframe.attr('data-title')).toBe(EXAMPLE_IFRAME_CHART_PARAMS.title);
+                test('THEN: it includes the aspect ratio class if set', () => {
```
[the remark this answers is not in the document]
precious-onyenaucheya-ons: I have now added the tests

src/components/chart/_macro.spec.js
```diff
@@ -1580,6 +1580,27 @@ describe('Macro: Chart', () => {
...
+                    const $ = cheerio.load(
+                        renderComponent('chart', {
+                            ...EXAMPLE_IFRAME_CHART_PARAMS,
+                        }),
+                    );
+                    const iframe = $('.ons-chart__iframe');
+                    expect(iframe.hasClass('ons-chart__iframe--16-9')).toBe(true);
+                });
+
+                test('THEN: the aspect ratio defaults to 16-9 when unsupported value is set', () => {
```
rmccar, line 1593: I think the descriptions of these tests need to be rewritten to match our agreed style and it might mean that there is a bit of a refactor because you shouldn't need to add a when on the end of a then.

It should be something like:
WHEN: iframeAspectRatio is not set
THEN: the aspect ratio defaults to 16-9

WHEN: iframeAspectRatio is passed an unsupported value
THEN: the aspect ratio defaults to 16-9

But to get this structure you might need to change how the given and whens higher up are structured. Usually the given would have the name of the param or have "Params: required" then under that it would have all the basic tests to make sure the functionality works. Then under separate givens you would have each optional param. This frees up the whens to have the above structure Ive suggested. In this case you have all the tests for the iframe under the same given.

src/components/chart/example-iframe-chart.njk
```diff
@@ -27,6 +27,7 @@ may not be available in the future once Florence is decommissioned. #}
                 }
             ]
         },
+        "iframeAspectRatio": "16-10",
```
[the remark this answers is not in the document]
precious-onyenaucheya-ons: Added this for testing, will remove this

rmccar (changes requested):
Just a few more structural and minor fixes I noticed to make it more consistant

src/components/chart/_macro.spec.js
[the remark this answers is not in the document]
precious-onyenaucheya-ons: updated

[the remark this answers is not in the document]
precious-onyenaucheya-ons: updated

rmccar (changes requested):
Few more things I noticed

src/components/chart/_macro.spec.js
```diff
@@ -545,61 +913,199 @@ describe('Macro: Chart', () => {
...
-                    expect($('.ons-chart__download-title').length).toBe(0);
+                test('THEN: the footnotes are hidden from screen readers', () => {
+                    expect($('.ons-chart__annotations-footnotes').attr('aria-hidden')).toBe('true');
                 });
 
-                test('THEN: it includes the Highcharts JSON config', () => {
-                    const configScript = $(`script[data-highcharts-config--column-chart-123]`).html();
-                    expect(configScript).toContain('"text":"X Axis Title"');
-                    expect(configScript).toContain('"text":"Y Axis Title"');
+                test('THEN: it includes the Annotations JSON config', () => {
```
rmccar, line 934:
```suggestion
                test('THEN: it includes the annotations JSON config', () => {
```

src/components/chart/_macro.spec.js
rmccar, line 1: Line 1220 should be "test('THEN: it includes the annotations JSON config', () => {"

src/components/chart/_macro.spec.js
```diff
@@ -1298,443 +2009,240 @@ describe('Macro: Chart', () => {
         });
     });
 
-    describe('FOR: Line chart with range annotation on the x axis', () => {
-        describe('GIVEN: Params: Range annotations', () => {
-            describe('WHEN: range annotations params are provided', () => {
-                const $ = cheerio.load(renderComponent('chart', EXAMPLE_LINE_CHART_WITH_RANGE_ANNOTATION_ON_X_AXIS_PARAMS));
+    describe('FOR: Iframe Chart', () => {
+        describe('GIVEN: Params: required', () => {
+            describe('WHEN: iframe url param is provided', () => {
```
rmccar, line 2014:
```suggestion
            describe('WHEN: iframeUrl param is provided', () => {
```

rmccar (approved):

PR #3737, opened 2025-09-26
Fix pagination accessibility issue

SriHV:

[ONSDESYS-472](https://officefornationalstatistics.atlassian.net/jira/software/c/projects/ONSDESYS/boards/2876?selectedIssue=ONSDESYS-472)

As per the ticket:

- Removed "Go to"  from "aria-label" attributes.

- When "aria-current" is set to true a default "aria-label" will only have the text set in position variable which is `"Page " ~ currentPageIndex ~ " of " ~ totalPages`(Eg When aria-current is over the 3rd page, voice over only reads `Page 3 of 4` instead of previous `Current Page(Page 3 of 4)`) 

- The ticket also mentions about adding "aria-hidden" to svg icons. We have this already so we don't need to make any changes to it 

Also user who raised this issue used total validator to check the issue. As total validator works with a paid tier. I've used [wave](https://chromewebstore.google.com/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh) to test the page. 

### How to review this PR

- Use Voice over and check that everything mentioned above is reflected and is as per the requirement 
- Use wave extension to check that there are no issues

### Checklist


src/components/icon/_macro.njk
```diff
@@ -835,7 +837,14 @@
...
         </svg>
     {%- elif params.iconType == "census-logo" -%}
-        <svg xmlns="http://www.w3.org/2000/svg" class="ons-icon--logo{{ iconClasses }}" width="152" height="16" viewBox="0 0 171 18">
+        <svg
+            xmlns="http://www.w3.org/2000/svg"
+            class="ons-icon--logo{{ iconClasses }}"
+            width="152"
+            height="16"
+            viewBox="0 0 171 18"
+            aria-hidden="true"
```
rmccar, line 846:
```suggestion
            aria-hidden="true"
            focusable="false"
```
I think we also need to add this here

rmccar (approved):

PR #3748, opened 2025-10-13
Upgrade Node to v24

SriHV:

[ONSDESYS-646](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-646)

- Updated Node to V24.10.0

- After upgrading, `yarn start` started throwing module resolution errors for imports like /design-system/lib/render-page-list.

- Updated imports to use proper relative paths (e.g. ./render-page-list.js)

- Updated `gulp` and `gulp-cli` to latest versions to resolve `(node:14077) [DEP0180] DeprecationWarning: fs.Stats constructor is deprecated.`

-  Deprecation warning - `(node:17155) [DEP0060] DeprecationWarning: The `util._extend` API is deprecated. Please use Object.assign() instead.` Found out `util` is used  by `http_proxy` a transitive dependancy brought in by `browser-sync` 
The issue is still open in that dependancy

- Warning:
`(node:8229) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///design-system/lib/dev-server.js is not specified and it doesn't parse as CommonJS. Reparsing as ES module because module syntax was detected. This incurs a performance overhead. To eliminate this warning, add "type": "module" to /design-system/package.json.`
This warning occurs because our JavaScript files currently use CommonJS syntax (e.g., require('<package_name>')) and ESM (ECMAScript Module) syntax (e.g., import ... from '<package_name>'). Node.js needs to know which module system to use when loading files. Since it detects ESM syntax without explicit configuration, it reparses the file, which causes this warning and a slight performance overhead. To resolve this, we should standardize on a single module system — either CommonJS or ESM — and configure it accordingly (e.g., by adding "type": "module" to package.json if we choose ESM. If we add it we need to resolve this error shown during `yarn start` - `ReferenceError: require is not defined in ES module scope, you can use import instead This file is being treated as an ES module because it has a '.js' file extension and '/design-system/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension`.)

These warning show up in node version 22 as well.

### How to review this PR

Run `nvm install` and then `nvm use` to update node to `v24.10.0` and test DS with `yarn start`, `yarn test`, `yarn test-visual` and check that there are no deprecations or errors

### Checklist

[ONSDESYS-646]: https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sveltifier:
Duplicate PR

PR #3755, opened 2025-10-31
Add aria-label to Accordion Button and update documentation

precious-onyenaucheya-ons:

https://github.com/ONSdigital/design-system/discussions/1377

This PR addresses the accessibility issue with accordion “Show all” and “Hide all” buttons having identical labels, which caused ambiguity for screen reader users when multiple accordion components were present on the same page.

To resolve this, two new optional parameters have been added:

- openAriaLabel – Defines a custom aria-label for the “Show all” button.
- closeAriaLabel – Defines a custom aria-label for the “Hide all” button.

These parameters allow each accordion instance to have unique and descriptive aria labels, providing better context for assistive technologies.
Default values are also included in case these parameters are not set.

### How to review this PR

- Open or create an accordion component that uses the allButton configuration.
- Add the new parameters to the configuration, for example:

```
"allButton": {
  "open": "Show all",
  "close": "Hide all",
  "openAriaLabel": "Show all glossary terms",
  "closeAriaLabel": "Hide all glossary terms"
}

```
- Inspect the rendered “Show all” and “Hide all” buttons in the HTML.
- Confirm that each button includes the correct aria-label values.
- Remove the parameters and confirm that the default aria-labels are still applied.
- Test with multiple accordion components on the same page to verify that each set of buttons can have unique aria-labels and are distinguishable by screen readers.

### Checklist


sveltifier (commented):
Great!

PR #3787, opened 2026-02-23
Add support for vertical alignment in Table cells

SriHV:
[ONSDESYS-805](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-805)

Added two new parameter `valign` which will set the vertical alignment of table cells to top, center or bottom

By default if these parameters are not provided, the alignment of the cells stays to the top.

### How to review this PR

Check the new examples created and check vertical alignment changes when the parameter options are set to top, center or bottom 

### Checklist

[ONSDESYS-805]: https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__snapshots__/layout/_template.spec.js.snap
+++ b/__snapshots__/layout/_template.spec.js.snap
@@ -1650,7 +1650,11 @@ exports[`base page template matches the full configuration snapshot 1`] = `
                     </div>
                 
                 <div class="ons-grid__col ons-col-auto ons-u-flex-shrink">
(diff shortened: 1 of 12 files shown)


Dino-ONS:
Just a suggestion... Maybe keep the vertical-align: top as the base default in the CSS and only override it when a parameter is present? Otherwise it may break existing clients that do not have a parameter setting and are using HTML instead. Also, there is no input validation and the alignment value is concatenated directly into the class name, so if someone passes an unexpected value, it will create an invalid class.

PR #3803, opened 2026-03-25
Update highlighting examples and styles for improved visual consistency

Dino-ONS:

Replaced the linear-gradient percentage approach in .ons-highlight with a background-size/background-position technique, removing the need for per-heading gradient overrides.
Added --ons-line-height CSS custom property to the font-size mixin in _typography.scss, which sets and updates the value at each breakpoint using the existing $type-matrix line-height values.
Updated .ons-highlight to use background-size: 100% calc(var(--ons-line-height, 28px) - 4px) so the yellow background always leaves exactly a 4px transparent gap — matching the box-shadow y-offset — regardless of heading level or viewport size.
Removed all per-heading SCSS overrides and breakpoint-specific gradient rules from _highlight.scss.
Refinement to PR [ONSDESYS-809](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-809)

### How to review this PR

Verify highlighted `<strong>` text that wraps across multiple lines renders a consistent yellow highlight on every line (no 2-tone effect).
Verify the black underline (box-shadow) appears on each wrapped line, not just the last line - foundations/highlighting/example-highlighting.html

### Checklist

[ONSDESYS-809]: https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rmccar (commented):
Few small things I spotted. Other than these it looks good

src/scss/utilities/_highlight.scss
```diff
@@ -1,9 +1,11 @@
 .ons-highlight {
+    background-color: transparent; // Overrides stylesheet default on <mark> to prevent a 2-tone effect behind the gradient
     background-image: linear-gradient(var(--ons-color-highlight), var(--ons-color-highlight));
     background-position: bottom;
     background-repeat: no-repeat;
     background-size: 100% calc(var(--ons-line-height, 28px) - 4px);
-    box-decoration-break: clone;
+    box-decoration-break: clone; // Safari/Chrome
+    box-decoration-break: clone; // Firefox
```
rmccar: I don't think we need this duplication here, we can remove line 7

SriHV (approved):
Looks good to me! May just need to check up on Rich’s comments regarding updating the VR tests.

PR #3825, opened 2026-04-21
Ensure that both the cookie banner and the skip link content are rendered inside landmark elements

helenb:

This is a small accessibility / semantic improvement to the cookie banner and skip links, which both sit outside of the `<header>` and `<main>` tags.

It relates to the requirement that all page content should be contained within a landmark - see https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11.html, particularly the lines "It is a best practice to include all content on the page in landmarks, so that screen reader users who rely on them to navigate from section to section do not lose track of content.".

Technically the cookie banner is already wrapped in a landmark, thanks to the use of `role="region"`, however it's best practice to use semantic HTML tags rather than use aria roles where possible - the `section` tag has an implied role of `region` when it has an accessible name provided by `aria-labelledby` or `aria-label`. See https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/. The cookie banner best suits the role of 'banner' as this is content that represents site level content.

The skip link requires an extra wrapping element because a link cannot be a landmark element. I have used a div with a region role, as this did not feel like site level content - open to feedback on that one.

### How to review this PR

On your local build, review the cookie banner and skip link, and check that there are no visual changes (VR testing in CI should confirm that). Review the code changes and check the W3 docs linked above to confirm you are happy with the semantic change.

Relates to this CMS issue: https://officefornationalstatistics.atlassian.net/browse/CMS-1169

### Checklist


helenb:
Converting to draft while I review the failed CI tasks.

helenb:
@sveltifier would you be able to create a jira ticket for this one so I can link to it from this PR?

helenb:
> Converting to draft while I review the failed CI tasks.

VR tests have passed now so the initial failure was probably some kind of temporary CI issue.

SriHV (approved):

helenb:
@SriHV following the discussion in stand-up I've merged in the `main` branch - do you now need to re-review?

SriHV:
> @SriHV following the discussion in stand-up I've merged in the `main` branch - do you now need to re-review?

Looks good to me, thanks @helenb 👍

PR #3831, opened 2026-05-07
Upgrade to Node 24

rmccar:

[ONSDESYS-646](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-646)

This PR upgrades the DS to Node 24. To do this there were some breaking changes that needed to be addressed. Mainly to do with the new changes in Node 22 for the importing of CommonJS and Module JavaScript files in scripts and how they are handled.

To do this some `.js` files were renamed to have a `.mjs` extension. These were being imported like Modules but were not called out as such. This change allows Node to see that they are modules and to import them without having to guess. The imports were updated to also reflect this change.

The jest config also needed updating to use babel to convert any scripts that it needed to in order to run them when running the tests.

This upgrade also caused some deprecation warnings to be flagged up. To solve these Gulp and Stylelint needed updating.

This upgrade also caused a lot of indentation warnings to be flagged up when running `yarn stylelint` which were false positives. To fix this I have added the `stylistic` plugin and set indentation to 4 spaces in the config.

There is a new deprecation warning after this change:
"(node:50537) [DEP0060] DeprecationWarning: The `util._extend` API is deprecated. Please use Object.assign() instead."
This comes from browser-sync@3.0.4 → http-proxy@1.18.1. Both are already at their latest published versions and http-proxy has not released a fix. The only way to eliminate it would be replacing `browser-sync` with an alternative dev server. This would have to be treated as a separate piece of work but at the moment this is just a warning that this could be an issue in future Node versions. I don't think its a reason not to upgrade.

### How to review this PR

- Check out the branch and run `nvm use` to switch to the correct version of Node (you may need to install this first using `nvm install v24.15.0`)
- Run the usual commands to make sure everything works as expected `yarn start`, `yarn test`, `yarn test-visual`
- Check that the DS still works, spins up as expected and all components still work and render as expected
- Make sure all the changes make sense to be making
(When switching back and fore between Node versions you may need to delete your local `node_modules` folder and re run `yarn` if you get errors when spinning up the DS)
- Run `yarn stylelint` and see that there are no warnings or errors

### Checklist

[ONSDESYS-646]: https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


helenb:
Is this a duplicate of https://github.com/ONSdigital/design-system/pull/3748?

rmccar:
> Is this a duplicate of #3748?

It is, but that PR still has some issues which this one resolves. For example the warning that is mentioned in that PR is resolved in this PR and so is the stylelint issue that Ive mentioned from the upgrade is also resolved.

sveltifier:
I just noticed this is still here - it's not on the DS board, but I'll request it gets added to the next sprint!

rmccar:
> I just noticed this is still here - it's not on the DS board, but I'll request it gets added to the next sprint!

Cool, didn't notice that, I put the card into review but maybe its not been added to the sprint

sveltifier:
Feels like we should include this in a major release? Would you say this is strictly a breaking change for DS?

rmccar:
> Feels like we should include this in a major release? Would you say this is strictly a breaking change for DS?

Yeah I can see your logic on that and why you might want to include it in a major release, which you could do but I wouldn't class this change as a breaking change. This should only affect the spinning up of the DS for development. Any service using the DS should see no difference and would not need to make any changes their side to upgrade, if there were changes needed then I would class that as a breaking change. The main changes here are to the js files all on the dev side, once the DS goes through the build process then the services themselves still just see and import a `main.js` file before and after this change. Therefore I would keep major releases for just breaking changes and not release this in a major release. But its up to you how you want to go about it, these are just my thoughts on it and what I would class as a major release in the context of the DS. There should be some ways of working documents in confluence on this. Maybe they need a review and a bit more clarification on what constitutes different types of releases just so you can provide a consistent service to the users of the DS. Thats what I was trying to work on while I was working on it. Let me know if there are any more questions.

andy-mair-ons (approved):
LGTM manually tested across multiple components locally all rendering as expected. No issues with `yarn test` or `yarn test-visual`. There are 16 warnings running `yarn stylelint` but no errors

PR #3834, opened 2026-05-07
Update deprecated css property names to modern replacements

rmccar:

Fixes: [ONSDESYS-895](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-895)

There are some attributes that we are using that are now deprecated and have replacements. I have replaced these in this PR.

word-wrap > overflow-wrap (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-wrap)
grid-column-gap > column-gap (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/column-gap)
WindowText > CanvasText (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/system-color)

Also fixes some Stylelint issues (duplicate css selectors, unnecessary selectors etc) and removes the line setting Stylelint errors to warnings. This line could be problematic because it is silencing any errors raised by Stylelint when the pre-commit hook runs. This means any css linting errors aren't being caught. If we have some linting errors that we want to ignore like this then we should do it on a rule by rule basis or change the linting rules in the Stylelint config.

### How to review this PR

VR tests pass without need for changes

### Checklist

[ONSDESYS-895]: https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andy-mair-ons (approved):
LGTM all VR tests pass

PR #3824, opened 2026-04-20
Improve landmark semantics in Layout component

helenb:
### What is the context of this PR?

As part of some semantic improvements to the structure of the site, we need to make sure that all content in the site is contained inside landmark elements (`main`, `header`, `banner` etc).

See https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-864

This pull request
- Moves the `main` tag higher up the tree
- Move breadcrumbs to sit inside the header tags rather than being part of the main landmark
- Render <header> tags in the base page template to enclose both the header and breadcrumbs components
- Add  `fullWidth` and `wide` variants to the breadcrumb component, because the breadcrumb is no longer surrounded by the page content wrapper classes
- Adjust the breadcrumbs CSS to account for the wrapper classes being rendered within the breadcrumbs component rather than outside it
- Ensure the breadcrumb 'variant' option is used if be set in pageConfig
- Update the header component so it can optionally use a div as the root element via a new option called `useDivForRootElement`
- In the base template, render the header component inside `<header>` tags and set `useDivForRootElement` to be true
- Create a block for the hero so that it can be overridden without having to override other markup surrounding it
- Move the `preMain` block so it sits above the newly positioned main tag
- Rename the `main` block to `mainContent` as it no longer relates to the `main` tag.
- Move the `main-content` id to the `main` tag so the skip link doesn't skip the hero if used
- Various CSS tweaks to ensure there are no issues from the markup changes
- Update the readme to assist with VR tests 

### How to review this PR

Describe the steps required to test the changes (include screenshots if appropriate).

The easiest way to review this is commit by commit, as there are quite a lot of changes - but many of them are simply updating examples to use the new blocks

The key commits to check are:
- def9531a28bb016f1d1cd9236a6a5892d12491e9 - this was the first commit, which moved the `main` tag and the `preMain` block. The change to the cookie banner in this commit was subsequently reverted because it happened in a different PR
- 511525495215f450e7d2cfc23e949fd79bea86be - this commit moves the breadcrumbs, renames the `main` block, updates the breadcrumbs to add the new variants, and updates the header to use the new `useDivForRootElement` option.
- 9b1dca4f0e6a2cbfb10524ba6c8dbad57e3f46c8 this commit updates all the examples to use the new block structure
- b4aa26ae2b6aeab167e08fa8ed0add2d3f5a2fed - add tests for the new parameters 
- The remaining commits catch bugs and issues with the CSS
[...]

PR #3837, opened 2026-05-13
Support full-width Hero component and add examples

helenb:

Adds a new parameter - 'fullWidth' to the hero component, and adds some extra examples of how to render the hero full width, including in the context of a page example.

See https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-897

Note that because the hero is not included from the base page template, the description in the ticket is actually not correct. We can't pass down the pageConfig.wide and pageConfig.fullWidth options to the hero, as it is only included by pages that extend the base page template.

This PR is built off https://github.com/ONSdigital/design-system/pull/3824 so that the examples can use the new 'hero' block.

### How to review this PR

Review the new examples to check everything renders as expected.

### Checklist


andy-mair-ons (approved):
LGTM all renders correctly locally. Tested at multiple screen sizes locally

PR #3838, opened 2026-05-14
Bundle charts js separately

rmccar:

This PR removes the charts js out of main.js bundle that is created when the DS is built. Because the js for the charts component is large, leaving it where it is could affect the performance of webpages where the charts component isnt even used. This charts js has been moved out into a chart.js file which will only be downloaded on a page where the charts component is used.

### How to review this PR

- On main spin up the DS, go to an example and check the dev tools network tab for js downloads and make a note of the size of main.js
- Then spin up this branch and check the size of the main.js and see that it has now substantially reduced
- Check the network tab on both a charts example and a non charts example. See that the chart.js is now only loaded on the charts examples
- Run `yarn cdn-bundle` and check your local repo for a new build folder and the new chart.js file is created in there

### Checklist


helenb:
I have tested this and it is working well - nice work. LGTM. I don't think I have permission to add an official review currently.

helenb (approved):
@rmccar I found I was able to approve after all. @sveltifier we can still look through together when we catch up later.

rmccar:
> @rmccar I found I was able to approve after all. @sveltifier we can still look through together when we catch up later.

Thanks! If you want to have a chat about it later ill hold off merging 👍

sveltifier (approved):

PR #3840, opened 2026-05-18
Update the table of contents component so that it includes an aria label for the aside element

helenb:

See https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-888

All landmarks on a page should be unique in order to aid users with screen readers.

This is only an issue if there is another non-unique aside element on the page, but this happens in the CMS if a user is logged in to Wagtail. In this scenario, there is an ‘edit’ link inserted into the page which also uses a non-unique aside. This has been resolved in Wagtail, and this PR also resolves it in the DS.

- Keeps the existing `ariaLabel` parameter on the `nav` element
- Adds a new `ariaLabelForAside parameter` for the `aside` element.
- Adds tests for the new parameter.

Note that both aria-labels default to "Table of contents". This seemed the most meaningful default for both, and there are no automatic issues reported by Axe when they are both the same.

I tested in Voice Over using the default labels, and it announces them as follows:

"Table of contents, complementary. You are currently in a complementary".

If you then enter the 'aside' it announces "Table of contents, navigation. You are currently in a navigation."

### How to review this PR

Test in your local build with a screen reader, and with an automated checker such as Axe or pa11y.

### Checklist


andy-mair-ons (approved):
LGTM axe tests passed and manual screen reader testing in browser confirm expected behaviour

SriHV (approved):

PR #3842, opened 2026-05-28
Table of Contents ordered & unordered variants

helenb:

See ticket https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-887

- Use an unordered list for the table of contents by default, reflecting the fact that the list is visually un-ordered
- Add a new option to render the table of contents with a visually ordered list, which may become a requirement for the CMS 

### How to review this PR

- Review the table of contents examples
- Check that they are now rendered with a `<ul>` but remain unchanged visually
- Review the new examples showing the ordered table of contents
- Verify the ordered table of contents examples are rendered with an `<ol>`

Screenshot of the ordered example:
 
Screenshot 2026-05-28 at 10 56 57

### Checklist

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


helenb:
We discussed the design of the ordered table of contents in stand-up, and whether we should reduce the padding so that the numbers align on the left with the heading. However, if we go into double figures, the numbers are right-aligned (which is better for legibility) so we'd end up with an overlap the other way. I propose we leave it with the default styling for now, and we can discuss with the designers if and when we need the ordered list in the CMS.

Screenshot 2026-06-09 at 13 01 51

helenb:
> We discussed the design of the ordered table of contents in stand-up, and whether we should reduce the padding so that the numbers align on the left with the heading. However, if we go into double figures, the numbers are right-aligned (which is better for legibility) so we'd end up with an overlap the other way. I propose we leave it with the default styling for now, and we can discuss with the designers if and when we need the ordered list in the CMS.

I've now discussed this with the designers and they are happy with the alignment we have here.

sveltifier (approved):

PR #3843, opened 2026-06-01
Fix and improve Table aria-labels

helenb:

See https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-892

This addresses an issue whereby multiple tables on the same page could end up with identical aria-labels, which would be an accessibility fail. In addition, it fixes an issue whereby if no caption was supplied, the aria-label would start with '. ' which is confusing.

- caption is no longer part of the aria label.
- Remove reference to 'scrollable table variant' in macro options, as all tables are now scrollable
- Added an example of a table with a hidden caption as an alternative way of visually describing the table for screen readers

### How to review this PR

- Review the new example tables
- Check the code changes and rendered aria-labels

### Checklist


src/components/table/_macro.njk
```diff
@@ -8,7 +8,7 @@
             class="ons-table-scrollable__content"
             tabindex="0"
             role="region"
-            aria-label="{{ params.caption }}. {{ params.ariaLabel | default('Scrollable table') }}"
+            aria-label="{% if params.caption %}{{ params.caption }}.{% endif %}{{ params.ariaLabel }}"
```
helenb: @SriHV @sveltifier do we want to keep the caption as part of the aria-label? I thought it might have some benefit to ensure the aria-label reflected the visible caption, but actually the screen reader will read the caption anyway, so this feels repetitive. If we remove it I'll need to update my PR on the CMS side but that's fine.

SriHV (dismissed):

src/components/table/_macro.njk
SriHV: As screen reader is reading the caption twice, should we remove it?

helenb: I think that it is probably best to remove it - to be clear though, just from the aria-label, not altogether.

helenb: @SriHV I've updated that now.

helenb:
> #### ️✅ There are no secrets present in this pull request anymore.
> If these secrets were true positive and are still valid, we highly recommend you to revoke them. While these secrets were previously flagged, we no longer have a reference to the specific commits where they were detected. Once a secret has been leaked into a git repository, you should consider it compromised, even if it was deleted immediately. Find [here](https://docs.gitguardian.com/platform/remediate/remediate-incidents) more information about risks.
> 
> 🦉 [GitGuardian](https://dashboard.gitguardian.com/auth/login/?utm_medium=checkruns&utm_source=github&utm_campaign=cr1) detects secrets in your source code to help developers and security teams secure the modern development process. You are seeing this because you or someone else with access to this repository has authorized GitGuardian to scan your pull request.

I think this was because I accidentally committed a `.venv` folder - after I did that I removed the commit it was in and force pushed.

SriHV:
Looks Good to me, thanks @helenb 👍

helenb:
Thanks @SriHV. I'll hold off merging as this is a breaking change.

helenb:
@SriHV and @sveltifier a minor comment: at the moment the label checker is failing for this PR, because it doesn't like the fact I've marked this as both 'accessibility' and 'breaking change' - I'm not sure why it should be opinionated about only having one of these labels - this PR genuinely relates to both.

sveltifier:
> @SriHV and @sveltifier a minor comment: at the moment the label checker is failing for this PR, because it doesn't like the fact I've marked this as both 'accessibility' and 'breaking change' - I'm not sure why it should be opinionated about only having one of these labels - this PR genuinely relates to both.

Yep it's exactly one label. My guess is that it's to support the Release Notes -- the label here is the heading under which the change is listed for documenting our release, and it can therefore only go under one heading or another.

src/components/table/_macro-options.md
```diff
@@ -5,7 +5,7 @@
...
 | caption      | string          | false                                 | The caption for the table component                                                                                                                                                                      |
 | hideCaption  | boolean         | false                                 | Set to “true” to visually hide the caption                                                                                                                                                               |
-| ariaLabel    | string          | false                                 | The ARIA label to be added if ”scrollable” variant set, to inform screen reader users that the table can be scrolled. Defaults to “Scrollable table“.                                                    |
+| ariaLabel    | string          | false                                 | The ARIA label for the table."                                                                                                                                                                           |
```
sveltifier: Extra `"`? Perhaps add that this isn't required

sveltifier (changes requested):

src/components/table/_macro-options.md
helenb: Updated in https://github.com/ONSdigital/design-system/pull/3843/changes/554ae8737012de729b22683dc36634a921fa2062

sveltifier (approved):

PR #3844, opened 2026-06-01
Fix dynamic island bug on iOS Edge

andy-mair-ons:
https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-876

- Adds viewport-fit="cover" to meta tag content fills edge-to-edge
- Adds the insets using CSS environment variables to control padding around hardware obstructions 

### How to review this PR

Check code works across browsers on iOS, android, mac and windows 

### Checklist


SriHV (approved):
Tested in Android. Working as expected!

src/components/header/_header.scss
```diff
@@ -367,6 +367,14 @@
             width: 100vw;
```
sveltifier: Try 100% or something and see if we can get rid of the extra special-case rule below

sveltifier (changes requested):

src/components/header/_header.scss
andy-mair-ons: Have converted the units to % and it should resolve the issue without the special case class

sveltifier:
Updated PR title to make a better release note.

src/layout/_template.njk
```diff
@@ -62,7 +62,7 @@
     <head>
         <meta charset="utf-8" />
         <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-        <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <meta name="viewport" content="width=device-width, initial-scale=1" viewport-fit="cover"/>
```
sveltifier: I believe this is invalid HTML, and therefore ignored. 


<meta name="viewport" content="width=device-width, initial-scale=1" viewport-fit="cover"/>


<meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover"/>

If so this can & should be removed.

sveltifier: <meta name="viewport" content="width=device-width, initial-scale=1" viewport-fit="cover"/>

    
    <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover"/>

sveltifier (changes requested):

src/components/header/_header.scss
```diff
@@ -364,7 +364,7 @@
         .ons-js-enabled & {
             position: relative;
             left: 50%;
-            width: 100vw;
+            width: 100%;
```
sveltifier, line 366: I think that 100% + left: 50% + translateX(-50%) is internally contradictory. When width: 100% equals the container's width, left: 50% shifts it off-centre and translateX(-50%) brings it back, net effect is zero movement. Why was this change necessary again? Chat if needed.

sveltifier (dismissed):

sveltifier:
Also, I tested on iOS, and the bug is fixed in Edge, and everything looked ok in Chrome and Safari too, so I think the CSS rule is working.

src/components/header/_header.scss
andy-mair-ons: Had to change from vw to % to avoid the header overflowing the safe-area-inset paddings instead of having a custom edge case class. The left 50% and translateX(-50%) was there originally and is a common centring technique. Have tested without it across all the example headers and doesn't seem to be doing anything so have removed it

PR #3848, opened 2026-06-11
Figure proof of concept

helenb:

https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-910
https://officefornationalstatistics.atlassian.net/browse/DPD-4848

This is a spike pull request for reference only.

The purpose is to validate that we could create a figure component which can use `caller` to nest other components. It could also be used in the CMS to nest other content, such as an iframed chart if and when that moves out of the design system.

This would be a breaking change because it would remove contextual information such as caption, heading and subtitle from the chart and image components, as these would be rendered by the figure in future.

This merge request also updates the featured-article (used by the CMS on a topic page) so that it too can use `caller()`  to include the figure component and then the image or chart components, rather than explicitly calling them. This would then allow the CMS to embed an iframed chart in the featured article if and when the iframed chart moves to the CMS.

Note that I have not updated the tests for the sake of the spike, and there are multiple VR tests that I would expect to fail due to changes to the image and chart components.

**Next steps:**

We need designs

We need to discuss the possible multiple ways of describing an image as a result of these changes - accessible description on the figure, visible caption, alt text etc with the datavis team.

We would need to decide how to handle if the image or chart were ever called without the figure component, or whether we should advise against that.

### How to review this PR

### Checklist


helenb:
Closing in favour of https://github.com/ONSdigital/design-system/pull/3879

PR #3854, opened 2026-06-17
Improve docs and examples for Table column widths

SriHV:

[ONSDESYS-806](https://officefornationalstatistics.atlassian.net/jira/software/c/projects/ONSDESYS/boards/2876?selectedIssue=ONSDESYS-806)

As per the ticket, we need to add column width parameter. As we already have the parameter around setting width of the table. We don't need to create a new one.

This PR updates the documentation of the `width` and example of table with numeric cells for better understanding of column width.

### How to review this PR

Check that description about width is understandable.

### Checklist

[ONSDESYS-806]: https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andy-mair-ons (dismissed):
LGTM

package.json
```diff
@@ -122,6 +122,7 @@
         "remark-cli": "12.0.0",
         "remark-lint": "9.1.2",
         "remark-preset-lint-recommended": "6.1.3",
+        "serve": "^14.2.6",
```
sveltifier: I think we shouldn't include changes to the packages /  lockfile in a documentation PR. I would make a separate PR, but there's a solution to this in the `next` branch anyway, and it avoids taking an unnecessary dependency.

> `npx http-server backstop_data/html_report -p 8080` - View the results of the yarn visual tests in your browser. Using http-server rather than simply opening the file ensures the associated JavaScript runs. 

So let's remove these changes here.

sveltifier (changes requested):

src/components/table/_macro-options.md
```diff
@@ -16,16 +16,16 @@
...
+| thClasses       | string  | false    | Classes to add to the `th` element                                                                                                                                      |
+| ariaSort        | string  | false    | Set to “ascending” or “descending” to set the default order of a table column when the page loads when setting `variants` to “sortable”. Defaults to “none”.            |
+| value           | string  | true     | The content for the `th` cell                                                                                                                                           |
+| numeric         | boolean | false    | Set to “true” if all the cells in the column contain numbers. This aligns the content to the right so numbers can easily be compared.                                   |
+| widthPercentage | integer | false    | Sets the `width` attribute (as a percentage) on the `th`, which controls the width of the whole column. Any column left without one sizes automatically to its content. |
```
sveltifier, line 25: I'm not sure what this means:

> Any column left without one sizes automatically to its content.

sveltifier: ...is this just describing normal HTML table behaviour? If so I think we should remove this, as it implies to me that we do something additional to normal HTML behaviour.

sveltifier (changes requested):

package.json
helenb: @sveltifier I think we might need @SriHV's changes here, because that method was is longer working for me since the upgrade to node 24. The images aren't saved so when I view the diffs I can see the number of failures, but not the changes. So we might need to just separate out the changes here to a new branch.

helenb: @sveltifier @SriHV just wondering if there is any update on getting the fixes for running the VR test locally that are in this PR merged? Could we separate them out to another branch? I think I should probably remove the changes I made to the readme in the `next` branch.

PR #3876, opened 2026-06-30
Add preBreadcrumbs and postHero slots to Layout

helenb:
Follow-on from [Jira issue](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-864)

I have been testing out the pre-release for 74.0.0 in the CMS and discovered a couple of places where we need extra blocks in order to slot in content. One needs to come after the hero, rendered full-width - I've added this as `postHero`. Then another is to slot between the onsHeader content and the breadcrumbs - I've called this one `preBreadcrumbs`. In theory we could just use the existing `hero` and `breadcrumbs` blocks but this feels more semantic and hopefully gives a bit more flexibility.

### How to review

Describe the steps to review and test these changes.

### Checklist

- [x] I have linked the Issue


src/foundations/page-template/example-base-page-template-block-areas.njk
```diff
@@ -134,6 +139,14 @@
     <div class="ons-ds-block">
         <div class="ons-ds-block__label">block: pageContent</div>
 
+        <div class="ons-ds-block">
```
helenb, line 142: This was an omission in the previous pr.

sveltifier (approved):

helenb (commented):
@sveltifier your changes LGTM

PR #3851, opened 2026-06-15
Figure proof of concept - alternative approach where onsChart and onsImage call the figure component to render

helenb:
### What is the context of this PR?

This builds on #3848 with an alternative approach. Instead of leaving it to consumers of the design system to render images or charts inside figure components, the chart and image directly call the figure component when being rendered. The figure component is also available to use separately to nest other content if desired.

This avoids an accessibility issue relating to which parameters are required. `audioDescription` is required for charts, but not needed for images which have alt text for that purpose. It also will allow us to avoid big changes to the service manual.

See #3448 for further information and ticket links.

### How to review this PR

Describe the steps required to test the changes (include screenshots if appropriate).

### Checklist

This needs to be completed by the person raising the PR.

-   [x] I have selected the correct Assignee
-   [x] I have linked the correct Issue

PR #3879, opened 2026-07-06
Figure component

helenb:

Follow on from the spikes:
https://github.com/ONSdigital/design-system/pull/3848
https://github.com/ONSdigital/design-system/pull/3851

#### New figure component
- Create a new figure component which is used internally by `onsImage`, `onsChart` and `onsTable` when they are rendered
- It can also be used independently with arbitrary child content
- Figure component renders a figure number, title, subtitle, caption, footnotes and downloads, which are all optional.
-  It can also use an audio description when relevant to describe the contents of the figure.

#### Featured article changes
- This PR also updates the featured article component to render arbitrary child content rather than explicitly rendering a chart or an image - this will pave the way to allowing the CMS to render an iframe inside the featured article

#### Image changes
- Update the image component to add figure number, title, subtitle, footnotes and downloads options, and to render itself using the figure component.
- Some text styles change as a result, and this is intentional.
- We do not allow an audio description to be added for the image as we use alt text to describe the image for screen readers.
- The existing caption option is now rendered using the figure component rather than the image component.

#### Chart changes
- Update the chart component to add the figure number option, and render itself using the figure component.
- The existing title, subtitle, caption, footnotes and downloads now all render with the figure component instead. Some text styles change, and this is intentional.
- The downloads are now rendered inside an onsDetails element when previously they were a list.
- The audio description option remains required for charts, but this is now passed through to the figure component to handle.

#### Table changes
- Update the table component to add figure number, title, subtitle, footnotes and downloads options, and to render itself using the figure component.
- In this case we render the onsFigure component with a div as it is not semantically correct to wrap a table inside a figure.
- The table caption stays as it was previously, and is not rendered using the onsFigure caption option, because it is semantically correct to render a table's `caption` inside the `table`.

#### List changes
- Update the list component to allow a `download` attribute to be set - this then allows e.g. `download="file"` to be passed to the downloads in the figure component.

Note that because of the text style changes for charts and images there are a lot of VR changes for this merge. I'll approve them once the rest of the changes have been reviewed and considered.

### How to review

- Check the changes for all affected components - chart, image, table and list.

### Checklist

- [x] I have linked the Issue


src/components/chart/_macro.njk
```diff
@@ -54,132 +44,110 @@
...
+                "figureNumber": params.figureNumber if params.figureNumber,
+                "title": params.title,
+                "subtitle": params.subtitle if params.subtitle,
+                "audioDescription": params.description,
+                "caption": params.caption if params.caption,
+                "footnotes": params.footnotes if params.footnotes,
+                "download": params.download if params.download
+            })
+        %}
+            <div class="ons-chart">
```
helenb, line 60: Query: should we set an id on the chart root element here as we do for table and image

helenb: Note that we set it on the iframe element if the chart is an iframe - this is what would be used and replaced by pym.js if it were currently functioning.

src/components/figure/_macro.njk
```diff
@@ -0,0 +1,88 @@
+{% from "components/list/_macro.njk" import onsList %}
+{% from "components/details/_macro.njk" import onsDetails %}
+
+{% macro onsFigure(params) %}
+    {# The heading levels are configurable so the figure slots into the correct place in the
+       document outline, but the title and subtitle always render at the same visual size
+       via the utility font-size classes below. #}
+    {% set headingLevel = params.headingLevel | default(2) %}
+    {% set openingTitleTag = '<h' ~ headingLevel ~ ' class="ons-figure__title ons-u-mb-2xs">' %}
```
helenb: I've used utility classes where they are available, e.g. for margins, but it might be cleaner to move everything to the scss file - I'm not sure what the convention is here. Below where we set the content margin dependent on whether there are elements above or below it is handy to use the utility classes.

src/components/chart/_macro.njk
helenb: Standardise just on the figure and remove from sub-elements.

src/components/figure/_macro.njk
helenb: Maybe move margins to the css file apart from where we have the top / bottom margin logic on content.

src/components/figure/_macro.spec.js
```diff
@@ -0,0 +1,439 @@
...
+                    },
+                },
+                CALLER_CONTENT,
+            );
+
+            expect(detailsSpy.occurrences[0].headingLevel).toBe(5);
+        });
+    });
+
+    describe('caption', () => {
```
helenb: Check if this is repeating the tests above.

src/components/chart/_macro.njk
helenb: Removed from the table element in [9ed0683dc46eb9a5018a11f0d52b9a2816237e1f](https://github.com/ONSdigital/design-system/pull/3879/changes/9ed0683dc46eb9a5018a11f0d52b9a2816237e1f)

src/components/figure/_macro.njk
helenb: Updated in [1bf9c76d01b01456ec39ee9efdc1ca533de96eb5](https://github.com/ONSdigital/design-system/pull/3879/changes/1bf9c76d01b01456ec39ee9efdc1ca533de96eb5)

src/components/figure/_macro.spec.js
helenb: Duplicate removed in [31ad33c3ae23b593ff266b2694dee870324fb5ad](https://github.com/ONSdigital/design-system/pull/3879/changes/31ad33c3ae23b593ff266b2694dee870324fb5ad)

sveltifier (approved):
Fantastic

PR #3883, opened 2026-07-15
Add 74.0.0 migration guide

helenb:
No related ticket

Adds the 73.x.x to 74.0.0 migration guide
Adds an extra breadcrumbs example for a 'wide' site

### How to review

Read through the migration guide and check it all makes sense, and check the new example

### Checklist

- [x] I have linked the Issue - marked as no related issue


migration_guides/73.x.x-to-74.0.0-migration-guide.md
```diff
@@ -1,43 +1,33 @@
...
+-   The `<preMain>` block has been moved so that it still renders before the `<main>` tag.
 
-    -   The breadcrumbs have been moved so that they render inside the `<header>` tag as this is more semantically correct.
+-   The `main` block has been renamed to `mainContent`.
 
-    -   The `<preMain>` block has been moved so that it still renders before the `<main>` tag
+-   New `breadcrumbs`, `preBreadcrumbs`, `hero` and `postHero` blocks.
 
-    -   Instructions added for any blocks that are rendered outside of an existing landmark to explain that they must use a landmark level element
+-   Instructions for any blocks that are rendered outside of an existing landmark to explain that they must use a landmark level element.
```
helenb: Nitpick I don't think this new wording quite makes sense now. And we should probably reference the example  so maybe  it could be

[Instructions have been added](https://github.com/ONSdigital/design-system/blob/main/src/foundations/page-template/example-base-page-template-block-areas.njk) for any blocks that are rendered outside of an existing landmark to explain that they must use a landmark level element.

sveltifier: Thanks, we'll add the links when we have Service Manual stuff done.

sveltifier (approved):

PR #3884, opened 2026-07-15
Add sourceNote option to Table 

helenb:

The CMS has a need for the table to display a visible 'source' below the table in addition to using the table caption in a hidden capacity. This updates the Table to allow it to display a 'sourceNote' below using the `onsFigure` caption option. It accepts a new parameter called `sourceNote` - the existing `caption` option is unchanged. Because we also have `useDivForBaseElement` set to `true` when the table renders, the figure caption will display using a `span` element.

### How to review

Test the new example and check it displays as expected, and ensure the new test is appropriate and the vr tests pass.

### Checklist

- [x] I have linked the Issue


sveltifier:
Cool, but I think we should consider renaming this property. 

- we already have `caption` and `hideCaption` options, so `figureCaption` is confusing
- requires the user to know about Figure / leaks implementation details of the table
- it's rendered as a span in this case, so not HTML caption semantics anyway

We could also then remove the implementation details in the docs, eg something like 

- ~~The caption rendered below the table using the onsFigure caption option. Note that for a table this will render as a span element.~~
- Additional note about the data displayed below the table. 

But what might be a good name for this? The example is the data source? That seems it would be a pretty universal option, so we could make it `source` or `sourceNote`? Or something like `note` with a doc of 'Can be used to name or describe the source data for the table'?

sveltifier (approved):
See comment re naming

helenb (commented):
Latest commit from @sveltifier looks good to me.

MaciekBaron (approved):
LGTM!

PR #3885, opened 2026-07-15
Reduce spacing between Table data and caption, footnotes or downloads 

helenb:

- Previously the table had a bottom margin of 1rem by default. The figure also adds a bottom margin of 1.5rem if there is a caption, footnotes or downloads to display.
- Now the default is to remove the bottom margin. The figure component still adds it's 1.5 rem if needed.
- In the case there is are no caption, footnotes or downloads below, the default 1rem bottom margin is re-applied

### How to review

Compare the examples for when there is figure content below and when there is not, and check the margins match what is expected.

### Checklist

- [x] I have linked the Issue


helenb:
@sveltifier the reason there is a change to `backstop_data/bitmaps_reference/ds-vr-test__foundations_type-scale_example-type-scale_0_document_2_mobile.png` is because it renders its own table markup without using the `onsTable` component. It uses the `ons-table` class which no longer has a bottom margin. It's a standalone example so doesn't matter really so I've just approved it.

sveltifier (approved):

PR #3886, opened 2026-07-15
Fix heading levels within Figure

helenb:

The figure heading level logic did not previously account for the fact that title and subtitle are optional. The new logic is as follows:

If there is a title and subtitle, downloads and footnotes need headingLevel + 2
If there is a title and no subtitle, downloads and footnotes need headingLevel + 1
If there is no title and no subtitle, downloads and footnotes need headingLevel + 0
If there is a subtitle and no title we should not display either, (and update the docs to make that clear) and downloads and footnotes still need headingLevel + 0

### How to review

Check the logic changes in the code all make sense and that tests and vr tests pass.

### Checklist

- [x] I have linked the Issue


src/components/figure/_macro-options.md
```diff
@@ -1,16 +1,15 @@
...
+| title                | string              | true                                                       | The title displayed above the nested content.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
```
sveltifier: Is `title` required?

sveltifier (dismissed):

MaciekBaron (approved):

PR #3891, opened 2026-08-10
Address results disappear in AddressInput

sdmiddletonSolirius:
See https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-981?atlOrigin=eyJpIjoiNWU1NDc3NDQ4ODA4NDA1YmIwNTI1NTRmZDlkOTM0MDQiLCJwIjoiaiJ9

Briefly describe what you have changed and why.

- Address input not clickable with mouse when user partially inputs a postcode, user could scroll to the addresses in a list, but on selection the address would clear and default to the search term not the individual address list.
- Replicated this on the **aims-diy** repo on the /typeahead page. 
- If you click on the Gate Reach address (29 addresses) you are presented with a list - when you click on gate reach with a mouse, the address list flashes for a moment but then clears (see video below).  However, you can still reach the address selection with keyboard directionals without any issues.

ex26ga-lookup

Result when clicking with mouse - defaults back to the parent search term:

default-back-to-search

**Fix:**
-  `autosuggest.ui.js` -  Cancel the blur timer in `selectResult` so a mouse click on a grouped postcode result keeps nested suggestions; reset `this.blurring = false` in `handleFocus` so that flag does not stay stuck after a cancelled blur.
- `autosuggest.address.spec.js` - Add mouse-click coverage for grouped suggestions (bucket request + nested options still present after 400ms / past the blur timeout), using the design-system’s faked address API.

### How to review

- Rebuild `main.js` and override `cdn.ons.gov.uk/sdc/design-system/73.1.1/scripts/main.js` while running the aims-diy UI (http://localhost:5001), since design-system does not hit a live postcode API.

**Video Evidence Before and After:**

**BEFORE:-** (address list flashes for a second and then dissappears)

https://github.com/user-attachments/assets/764388bb-d312-49d1-8a25-ec6a741b5ffb

**AFTER:-**

https://github.com/user-attachments/assets/89079262-5f9e-47f2-b3e8-f4561efccc33

### Checklist

- [X] I have linked the Issue
- [X] I have selected the Assignee
- [X] I have selected the most helpful Label
- [X] I have suggested an excellent Title for our release notes

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,3 +231,5 @@ npm publish --access public
 -   [SonarLint](https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarlint-vscode)
 -   [Sass](https://marketplace.visualstudio.com/items?itemName=Syler.sass-indented)
 -   [Stylelint](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint)
+
+<!-- Add and remove -->
--- a/README.md
+++ b/README.md
@@ -231,5 +231,3 @@ npm publish --access public
 -   [SonarLint](https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarlint-vscode)
 -   [Sass](https://marketplace.visualstudio.com/items?itemName=Syler.sass-indented)
 -   [Stylelint](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint)
-
-<!-- Add and remove -->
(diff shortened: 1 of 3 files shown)


SriHV (dismissed):
Looks good to me. All the tests pass. Videos(before and after) helped me to understand the issue and fix clearly 👍

sveltifier (approved):

PR #3892, opened 2026-08-12
Header search label is not shown in Welsh

sdmiddletonSolirius:
See [[ONSDESYS-938]](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-938)

Screenshot 2026-08-13 at 13 53 07

Briefly describe what you have changed and why.

### How to review

Describe the steps to review and test these changes.

- Updated `src/components/header/_macro.njk` so the default search input label uses `currentLanguageIsoCode == 'cy'` instead of `params.lang == 'cy'`.
- Added regression test:  WHEN: inputLabel is not provided and Welsh is the current language → asserts #header-search-input-label is Chwilio’r SYG.

- The Welsh default was wired to `params.lang`, but the header does not receive `params.lang` from the page template.
- Language is passed via `params.language.languages`, and the header already derives `currentLanguageIsoCode` from the entry with current: true (also used for logos / browser banner).
- Using that same source means Welsh pages without an explicit `inputLabel` correctly fall back to Chwilio’r SYG; English still falls back to Search the ONS.
- Explicit overrides (`params.search.form.inputLabel / searchLinks.searchNavigationInputLabel`) are unchanged and still take precedence.
- The new unit test locks in the failing scenario (Welsh current + no inputLabel) so the bug can’t return unnoticed.

**For testing purposes** - In the absence of a live in browser toggle switch, I've removed the explicit `search.form.inputLabel` (which overrides languages) and manually changed current language to `true` or `false` for `en` or `cy` use cases, showing that the languages switch to the correct translation.

https://github.com/user-attachments/assets/16a6b77d-2912-4583-b5b4-fb2d5c04ca18

### Checklist

- [X] I have linked the Issue
- [X] I have selected the Assignee
- [X] I have selected the most helpful Label
- [X] I have suggested an excellent Title for our release notes

[ONSDESYS-938]: https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,3 +231,5 @@ npm publish --access public
 -   [SonarLint](https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarlint-vscode)
 -   [Sass](https://marketplace.visualstudio.com/items?itemName=Syler.sass-indented)
 -   [Stylelint](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint)
+
+<!-- Add and remove -->
--- a/README.md
+++ b/README.md
@@ -231,5 +231,3 @@ npm publish --access public
 -   [SonarLint](https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarlint-vscode)
 -   [Sass](https://marketplace.visualstudio.com/items?itemName=Syler.sass-indented)
 -   [Stylelint](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint)
-
-<!-- Add and remove -->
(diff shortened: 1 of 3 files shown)


SriHV (dismissed):

SriHV:
Just a small thing. The commits on this PR needs to be verified. Otherwise, looks good to me!

sveltifier (approved):

PR #3896, opened 2026-08-19
Improve linting, pre-commit checks and tools

sveltifier:
No Jira issue.

- Don't silently fix/alter code on commit. Fail instead.
- Make a global `lint` script that runs locally and can (in future) be added to CI
- Make a global `lint:fix` script that can (in future) be used to fix lint issues repo-wide
- Include `.vscode` tool recommendations ~~and repo-level settings~~

### How to review

Check linting and commit behaviour.

### Checklist

- [x] I have linked the Issue


.vscode/settings.json
```diff
@@ -0,0 +1,19 @@
+{
```
helenb: I am not sure I agree with enforcing vs code settings in the repo, as people usually have their own vs code settings in place. Not sure if it is something that is common elsewhere in ONS though?

MaciekBaron: I agree with Helen, I think it would be better to update the `.editorconfig` if necessary

sveltifier: OK! Thanks, removed.

SriHV (approved):

PR #3905, opened 2026-09-16
Fix Back to Top in nested layouts

helenb:
See [Link to Jira issue](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-996)

- Correctly calculate the location of the previous content element even if the back to top link is nested several elements deep
- If there is no content at all above the back to top component, don't try and calculate the content height or make it sticky
- Add tests for the above scenarios to ensure no errors are thrown

### How to review

- You may need to create some new example files to test this easily in the browser.
    - Create one where the back-to-top component has content before it but the component call is nested several levels deep e.g. inside several `<div>`s
    - Create another example where the back-to-top component is the only element on the page
    - For both examples resize the browser and scroll and ensure no console errors are thrown
 - Make sure all tests pass, including the new tests

### Checklist

- [x] I have linked the Issue


sveltifier (approved):
Looks great, examples created locally

PR #3906, opened 2026-09-17
Upgrade dependencies for vulnerabilities

sveltifier:
See https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-988

Upgrades the Design System’s dependency graph to address reported vulnerabilities and bring the build, lint, test, formatting, and browser automation tooling up to date.

Highcharts remains on `12.1.2`: the attempted Highcharts 13 upgrade caused visual regressions in chart annotations and axis labels, particularly at tablet and mobile widths.

Adds targeted Yarn resolutions where vulnerable packages remain nested beneath dependencies that have not yet updated their declared ranges:

- BackstopJS’s Puppeteer dependency is aligned to Puppeteer `25.11.0`, removing its vulnerable older Puppeteer / `extract-zip` chain.
- `decode-uri-component` is pinned to `0.5.0` for the legacy dependency below `gulp-sourcemaps`.
- Existing Lodash, Jest environment, PostCSS, and UUID overrides are retained; redundant Micromatch and YAML overrides have been removed.

Updates the test infrastructure for compatibility with the newer packages:

- Cheerio 1.2 changed its export/API behavior and pulls in Undici. Jest is configured to load Cheerio’s Node-compatible export, and a test setup file supplies the Node web APIs that Undici expects but JSDOM does not expose by default.
- The Cheerio test helper now uses the supported collection API rather than the removed callable default import.
- Puppeteer 25’s public module is ESM-only. The two affected test suites now use a shared local iPhone X emulation fixture instead of importing `KnownDevices`.
- Browser test setup timeouts are increased from five to ten seconds to accommodate parallel Puppeteer initialization under the upgraded tooling.

These code changes are limited to test/build compatibility and do not alter the published component templates, CSS, or client-side component behavior.

Validation:

- `yarn test`: 114 suites passed, 2,509 tests passed, 10 snapshots passed.
- Visual regression passing.

Remaining vulns:

- `yarn audit`: only one remaining low-severity issue, `elliptic` via the legacy Browserify crypto shim chain. No patched release is available for that dependency.

### How to review

Apart from visually reviewing the code, there's little more that can be done than relying on the fact that the all the test suites still pass. 

### Checklist

- [x] I have linked the Issue


package.json
```diff
@@ -61,98 +61,99 @@
...
+        "whatwg-fetch": "3.6.20"
     },
     "publishConfig": {
         "access": "public"
     },
     "dependencies": {
         "highcharts": "12.1.2",
-        "pym.js": "1.3.2"
+        "pym.js": "1.3.2",
+        "uuid": "14.0.2"
```
sdmiddletonSolirius: `uuid@14.0.2` was added as a direct dependency, but nothing in the repository imports it. Yarn consequently installs both `uuid@14.0.2` and the transitive `uuid@11.1.1` via prepend-file/temp-write. The uuid@11.1.1 resolution also overrides `temp-write’s` declared `uuid@^3.3.2` range, causing the build warning. 

`warning Resolution field "uuid@11.1.1" is incompatible with requested version "uuid@^3.3.2"a`

Removing both the direct dependency and the uuid resolution should eliminate the warning and unnecessary published dependency.

sveltifier: Good spot

sdmiddletonSolirius (approved):
Just the minor UUID dependency hygiene comment (which could be moved to a separate PR) but not a blocker for vulnerability fixes, so all looks good.

PR #3907, opened 2026-09-18
Charts - reduce the buffer zone for the tooltip hover

helenb:
See [Link to Jira issue](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-1014)

For all tooltips, reduce the hover area that triggers the tooltip to be 2px. It is not possible to restrict this setting to just one chart type or series type.

Adds an example for the stacked bar chart with line.

I considered using a larger buffer for touch interactions, but testing on the ONS touchscreen desktop it seems to trigger ok when I touch the line.

### How to review

- Check the new example
- If you hover over a small column that is also behind a line, you should be able to view the tooltip for the column, not the line tooltip

<details>
<summary>See screenshots</summary>
Screenshot of hover without the new snap setting - the line tooltip displays even though the cursor is over the purple area

Screenshot 2026-09-18 at 11 10 39

Screenshot of hover with the new snap setting - the column tooltip displays

Screenshot 2026-09-18 at 11 09 57
</details>

### Checklist

- [x] I have linked the Issue


sveltifier (approved):
Simple change 👍

